PR #1, opened 2017-12-06
RFC: New version of context

acdlite:
### Superseded by #2. Please go there instead. I encountered a GitHub bug and had to re-open.

A proposal for a new version of context addressing existing limitations.

[Rendered](https://github.com/acdlite/rfcs/blob/new-version-of-context/text/0000-new-version-of-context.md)

TODO:

- [ ] Implementation notes
- [ ] High-priority version of context, for animations. (May submit this separately.)

### DIFF
--- a/text/0000-new-version-of-context.md
+++ b/text/0000-new-version-of-context.md
@@ -383,6 +383,13 @@ render() {
 }
 ```
 
+## Add displayName argument to createContext for better debugging
+
+For warnings and React DevTools, it would help if providers and consumers
+had a `displayName`. The question is whether it should be required. We could
+make it optional, and use a Babel transform to add the name automatically. This
+is the strategy we use for `createClass`.
+
 ## Other
 
 * Should the consumer use `children` as a render prop, or a named prop?


acdlite:
@ryanflorence Pretty cool! Just so nobody reading this is confused, one difference is that the argument passed to `React.createContext` in this proposal is the default value. If a consumer is rendered without a provider, the default value is used. This means context can be statically typed!

```js
type Provider<T> = React.Component<{
  value: T,
  children?: React.Node,
}>;

type Consumer<T> = React.Component<{
  children: (value: T) => React.Node,
}>;

interface React {
  createContext<T>(defaultValue: T): Context<T>;
}
```

text/0000-new-version-of-context.md
```diff
@@ -0,0 +1,400 @@
...
+
+# Basic example
+
+```js
+type Theme = 'light' | 'dark';
+// Pass a default theme to ensure type correctness
+const ThemeContext: Context<Theme> = React.createContext('light');
+
+class ThemeToggler extends React.Component {
+  state = {theme: 'red'};
```
pspeter3: This doesn't match the theme type. Maybe that was intentional?

acdlite: Oops, oversight. Good catch!

jlongster:
> This means context can be statically typed!

Couldn't the type just be `Optional<T>`? (or whatever your type system's version of Optional is.) What is the use case for allowing consumers to be rendered for detached? I can't think of any, and seems like it should throw a hard error to make sure the user set things up correctly. (I run components in isolation in tests and stuff, but I'm happy to render a top-level component that provides the mock value for my tests in there. Point is, in those cases, you'll likely need something other than the default value if you're doing that kind of thing)

I love the render prop API.

acdlite:
@jlongster See the "Unresolved questions" section:

> ### Should we warn if a consumer is rendered without a matching provider?
> There are valid use cases for relying on the default context value. In many
> or most cases, it's likely a developer mistake. We could print a warning in
> development when a consumer is rendered without a matching provider. To supress
> the warning, the developer passes `true` to `allowDetached`.

jlongster:
I saw that but it doesn't actually list the valid use cases :) I would argue we should treat it always as a mistake and avoid an additional `allowDetached` prop.

acdlite:
I think the use cases are similar to `defaultProps`. Like a button that has a default theme, but changes once you place it inside a Theme provider.

acdlite:
We're also considering the ability to set `active={false}` to stop listening for changes. I couldn't think of any uses cases except ones that depend on time or animation (which I'll likely submit as a separate proposal), so I left it out.

text/0000-new-version-of-context.md
```diff
@@ -0,0 +1,400 @@
...
+  return (
+    <Provider value={this.state.contextValue}>
+      {this.props.children}
+    </Provider>
+  );
+}
+```
+
+To update the context value, the parent re-renders and passes a different value.
+Changes to context are detected using `Object.is` comparison. (Referred to as
```
milesj: May have overlooked it, but how exactly is a change detected in the consumer?

acdlite:
Weird, I pushed a commit but it's not showing up here...

text/0000-new-version-of-context.md
acdlite: When the provider receives a new value, React re-renders the consumers.

I'll add more details about the implementation shortly.

jaredpalmer:
This is awesome. I like the render prop API as it just seems like less ceremony and typing. I agree with @jlongster about `allowDetached` too

text/0000-new-version-of-context.md
```diff
@@ -0,0 +1,400 @@
...
+using subscriptions even with the new API, if that makes the most sense for
+their use case.
+
+Coordinate with library authors to migrate to the new API. Once the major
+libraries are ready, and we've allowed enough time for users to migrate, we will
+deprecate the old API in a minor update.
+
+In the following major release, we'll remove the old API.
+
+# Unresolved questions
```
milesj: I think a life-cycle hook in regards to context changes should be discussed. Relying on a child function render for detecting this isn't exactly feasible for all situations. Perhaps another prop on the consumer?

```jsx
<Consumer onChange={this.handleChange} />
```

mjackson:
Hmm, looks like my previous comments got buried in the "show outdated" view. Please pardon for re-posting down here...

- The ability for subscribers to render out of context is useful primarily in testing scenarios where people are too lazy/don't want to set up the correct context before they run their test. We heard a lot of feedback about this in early versions of react-router with our `<Link>` element. People wanted to render a component with a `<Link>` inside it w/out having to set up the rest of the router context. In these situations, using the context's default value is fine. react-broadcast [recently introduced a `<Subscriber quiet>` prop](https://github.com/ReactTraining/react-broadcast/commit/acb3e752693e135c07e3ee70887d7e1c20223f62) that serves this purpose.
- We recently [added a `compareValues` prop](https://github.com/ReactTraining/react-broadcast/commit/e20019bf282b523221cb644df813398a62c7a884) to react-broadcast that allows publishers to declare a function they can use to compare values to see if they need an update or not. This seems like a useful escape hatch since we can't assume everyone is using immutable types.

jlongster:
> I think the use cases are similar to defaultProps. Like a button that has a default theme, but changes once you place it inside a Theme provider.

The difference to me is props are meant to be used within the component. Context is meant to be used with a Provider, always, even if the provider just used the default value. Making this strict might be a way to keep the API simple and usage clear.

acdlite:
Sorry folks... I hit a weird GitHub bug. GitHub doesn't recognize changes to this PR since I switched the repo from private to public. Going to close and open a new one. Sorry for the disturbance, let's continue over there!

#2

jlongster:
@mjackson Interesting. Personally I was happy to add a top-level MemoryRouter for tests/etc and I thought that was quite elegant. In fact it allowed me to specify the initial state of the Router which was very useful, like `<Router initialEntries={[selected]} initialIndex={0}>`.

mjackson:
@jlongster You and me both! But it really bothered some people. I can see how the default context value will help with that, though. Now you've got the default value built into the component, instead of needing to get it from another one further up the tree.

gaearon:
Locking since the discussion has moved to #2. Please continue there!

jlongster:
poop

Issue #3, opened 2017-12-07
Contribution Guide/Syntax Preferences


alexkrolick:
Looking over the Context proposal (#2 ), it's using ~~~Stage-2~~~ _Stage-3_ ES and Flow typings, which are great, but for someone not familiar with Flow a bit hard to follow in parts. However, React's codebase _does_ use Flow, so it might be considered fair game. We've added syntax guidelines [to the docs]( https://github.com/reactjs/reactjs.org/blob/master/CONTRIBUTING.md#guidelines-for-code-examples) already, do you think something like that would be warranted for the RFCs, too?

TrySound:
@alexkrolick Flow in the proposal is example of new api benefits for static type checking. Class fields are now in stage-3.

acdlite:
Flow is helpful for documenting APIs, and proving that they can be soundly typed.

I'm not too worried about using stage 3 features in RFCs. We've avoided class property syntax in the docs because some of our audience may still be learning JavaScript. But I don't think that's a concern here. @gaearon may have opinions on this.

Syntax guidelines are a nice idea. A PR would be appreciated, if you're up for it.

gaearon:
>@gaearon may have opinions on this.

My opinion is we actually *should* be using features we're pushing since that's the end game we're driving towards in RFCs.

PR #5, opened 2017-12-08
Add RFC Guidelines

alexkrolick:
#3


trueadm:
I'd be happy with either Flow or TypeScript types. In my experience, both offer the same benefits in cases of RFCs. I don't feel like we should dictate a type annotation preference to a community; in fact it will likely alienate developers from contributing potentially awesome ideas.

alexkrolick:
Good point, changed.

PR #7, opened 2017-12-09
Add dangerouslyRecycleNode RFC

zackargyle:
Allow a DOM node from an unmounted component to be reused within a newly mounted component. This is particularly helpful when doing cross-route transition animations, or forwarding video playback to a detailed view from a feed.

This is a new feature request, and would have no backwards incompatibility issues.

The proposed solution is to add a `dangerouslyRecycleNode={id}` react internal prop that can be added to raw dom nodes (like `dangerouslySetInnerHTML`). If an unmounting node contains the prop, React caches the node and if a mounting node has the same `id`, it will reuse the exact DOM node from the previously unmounted component.

[Proposal](https://github.com/zackargyle/rfcs/blob/dangerouslyRecycleNode/0000-dangerouslyRecycleNode.md)

### DIFF
--- a/0000-dangerouslyRecycleNode.md
+++ b/0000-dangerouslyRecycleNode.md
@@ -1,0 +1,76 @@
+- Start Date: 2017-12-09
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Allow a DOM node from an unmounted component to be reused within a newly mounted component. This is particularly helpful when doing cross-route transition animations, or forwarding video playback to a detailed view from a feed.
+
+This is a new feature request, and would have no backwards incompatibility issues.
+
+# Basic example
+
+The proposed solution is to add a `dangerouslyRecycleNode={id}` react internal prop that can be added to raw dom nodes (like `dangerouslySetInnerHTML`). If an unmounting node contains the prop, React caches the node and if a mounting node has the same `id`, it will reuse the exact DOM node from the previously unmounted component.
+
+```js
+class FeedPage extends React.Component {
+  changeRoute(id) {
+    this.history.push(`/feedItem/${id}/`);
+  }
+  render() {
+    const { feedItems } = props;
+    return (
+      <div>
+        {feedItems.map(({ imageSrc, id }) =>
+          <img
+            class="small-img"
+            key={id}
+            src={imageSrc}
+            onClick={this.changeRoute.bind(this, id)}
+            dangerouslyRecycleNode={id}
+          />)}
+      </div>
+    );
+  }
+}
+class DetailPage extends React.Component {
+  render() {
+    const { imageSrc, id } = this.props.feedItem;
+    return (
+      <img
+        class="large-img"
+        src={imageSrc}
+        dangerouslyRecycleNode={id}
+      />
+    );
+  }
+}
+```
+
+# Motivation
+
+Reusing DOM nodes between routes can be a wonderful user experience, but is impossible within React core currently. The library [react-overdrive](https://github.com/berzniz/react-overdrive) was the inspiration for the API of this RFC. This would be a fairly simple addition to React core that would allow for some beautifully declarative route transitions.
+
+## Detailed Design
+
+The proposal is to add a `dangerouslyRecycleNode` prop to raw DOM nodes. Internally on unmount if a node has the prop set, it will add it to the internal `domNodeCache[key]`. If another node with the matching prop + recycle id is then mounted, React will not create a new node, but will extract the cached DOM node and use it in the newly mounted component. It will then immediately remove it from cache.
+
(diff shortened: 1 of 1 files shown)


0000-dangerouslyRecycleNode.md
```diff
@@ -0,0 +1,76 @@
...
+    );
+  }
+}
+class DetailPage extends React.Component {
+  render() {
+    const { imageSrc, id } = this.props.feedItem;
+    return (
+      <img
+        class="large-img"
+        src={imageSrc}
```
NE-SmallTown: class -> className ?

0000-dangerouslyRecycleNode.md
```diff
@@ -0,0 +1,76 @@
...
+}
+```
+
+# Motivation
+
+Reusing DOM nodes between routes can be a wonderful user experience, but is impossible within React core currently. The library [react-overdrive](https://github.com/berzniz/react-overdrive) was the inspiration for the API of this RFC. This would be a fairly simple addition to React core that would allow for some beautifully declarative route transitions.
+
+## Detailed Design
+
+The proposal is to add a `dangerouslyRecycleNode` prop to raw DOM nodes. Internally on unmount if a node has the prop set, it will add it to the internal `domNodeCache[key]`. If another node with the matching prop + recycle id is then mounted, React will not create a new node, but will extract the cached DOM node and use it in the newly mounted component. It will then immediately remove it from cache.
```
kyleshevlin: Asking for some clarity here. Your use case it to maintain a DOM node for component/route transitions, but then you say to "immediately remove it from cache". What happens when the user wishes to use the Back button? Would we need to maintain the node in cache, or would we be able to setup the same transition, just in reverse?

zackargyle: That's a great point. Maybe it should remain in cache until there is no element with the corresponding ID. So don't remove it from cache immediately after mounting the new tree, but set a timeout after unmounting the recycleable node. The cache could have some kind of data structure like this?
```
type RecycleCache = {
  [id: string]: {
    node: HTMLElement,
    unmountTime: null | number,
  }
};
```

kyleshevlin: I think that would be an option. I think the other is to clear the cache but enable transitions in both directions somehow. I mean, in theory, the same DOM node could transition across more than one dismount of parent components. Having the ability to do this no matter how many destinations I think could be useful.

0000-dangerouslyRecycleNode.md
```diff
@@ -0,0 +1,76 @@
+- Start Date: 2017-12-09
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Allow a DOM node from an unmounted component to be reused within a newly mounted component. This is particularly helpful when doing cross-route transition animations, or forwarding video playback to a detailed view from a feed.
+
+This is a new feature request, and would have no backwards incompatibility issues.
```
mattbasta: This would, however, be backwards incompatible with any existing component that--though unlikely--uses a prop named `dangerouslyRecycleNode`.

mattbasta:
Perhaps a solution to the memory leak issue is to require the developer to specify their own cache which they manage manually. For instance:

```js
const myCache = new Map();

const RecycledCompoonent = (props) =>
  <MyComponent
    dangerouslyRecycleNode={{cache: myCache, key: props.id}}
    {...props}
  />;
```

This would also solve many unaddressed problems:

- Third party libraries wouldn't need to worry about conflicting with IDs from other libraries
- Developers could add their own cache eviction policy. This RFC is fundamentally a performance enhancement, and performance is not one-size-fits-all solution.
- The cache could be traversed manually, flushed, or otherwise manipulated if a change in application state makes cached values invalid.
- The cache key is not limited to strings.
- Developers could implement their own `.set()` method that rejects nodes from the cache, potentially saving resources under memory pressure.
- etc.

0000-dangerouslyRecycleNode.md
zackargyle: This could be fixed with a codemod to rename their prop.

sophiebits:
FWIW, one workaround you can do today is to render the node you want to move into its own React root (via a portal), and then manually move that container around using DOM operations. However it is clumsy and does require extra wrapper elements.

gaearon:
The transition use case is interesting and something @sebmarkbage has recently been thinking about. However, it is unlikely that recyling the DOM node is how we'll go about it. Instead, we might want to have a first-class way for components to specify their exit animations. This will likely tie into other Concurrent Mode features. So I'll close this particular RFC but we care about the use case.

Note that the "reverse portal" use case can be implemented in userland. I describe how here: https://github.com/facebook/react/issues/12247#issuecomment-486378549.

PR #6, opened 2017-12-08
async-safe, static lifecycle hooks

bvaughn:
Replace error-prone render phase lifecycle hooks with static methods to make it easier to write async-compatible React components. Provide a clear migration path for legacy components to become async-ready.

Note that I plan to submit a related RFC soon for a new server-side lifecycle, `componentDidServerRender`, to offset any server-rendering functionality lost by deprecating `componentWillMount`.

Please leave comments below for high-level topics. For feedback on specific parts of this proposal, please leave inline comments (on the markdown file).

[View proposal with formatting](https://github.com/bvaughn/rfcs/blob/static-lifecycle-methods/text/0000-static-lifecycle-methods.md)

PR #8, opened 2017-12-11
static server-only post-render lifecycle method

bvaughn:
Provide a place for initialization logic with side effects to safely run on the server.

This RFC relates to the proposal to deprecate `componentWillMount` ([RFC #6](https://github.com/reactjs/rfcs/pull/6)) and has been [previously discussed on GitHub](https://github.com/facebook/react/issues/7671).

Please leave comments below for high-level topics. For feedback on specific parts of this proposal, please leave inline comments (on the markdown file).

[View proposal with formatting](https://github.com/bvaughn/rfcs/blob/component-did-server-render/text/0000-component-did-server-render.md)

### DIFF
--- a/text/0000-component-did-server-render.md
+++ b/text/0000-component-did-server-render.md
@@ -1,0 +1,79 @@
+- Start Date: 2017-12-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Provide a server-side rendering equivalent for `componentDidMount`.
+
+This RFC relates to the proposal to deprecate `componentWillMount` ([RFC #6](https://github.com/reactjs/rfcs/pull/6)) and has been [previously discussed on GitHub](https://github.com/facebook/react/issues/7671).
+
+# Basic example
+
+Because `componentWillMount` is the only lifecycle method currently invoked during server rendering, it is sometimes used for things like logging, eg:
+```js
+class Example extends React.Component {
+  componentWillMount() {
+    if (typeof window === 'undefined') {
+      // Server-side
+    }
+  }
+
+  componentDidMount() {
+    // Client-side
+  }
+}
+```
+
+This RFC proposes to add a new lifecycle hook named `componentDidServerRender` to be called after initial server render, eg:
+```js
+class Example extends React.Component {
+  componentDidServerRender() {
+    // Server-side
+  }
+
+  componentDidMount() {
+    // Client-side
+  }
+}
+```
+
+# Motivation
+
+Provide a place for initialization logic with side effects to safely run on the server.
+
+Typically, `componentWillMount` is used for this, but [RFC #6](https://github.com/reactjs/rfcs/pull/6) proposes to remove this method. (It is not a very ergonomic solution anyway.) The logic could be moved to the component constructor but it's generally considered bad practice for a constructor to have side effects.
+
+# Detailed design
+
+## `componentDidServerRender()`
+
+`componentDidServerRender()` is invoked immediately after a component is server rendered. It is the server equivalent to `componentDidMount()`.
+
+Calling `setState()` in this method will trigger an extra rendering, but it is guaranteed to flush during the same tick. This guarantees that even though the `render()` will be called twice in this case, the user won't see the intermediate state.
+
+Use this pattern with caution because it can impact performance. However, it is sometimes necessary for cases like localization or routing where you need to override the results of the initial render based on new information.
+
+# Drawbacks
+
+This proposal is backwards compatible.
+
+# Alternatives
+
+The alternative to this new lifecycle hook is to continue to feature test (eg `typeof window`) inside of either `componentWillMount` or the class constructor.
+
+# Adoption strategy
+
+Release a minor update to version 16 with support for the new lifecycle hook.
+
+Pending the outcome of [RFC #6](https://github.com/reactjs/rfcs/pull/6), include messaging about this new hook in the `componentWillMount` deprecation warning.
+
+Coordinate with popular libraries (eg [react-router](https://reacttraining.com/react-router/), [react-intl](https://github.com/yahoo/react-intl)) to ensure this new hook meets their needs before `componentWillMount` is deprecated. 
+
+# How we teach this
+
+Write a blog post for [reactjs.org](https://reactjs.org/) explaining the motivation for this new hook as well as showing a basic example of its usage. Add it to the [component reference API](https://reactjs.org/docs/react-component.html) as well.
+
+# Unresolved questions
+
+None?
--- a/text/0000-component-did-server-render.md
+++ b/text/0000-component-did-server-render.md
@@ -42,10 +42,210 @@ class Example extends React.Component {
 
 Provide a place for initialization logic with side effects to safely run on the server. This includes things like:
 * Logging/metrics
+# Smarter default server-side behavior for virtualization/windowing libraries
 * Routing redirects if match was found (eg the legacy react-router `Match`/`Miss` components)
 * Initializaing shared localization data
 
 Typically, `componentWillMount` is used for this, but [RFC #6](https://github.com/reactjs/rfcs/pull/6) proposes to remove this method. (It is not a very ergonomic solution anyway.) The logic could be moved to the component constructor but it's generally considered bad practice for a constructor to have side effects.
+
+## Logging example
+
+This example shows how you might use the new `componentDidServerRender` method to do log analytics data for server renders.
+
+```js
+import API from './your-api-abstraction';
+
+class AnalyticsComponent extends React.Component {
+  componentDidServerRender() {
+    API.logSomeData();
+  }
+}
+```
+
+## Server-side fallbacks for virtualization/windowing libraries
+
+This example shows how a windowing library (like [react-virtualized](https://github.com/bvaughn/react-virtualized)) might implement a smarter default behavior when being rendered on the server.
+
+Windowing libraries typically require their dimensions to be injected via props in order to function. These measurements typically come from the DOM, after initial mount, and are unavailable on the server. This means that be default, no windowed content will render. A component like [`AutoSizer`](https://github.com/bvaughn/react-virtualized/blob/master/source/AutoSizer/AutoSizer.js) could use this new lifecycle hook to determine that it's been server-rendered and default to showing a fixed amount of content:
+
+```js
+class AutoSizer extends React.Component {
+  static defaultProps = {
+    serverHeight: 400,
+    serverWidth: 800,
+  };
+
+  state = {
+    height: 0,
+    width: 0,
+  };
+
+  componentDidMount() {
+    // Measure actual DOM element size and store in state
+  }
+
+  componentDidServerRender() {
+    // Use a default size to render *some* content
+    this.setState({
+      height: this.props.serverHeight,
+      width: this.props.serverWidth,
+    });
+  }
+
+  render() {
+    const { height, width } = this.state;
+
+    return this.props.children({
+      height,
+      width,
+    });
+  }
+}
+```
+
+The above example could be re-written to _assume_ a server context by default (with a flag in `state` that is updated by `componentDidMount`). This could result in the windowing component creating more content than is necessary on the initial render though (eg if the server fallback was larger than the actual measured size).
+
+## Routing redirect example
+
+This example shows how you might use the new `componentDidServerRender` method (along with `context`) to create a router that renders a default view if the current URL does not match one of the specified patterns. (Note that this example is over-simplified and a little silly.)
+
+```js
+class Router extends React.Component {
+  static childContextTypes = {
+    hasMatches: PropTypes.func.isRequired,
+    registerMatch: PropTypes.func.isRequired,
+  };
+
+  static GlobalState = {
+    matched: false,
+    hasMatches: () => Router.GlobalState.matched,
+    registerMatch: matched => {
+      if (matched) {
+        Router.GlobalState.matched = true;
+      }
+    },
+    resetMatched: () => {
+      Router.GlobalState.matched = false;
+    },
+  };
+
+  getChildContext() {
+    return {
+      hasMatches: Router.GlobalState.hasMatches,
+      registerMatch: Router.GlobalState.registerMatch,
+    };
+  }
+
+  componentWillUpdate() {
+    Router.GlobalState.resetMatched();
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+class Match extends React.Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    location: PropTypes.object.isRequired,
+    pattern: PropTypes.string.isRequired,
+  };
+
+  static contextTypes = {
+    registerMatch: PropTypes.func.isRequired,
+  };
+
+  componentDidMount() {
+    this._registerMatch();
+  }
+
+  componentDidServerRender() {
+    this._registerMatch();
+  }
+
+  componentDidUpdate() {
+    this._registerMatch();
+  }
+
+  render() {
+    return this._doesMatch()
+      ? this.props.children
+      : null;
+  }
+
+  _doesMatch() {
+    const { location, pattern } = this.props;
+
+    // This is over-simplified :)
+    return location.pathname === pattern;
+  }
+
+  _registerMatch() {
+    const { registerMatch } = this.context;
+    const { pattern } = this.props;
+
+    registerMatch(this._doesMatch());
+  }
+}
+
+class Miss extends React.Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+  };
+
+  static contextTypes = {
+    hasMatches: PropTypes.func.isRequired,
+  };
+
+  state = {
+    hasMatchesInContext: true
+  };
+
+  componentDidMount() {
+    this._checkForMatches();
+  }
+
+  componentDidServerRender() {
+    this._checkForMatches();
+  }
+
+  componentDidUpdate() {
+    this._checkForMatches();
+  }
+
+  render() {
+    return this.state.hasMatchesInContext
+      ? null
+      : this.props.children;
+  }
+
+  _checkForMatches() {
+    const { hasMatchesInContext } = this.state;
+    const { hasMatches } = this.context;
+
+    if (hasMatches() !== hasMatchesInContext) {
+      this.setState({
+        hasMatchesInContext: hasMatches(),
+      });
+    }
+  }
+}
+
+const App = () => (
+  <Router>
+    <Match location={location} pattern='/some/url'>
+      <SomeView />
+    </Match>
+    <Match location={location} pattern='/some/other/url'>
+      <SomeOtherView />
+    </Match>
+    <Miss>
+      <DefaultView />
+    </Miss>
+  </Router>
+);
+```
 
 # Detailed design
 
--- a/text/0000-component-did-server-render.md
+++ b/text/0000-component-did-server-render.md
@@ -47,7 +47,7 @@ class Example extends React.Component {
 
 Provide a place for initialization logic with side effects to safely run on the server. This includes things like:
 * Logging/metrics
-# Smarter default server-side behavior for virtualization/windowing libraries
+* Smarter default server-side behavior for virtualization/windowing libraries
 * Routing redirects if match was found (eg the legacy react-router `Match`/`Miss` components)
 * Initializaing shared localization data
 
--- a/text/0000-component-did-server-render.md
+++ b/text/0000-component-did-server-render.md
@@ -4,9 +4,9 @@
 
 # Summary
 
-Provide a server-side rendering equivalent for `componentDidMount`.
+Add a new, server-side only, static lifecycle method to be invoked after a component and its children have been rendered.
 
-This RFC relates to the proposal to deprecate `componentWillMount` ([RFC #6](https://github.com/reactjs/rfcs/pull/6)) and has been [previously discussed on GitHub](https://github.com/facebook/react/issues/7671).
+This RFC relates to the proposal to deprecate `componentWillMount` ([RFC #6](https://github.com/reactjs/rfcs/pull/6)) and has been previously discussed on GitHub (see [facebook/react/issues/7671](https://github.com/facebook/react/issues/7671) and [facebook/react/issues/7678](https://github.com/facebook/react/issues/7678)).
 
 # Basic example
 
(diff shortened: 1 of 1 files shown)


yordis:
@bvaughn I a curious about something. This server stuff is related to Web environment, I wouldn't have any SSR on React Native, React VR, React [Insert Renderer that is not Web here] so why would be open functions based on specific platform?

I would be fine if put it on React DOM but I don't think how that would work.

bvaughn:
You're right, at the moment a server-side hook primarily impacts the `react-dom` renderer.

I think it might be possible for other renderers, like `react-native`, to be server-side rendered too. Any renderer can support server-rendered content by implementing the [`HydrationHostConfig`](https://github.com/facebook/react/blob/72992382788a46e901f257c8314d30cce57ce0e7/packages/react-reconciler/src/ReactFiberReconciler.js#L157) interface. (Put another way, the concept of server-rendering isn't specific to the DOM renderer, even if that's the only renderer currently taking advantage of it.)

text/0000-component-did-server-render.md
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  componentDidMount() {
+    // Client-side
+  }
+}
+```
+
+This RFC proposes to add a new lifecycle hook named `componentDidServerRender` to be called after initial server render, eg:
```
gaearon: I'm not a fan of this naming but I don't have better suggestions.
It does feel odd to codify a specific environment into the core component API.

text/0000-component-did-server-render.md
```diff
@@ -0,0 +1,79 @@
...
+
+  componentDidMount() {
+    // Client-side
+  }
+}
+```
+
+# Motivation
+
+Provide a place for initialization logic with side effects to safely run on the server.
```
gaearon: What kind of side effects? Can we consider a few specific ones? It's not entirely obvious.

bvaughn: Based on use-cases I've seen mentioned:
* Logging/metrics
* Localization initialization
* Redirect routing if no path-match was found

I can add some examples to the markdown though, sure!

text/0000-component-did-server-render.md
mbrevda: perhaps `componentDidStaticRender`, to signify that there will be no more render's after this? Or `componentDidFinalRender`?

bvaughn: > perhaps `componentDidStaticRender`, to signify that there will be no more render's after this?

There could be an additional render if `setState` is called. This is like the old react-router `Match`/`Miss` use case.

> `componentDidInitialize`

I think this name is too ambiguous. (What does it mean to "initialize"? Does that happen before or after render? Does it happen on the server or the client? etc)

patrys: `applicationDidRender`?

bvaughn: I think the `component` prefix is pretty standardized for lifecycle hooks 😄 

"did render" could work if we wanted the method to be invoked after render on both client _and_ server. I suppose that could be useful too, although it's easy enough to "share" logic between cDM and cDSR via a helper method.

patrys: It is but maybe it would be useful to have a trigger that is fired only once and only after the entire tree is rendered (so after the last component would be mounted) and immediately before it gets dehydrated for transport?

text/0000-component-did-server-render.md
```diff
@@ -40,7 +40,10 @@ class Example extends React.Component {
 
 # Motivation
 
-Provide a place for initialization logic with side effects to safely run on the server.
+Provide a place for initialization logic with side effects to safely run on the server. This includes things like:
+* Logging/metrics
+* Routing redirects if match was found (eg the legacy react-router `Match`/`Miss` components)
```
gaearon: Can you point me to the code examples? I'm still finding it hard to understand how those components worked (I'm not sure which version is the "legacy" one). Was it through context?

text/0000-component-did-server-render.md
gaearon: `componentDidFlush`?

evan-scott-zocdoc: I like `componentDidFlush`, but it could be confusing in that often times pushing changes to the DOM are referred to as "flushing to the DOM".

This might be a little vague too, but what about `componentDidOutput`?

gaearon: I guess also, if re-rendering is supported, `DidFlush` isn't correct since we haven't actually flushed yet.

mbrevda: `componenetDidInitalRender`? This is not server specific, and can also be used for things like rendering a view with cached data and fetching more up to date data only after something was rendered to the user

gaearon: You (probably?) don't want to trigger data fetching on the server side though. Since the output has already flushed.

bvaughn: Dan's right. The proposed lifecycle hook (like the others) is sync, and since it's intended for server rendering, there's nothing left to update after an async request.

That's another argument (in my mind) for keeping an explicitly separate did-server-render (server) and did-mount (client) separate. Less confusion about which is appropriate for async requests.

gaearon: The “server” part is still misleading IMO. The output doesn’t have to be served. You can use it on the client for rendering into HTML.

gaearon: `componentDidRenderToString`?

bvaughn: I like this suggestion!

gaearon: Here’s a few more:

`componentDidSerialize`
`componentDidStream`

bvaughn: I think "serialize" is too ambiguous. (I like the word "render" being in the name. I think it's much clearer.)

Similar concern about "stream". (Maybe people would think it only works for `renderToNodeStream`)

gaearon: `componentDidPrint` 🙃

gaearon: My problem with `componentDidRenderToString` is that it’s not like all the others (more than a verb) and implies `componentDidRender` also exists.

bvaughn: That's a benefit of `componentDidServerRender` 😁  (No implication of `componentDidServer` 😆 )

gaearon: You can still think `DidRender` or `DidClientRender` exists based on it.

text/0000-component-did-server-render.md
```diff
@@ -0,0 +1,82 @@
...
+
+Typically, `componentWillMount` is used for this, but [RFC #6](https://github.com/reactjs/rfcs/pull/6) proposes to remove this method. (It is not a very ergonomic solution anyway.) The logic could be moved to the component constructor but it's generally considered bad practice for a constructor to have side effects.
+
+# Detailed design
+
+## `componentDidServerRender()`
+
+`componentDidServerRender()` is invoked immediately after a component is server rendered. It is the server equivalent to `componentDidMount()`.
+
+Calling `setState()` in this method will trigger an extra rendering, but it is guaranteed to flush during the same tick. This guarantees that even though the `render()` will be called twice in this case, the user won't see the intermediate state.
```
gaearon: Can you explain how this works with streaming (its current implementation)? Do we disable streaming if there’s a component with this method anywhere in the tree above? I don’t see how we could “retract” what we’ve already written to the client. Same reason we didn’t implement error boundaries.

text/0000-component-did-server-render.md
gaearon: I would assume that gets called on the client.

yordis: I am really confused why this exists and why would be on React when I will be using React in different platforms ....

Can somebody explain me what is the current issue and why we need this a little be more verbose? Sorry for my ignorance!

bvaughn: Wouldn't `componentDidHydrate` just be `componentDidMount`?
And `componentWillHydrate` would be like `componentWillMount` (which we hope to deprecate)?

text/0000-component-did-server-render.md
bvaughn: Good question, Dan. I'm not familiar with our streaming implementation. Sounds like maybe this just _wouldn't_ work with streaming.

text/0000-component-did-server-render.md
gaearon: @yordis

In theory you could build a prerenderer for React Native, and then hydrate an RN tree. I don’t see it as a web specific feature. We just need to figure out the naming that implies we’re flushing to an ephemeral serializable output instead of to a living and breathing tree that changes over time.

bvaughn: @yordis Read the issue the description reads to: https://github.com/facebook/react/issues/7671

ljharb: @bvaughn yes, fair. altho `componentDidHydrate` would only be called if `ReactDOM.hydrate` was used; `ReactDOM.render` wouldn't invoke it.

text/0000-component-did-server-render.md
gaearon: Oh. I see. 😬 I’m fairly sure it wouldn’t then.

My understanding was that we wanted to revisit how SSR works in general eventually and restructure it around the idea of pagelets. That’s when I thought we’d look into APIs like this as well as support error boundaries. But I don’t see how it could be compatible with the current model.

text/0000-component-did-server-render.md
bvaughn: If there's a useful use-case for a did-hydrate lifecycle hook, that would be a nice follow-up RFC (that maybe you could file @ljharb? 😁 ). I don't know of a use case for it.

yordis: Thanks a lot folks!

I like the `Hydrate` which is what is doing no?

`componentWillHydrate`
`componentDidHydrate`

text/0000-component-did-server-render.md
bvaughn: Yeah. @sebmarkbage has been thinking about this stuff recently. I'll defer to him on this question. 😄

gaearon: I think it might be premature to name the method until we’ve decided how it works (including streaming). Since doing that later might expose an inconsistency with naming.

bvaughn: Sure, although Seb proposed the name and is also the one who's been thinking about the pagelet stuff, so...I'll still defer to him here.

gaearon: Maybe it’s with looking at it in tandem with other APIs Seb proposed? Our current naming scheme by @leeb was very well thought through and internally consistent. That’s because all lifecycles were named together. I’m worried separate proposals for each will hurt this consistency and cost us the ability to redesign them to make sense together.

bvaughn: Sure!

Seb specifically suggested I propose this one in tandem with #6 😄  but that may have been just to get the discussion rolling.

thysultan: Are there reasons that this needs to be called after the `render` method? For example it could be called before the `render` similar to `componentWillMount` and still allow `setState` calls to update state synchronously with a name along the lines of `componentWillSerialize`.

gaearon: I think the idea was to afford the use case where you let the children render (and possibly execute their own side effects) and *then* decide to render something else. Calling it before render makes this impossible.

That said again, the proposal is incompatible with our implementation of streaming. So it’s not clear to me if we want this part (yet?) or not.

bvaughn: > I think the idea was to afford the use case where you let the children render (and possibly execute their own side effects) and then decide to render something else. Calling it before render makes this impossible.

Yeah. I'm writing a (maybe kind of contrived) example now that illustrates a use case like this.

Also, the general trend of RFC #6 is to move away from the early will-* lifecycle hooks because of their added complexity in async.

aickin:
I don't see anything currently in this proposal that addresses the fact that it seems to be moving the lifecycle call from before render to after. Have you (or someone else) done research into how `componentWillMount` is currently used by a range of SSR'ed components? I don't have any particular examples, but it really wouldn't surprise me if some of the usages of `cWM` depended on the fact that they were pre-`render`. 

And is there a way to port components that need a lifecycle call before the call to `render`? Just move the code to the constructor, maybe?

bvaughn:
> I don't see anything currently in this proposal that addresses the fact that it seems to be moving the lifecycle call from before render to after.

I don't think this is quite accurate. This proposal is to add a _new_ method- basically a sibling to `componentDidMount`, but one that actually runs on the server.

The reasons `componentWillMount` is problematic are covered in depth on RFC #6. That being said, you could still do pre-`render` things on the server side you if you want using either:
* The constructor
* `componentWillMount` (or `unsafe_componentWillMount`)
* The `render` method itself

jaredpalmer:
This would be awesome. Just noting that this is (partially) possible to accomplish right now with React Router. 

```js
// the highest leaf node (App.js)
  componentWillUpdate(nextProps) {
    if (nextProps.history.action !== 'POP') {
      this.previousLocation = this.props.location;
    }
  }

  render() {
    const initialRender = !!(this.previousLocation === this.props.location);
    return (
      <Switch>
        {this.props.routes.map((r, i) => (
          <Route
            key={`route--${i}`}
            path={r.path}
            exact={r.exact}
            render={props => {
              return React.createElement(r.component, {
                initialRender,
                ...props,
              });
            }}
          />
        ))}
      </Switch>
    );
  }
}
...
```

text/0000-component-did-server-render.md
arunoda: Input based on Next.js:

The naming and the usage of this should depends on what we are doing here. 

Is there any blocking stuff here? Then are we going to return a promise?
(So a function starting with getXXX makes sense)

I don’t like running setState inside this as it might introduce unpredictable states and loops inside the server. 

Here’s our proposal:

* We hope the idea of this hook to init some data for the component in the server. 
* So what happen if this component initialize on the client? How to fetch that data?
* The best idea is to have a static function which accept intial props and return a promise. 
* React will resolve that promise and use it as the props for the component. 

We have used something similar for Next.js top level pages and it’s a success. 

If we can have this for every component, that will be amazing.

steida:
I would not add special lifecycle method depending on something internal from React (server rendering, or any else kind of rendering) without specifying what is so special. But of course, we should follow real use cases. Naming is weird, what server render is anyway? Isn't it "first render" to fetch data? I believe Next.js solved it properly with getInitialProps. Maybe React could incorporate something from Next.js, and that would be good enough IMHO.

text/0000-component-did-server-render.md
jaredpalmer: I agree with @arunoda.  `setState` on the server feels weird because there isn't state on the server (unless we want there to somehow be).

On the client, would you still block render as well (like Next.js does)? Could a loading component (maybe with an api like ErrorBoundary) be added?

aickin:
@bvaughn I realized in my previous comment that I forgot to thank you for your work and diligence in responding to comments. Thanks!

> you could still do pre-render things on the server side you if you want using either:
>
>* The constructor
>* `componentWillMount` (or unsafe_componentWillMount)
>* The render method itself

Wait, now I'm really confused. When I asked about the fact that RFC 6 was proposing deprecating `componentWillMount` and replacing it with a method that wouldn't be called on the server, I thought I was told that that would be addressed in this RFC. I assumed this RFC would be a server-side replacement for `cWM`. But now it seems like maybe you are saying that `cWM` maybe is still kosher on the server? I think I must be misunderstanding something.

To be clearer, I guess my practical question is: let's say I have a currently SSR-ed component that uses `cWM`. If this RFC **and** RFC 6 both go into effect, how should I upgrade the component to not use deprecated methods? 

It sounds like the foolproof answer is "move all the code in `cWM` into the constructor", and the more complicated answer is "move all client-side code in `cWM` to `prerender`, and then determine via introspection if the server-side code in `cWM` can run after `render`. If so, move that code to `componentDidServerRender`. If not, put it in the constructor." Is that right, or am I reading these RFCs all wrong?

text/0000-component-did-server-render.md
bvaughn: I added some examples to the markdown file FWIW. I don't know how helpful or compelling they are though 😄  I've reached out to several OSS people asking for additional use-case examples from their real-world projects though. I'll update this section as I get info.

text/0000-component-did-server-render.md
arunoda: Anyway, it seems like the idea of this this RFC is mainly for side-effects. 

If so, my ^^^ comment should go into a different RFC. 

But promise support for this hook will be great.

bvaughn:
@steida Thanks!

> I would not add special lifecycle method depending on something internal from React (server rendering, or any else kind of rendering) without specifying what is so special.

The thing that's "special" in this case is that (1) a full tree render has occurred but (2) there is no client environment so things like refs can't be used to access eg browser APIs.

> But of course, we should follow real use cases.

I've attempted to list a few real use-case examples FWIW. I'd also welcome others to add their own.

* * *

@aickin You're welcome! And thanks to you and others for the helpful feedback.

> But now it seems like maybe you are saying that cWM maybe is still kosher on the server? I think I must be misunderstanding something.

I understand the confusion! Sorry! I'm actually just saying that cWM will still be around (even if it's renamed with an "unsafe" prefix) but we don't recommend using it (for reasons mentioned on #6).

> To be clearer, I guess my practical question is: let's say I have a currently SSR-ed component that uses cWM. If this RFC and RFC 6 both go into effect, how should I upgrade the component to not use deprecated methods?

Depends on the use-case. Short answer would be to just use `componentDidMount` and `componentDidServerRender` instead of cWM unless you _can't_ for some reason 😄

text/0000-component-did-server-render.md
bvaughn: Thanks @arunoda!

To answer your questions- based on the current line of thinking:

> Is there any blocking stuff here? Then are we going to return a promise?

No. This lifecycle would synchronous, like all others.

A few people have mentioned the idea of returning promises (on both this RFC and #6). The scope for that feels much larger though.

> I agree with @arunoda. setState on the server feels weird because there isn't state on the server (unless we want there to somehow be).

`state` does exist on the server. It gets initialized during construction can be referenced in `render`. (Without this, components would be really complicated to write.) It's just that `state` on the server is short-lived and not responsive to user-input.

> I don’t like running setState inside this as it might introduce unpredictable states and loops inside the server.

I'm not sure what about this proposal would make components unpredictable or likely to cause loops (any more so than is currently possible on the client). Do you mean elaborating?

aickin:
> Depends on the use-case. Short answer would be to just use componentDidMount and componentDidServerRender instead of cWM unless you can't for some reason 😄 (in which case let's talk more about why)

Two thoughts:

1. This is why I asked if you or others had done any research into how SSR-ed currently use componentWillMount, akin to the experiment you mentioned in RFC 6. I don't know if such use cases exist, but it seems like a question that others in the community might know the answer to.
1. From this comment, it sounds like this RFC will not be dependably codemod-able. Is that right? If so, I think that should be called out directly in the RFC, similar to how you discussed codemod-ability in RFC 6.

bvaughn:
> This is why I asked if you or others had done any research into how SSR-ed currently use componentWillMount, akin to the experiment you mentioned in RFC 6. I don't know if such use cases exist, but it seems like a question that others in the community might know the answer to.

Facebook doesn't really use React's server-side rendering so there's not a lot of internal research I can do. I have looked at some popular open source libs and I've also reached out (via this RFC and via Twitter) to their maintainers asking for input.

> From this comment, it sounds like this RFC will not be dependably codemod-able. Is that right? If so, I think that should be called out directly in the RFC, similar to how you discussed codemod-ability in RFC 6.

This is a _new_ method so no codemod is necessary (or really possible).

bvaughn:
After chatting (offline) for a while with @sebmarkbage about this, I'm going to narrow the scope of this RFC in the following ways:
* Change the newly proposed lifecycle to a _static_ method (which will remove support for `setState`).
* Rename it to avoid confusion with the current componentWill* and componentDid* _instance_ methods.

I'll upload an updated copy of it in the morning. In the meanwhile, maybe we can put the discussion on pause to avoid confusion. My apologies for the trashing!

aickin:
> This is a new method so no codemod is necessary (or really possible).

Right, but in combo with RFC 6, you intend to remove functionality. I should have said that this RFC _plus RFC 6_ would be un-codemod-able. Or maybe more precisely that RFC 6 is un-codemod-able on its own because of SSR concerns, and that this RFC doesn't fix that problem. Is that fair?

text/0000-component-did-server-render.md
```diff
@@ -30,10 +31,14 @@ This RFC proposes to add a new lifecycle hook named `componentDidServerRender` t
 class Example extends React.Component {
   componentDidServerRender() {
```
yordis: `componentBeforeMount` ?

bvaughn:
@aickin It makes sense, although RFC #6 _will_ be code-modable (b'c I'll provide a codemod for renaming componentWill* with the "unsafe_" prefix) but... yeah. 😄 I understand what you're saying.

text/0000-component-did-server-render.md
bvaughn: Looks like this is going to become a static method (look for an RFC update tomorrow) so the name will likely change to _not_ begin with the words "componentWill" (because I think that might be confusing given the instance method names).

bvaughn:
I've updated the RFC and examples.

bvaughn:
> An “is” prefix tells me it’s a predicate - ie, it returns a boolean. Why not keep with the component prefix?

I don't understand. The new proposed name is `instanceMounted`.

As for why I'm proposing to _not_ use the `component` prefix - I think it would cause confusion with legacy/existing _instance_ methods.

ljharb:
Oh, my bad, i read “isMounted”. “instanceMounted” seems like it’d be an instance method to me, too :-/

bvaughn:
No worries 😄  Naming is hard. I'm open to other naming suggestions.

gaearon:
I do agree naming is confusing although I don't want to derail the proposal before other things are considered.

It's confusing because "mounting" strongly implies it's client-only (since `DidMount` is client-only). The "instance" is also a bit misleading because you don't actually *get* that instance anyhow. Finally, we never used the past tense in component API before (`mounted`).

bvaughn:
I spent only about 30 seconds considering the name and just picked this one so I could move forward with the rest of the proposal. 😄 

`static instanceRendered` would be fine. I don't really think "instance" in the name implies that you get the instance, but I don't feel strongly about it.

gaearon:
>I spent only about 30 seconds considering the name

I thought so, that’s why I didn’t comment on the name at all initially. :-) If you’re ready to bikeshed it I’m ready too though!

>static instanceRendered would be fine.

Again, my minor concern is we never used past tense in method names before. Why do we have `componentDidMount` but `instanceRendered`? I would expect us to use `componentMounted` then.

gaearon:
It is also odd that `component*` names live on the instance but `instance*` names (currently just one) live on the component.

gaearon:
>static instanceRendered would be fine. 

I would expect it to fire on the client.

j-f1:
How about `componentWillSerialize`? `mount` would reference installing the component in some sort of user-visible UI (mobile app? web page? desktop app?), while `serialize` would reference converting the component into some form of static object for transport.

gaearon:
It needs to be different enough to be clear it's not an instance method.

j-f1:
What are the benefits of it being a static method as opposed to an instance method?

TrySound:
@j-f1 You do not have an access to `this`. It forces purity.

j-f1:
How about `instanceWillSerialize` then? Same rationale as in my previous comment.

kof:
## Naming ideas:

- `getDerivedStateFromNextProps ` => `mapState`

Like Array#map but for state. Fixes:
  - fromNextProps implies that return value will be derived from next props, though in fact user can build it from anything else, because it receives `nextProps, prevProps, prevState`.
  - more readable in my opinion
  - more flexible (good and bad) 

- `optimisticallyPrepareToRender` => `onBeforeRender`

Fixes:
1. "optimistically" Verbosity which isn't explaining much. Optimistically can mean anything here, without docs I still don't know what that means.
2. "prepare" sounds to me like that method should return something or should update something immediately, but it doesn't

TrySound:
@kof I believe you meant this RFC https://github.com/reactjs/rfcs/pull/6

kof:
@TrySound thanks, wrong tab, moved

victoriafrench:
I would like to suggest the name of `componentDidInitialize`. It stays with the naming convention and sort of makes sense for the purpose it is used for. I mean people are basically placing initialization code on `componentWillMount`. It also leaves open the possibility for a pre and post hook, though I am not sure why one would ever be needed.

kof:
@victoriafrench initialize implies it happens just once, willMount happens though each time after unmount.

djfarly:
Thinking about it I kinda like the thought of naming it *willSerialize or something similar. The reason being, that it focused on what will happen next, and not what happend on the way there.
Serialize (or Stringify maybe?) also point towards the method only being called on the server, because that will never happen on the client (until renderToString is called).

One question though - the application of this seems to be mostly analytics? Are the any other significant real world uses? (Now that the this and setState are gone)
I find it kinda hard to get my head around what it will enable.

bvaughn:
Regarding the proposed `componentDid*` and/or `componentWill*` names: There was some discussion about about avoiding these prefixes to avoid potential confusion with the various instance lifecycle hooks. The proposal included this prefix _initially_ because it was initially a proposed instance method. Now that the proposal is for a static method, it's probably nice to choose a name that's clearly separate.

> One question though - the application of this seems to be mostly analytics? Are the any other significant real world uses?

That's correct. I'm not really considering other use-cases currently.

ljharb:
On the server, `componentWillMount` only ever runs once; a lifecycle method specifically for "about to server-render", or a new one for "just finished server-rendering" - both of which would only ever run once - would make sense.

bvaughn:
@ljharb We intentionally want to direct traffic away from "about to render" (server or otherwise) for reasons outlined on RFC #6. This proposal originally suggested an _instance_ method to be run "just after server-rendering" but this was updated to be a static method because it would not have supported eg `setState` and so it didn't really make sense to define on the instance.

arunoda:
@bvaughn the use of a static method is pretty good. (prevents exposing public APIs)
But this feature really doesn't add anything to SSR.

static server-only post-render lifecycle method

^^^ is not useful for any SSR app.

In SSR, it's about data fetching. I don't know how to fetch data from a remote service from this API.
Logging is a one use case, but it's not SSR and not many will use it.
(In a SSR app, you mostly need to log routes in a running app. We can do it in the routing layer)

I'm sorry to say, but this RFC doesn't add any value to React in the context of SSR

If we need to support a SSR here's my suggestion

static hook just like this: getInitialProps
It will accept props passed to the React element
Then it'll return a promise which resolves an object
React render the component using ^^^ object as the props
This is just the start, there are many questions we need to address if we are interested in something like above.

sophiebits:
@arunoda This RFC is not designed to make data fetching better so it’s no surprise that it doesn’t. We’re aware that async data fetching during server render is a pain point for some people but it’s not what we’re trying to solve in this thread so let’s try to stay on topic.

It’s OK if you don’t have the problem that this addition solves — just don’t use this feature! If you want to submit an RFC about getInitialProps then feel free and we can discuss it there.

arunoda:
@sophiebits 

Sure'll open a new RFC

> It’s OK if you don’t have the problem that this addition solves — just don’t use this feature! If you want to submit an RFC about getInitialProps then feel free and we can discuss it there.

**just don’t use this feature.**

arunoda:
@sophiebits

Sure'll open a new RFC.

> It’s OK if you don’t have the problem that this addition solves — just don’t use this feature! If you want to submit an RFC about getInitialProps then feel free and we can discuss it there.

**just don’t use this feature.**
Of course.

I know this is going to introduce in the favor of the remove of `componentWillMount`

And this is to be used in the Server.
So, tell me a use case where where React will be used in the Server other than for SSR?

If we introduce a new API, it should add value over the existing.
So, does this?

sophiebits:
We’re not removing componentWillMount, but the semantics will change slightly in the future, so it can’t have side effects. This RFC is for a new method used for SSR that would be an alternative for the use cases when people are currently logging in componentWillMount on the server. Since componentWillMount will no longer be suitable for this use case. It does add value.

bvaughn:
The lifecycle proposed by this RFC is for a _very specific_ purpose (mentioned in the [RFC examples section](https://github.com/bvaughn/rfcs/blob/component-did-server-render/text/0000-component-did-server-render.md#logging-example)).

It's possible that it may also be useful for some other edge-case things done by component libraries. (I don't know.) Part of the reason for posting the RFC is to see if others can think of additional ways it could be used.

The primary reason I've outlined though isn't really a common one, and it doesn't sound like it's one you need or use, which is okay. 😄

What you're describing sounds much more complex- (although also probably useful to way more people than this small RFC). So by all means, please feel free to start a new discussion around blocking/async server rendering support in a new RFC! 👍

arunoda:
@bvaughn @sophiebits got it.

taion:
What would be the drawback of making the proposed callback fire even in the client-rendering case?

It seems easy enough in userspace to make things only run when doing server-side rendering. I don't even mean doing `typeof window` – I can even explicitly look at some global-ish flag.

As such, regardless of the specific semantics of the new hook in timing, static-ness, &c., it still seems like it would be cleaner to make the execution of the hook unconditional, and let determination of SSR-ness happen in userspace.

taion:
As a strawman, consider something like a `componentDidRender` hook.

On the server, this would be called exactly once, after the initial render. On the client, this would be called after every render – that is, before each invocation of `componentDidMount` and `componentDidUpdate`.

Obviously this is a bit problematic for reasons of redundancy, and out of the requirement to block `this.setState` on the server, but it seems a bit more "obvious" for lack of a better term. It's more clear when this will run, and it isn't extremely specific the way a server-rendering-only hook would be – i.e. it means there wouldn't be a lifecycle hook that would be completely irrelevant for React Native, say.

A bit tangential here, but it'd also clean up user code that needs to reconcile non-React DOM things, e.g. all the cases where you just call into the same DOM manipulation functionality from both cDM and cDU.

Ephem:
I thought I'd add some anecdotal feedback for how SSR-apps use componentWillMount currently. Based on my experience (and a quick search through 5 medium-sized apps I work with) it is mostly used in erroneous ways, for things it shouldn't (like dispatching actions and other side-effects that RFC 6 tries to prevent). I saw no valid use-cases, at least that could not (and will!) be refactored.

This means I approve both of marking it unsafe but also that for **us** this RFC does not *currently* solve a use-case. I definitely don't mind the RFC and I can see how logging could be one such use-case though.

Ephem:
That being said, I do think @taion brings up some valid points. If I did have a use-case for doing something after render that did not involve things like dispatching an action, doing something with the DOM etc, there is a good chance I would want to do it both on the server and on the client. The current proposal would entail triggering that both in cDM and in instanceMounted and while not a big deal that seems unnecessarily confusing.

Being naive about the implementation and future plans my only question about this proposal is this: Will it in any way hinder/make harder the future development/refactoring of the SSR-functionality or React in general? If that is the case, it would be nice to see some more real world use-cases that this solves before deciding if adding the complexity is worth it. :)

mxstbr:
Just another datapoint: we'll likely need this lifecycle hook to switch to the new lifecycles in styled-components, otherwise we can't make server-side rendering work safely. Ref https://github.com/styled-components/styled-components/pull/1664

bvaughn:
Cool! Thanks for sharing another use case Max 😄

gaearon:
Seems like the interest for this has waned and there have been no new use cases. I'm taking the liberty of closing it until we hear about the need again with more concrete examples.

PR #9, opened 2017-12-15
New rfc

giorgim:


j-f1:
Can you revert your first commit (https://github.com/reactjs/rfcs/pull/9/commits/a0a8563c8af5c7fe2c4e6f764b37412339a99f0d) please? I think they’d like to keep the template for future RFCs.

giorgim:
@j-f1 what will happen with rfc doc I created then in that case? (I am not very seasoned in git)

Attrash-Islam:
Async state one of the major performance boosts in the React, When I asked about this in the past my request was rejected from Dan Abramov by explaining to me how should I correctly update the state, then I didn't faced any problems with the state management

kyleshevlin:
I'd like to push back on this RFC just a bit. Async state is a huge performance boost to React overall. A synchronous state setter would block the thread, and that's not something we really want to do. You may have many other components competing for resources at the same time and being able to defer computation is a big win.

I will agree that `setState` has been confusing for many people, but there are very few quirks in React to know about. I think knowing that `setState` is async and has two function signatures is an acceptable quirk to have. Learning how to use these two signatures appropriately is important for React devs and honestly something I believe every person using React should know. Personally, I'd rather improve education on the topic than create a third method and force people to understand all three.

Lastly, not many computations in `setState` should be synchronous in general. You should be able to write your functions in such a way as to derive your future computations based on the current state and what you wish to do. If, you need synchronicity, this is where the second argument callback is useful, as it will essentially act as a means of making the nextState available to you since we know it will have been computed by the time the callback is fired. If you are chaining many computations in the callbacks of `setState`, it may be a code smell. There is probably a better architecture available to solve your problems.

david-jam:
referring to setState callback which @kyleshevlin mentioned.
In this example
```

 constructor(props)
  {
   super(props);
   this.state={val:0}
  }
  
  click(){
    this.setState({val:1}, ()=>{console.log(this.state.val)})
     this.setState({val:2})
  }
```

Say you want to get value on console but are interested in that only when you assigned 1 to value - hence the callback. What will be output in your opinion of above code? answer: 2.

So is this not confusing that user might have thought that hey in `setState `I am assigning value 1, now I specify a function in callback which should fire when the value is set to 1. However, in reality that function is called when value was assigned 2. 

Isn't that confusing? 

Maybe those are the kind of things OP referred to.

kyleshevlin:
@david-jam I meant that you could **nest** another setState within your callback. The way you have written that, someone who knows and understands React would know that the objects `{ value: 1 }` and `{ value: 2 }` would be merged into one object before the update occurs.

david-jam:
@kyleshevlin then it is a bit confusing why tie a callback with a particular function invocation if that callback might not be called when the function change takes effect e.g. like with
`    this.setState({value:1}, ()=>{console.log(this.state.val)})
`
I am just saying naively a user might be expecting hey, I called set state and this is callback I want to fire when the value I specified in set state is set. otherwise why not just use componentDidUpdate?

TrySound:
@david-jam Using setState callback should not be naive. It's advanced feature which is not required in regular code (I used it twice in large applications).

david-jam:
@TrySound so you say it is not possible to arrive at a gotcha situation with current approach of async react state management? (that you would not arrive with a sync state). I am asking out of curiosity not criticizing someone or smth.

TrySound:
@david-jam I wouldn't say it's difficult, however it requires some knowledge about how API works. That's I meant as "naive". Just follow rules and nobody will get hurt.

david-jam:
@TrySound even if you know how API works I think it is beyond doubt that it is easier to manage app state when that is synchronous.

TrySound:
Easier !== better. Programming is hard AFAIK. But if you know things you CAN work with them.

david-jam:
@TrySound of course easier is better in this case, why introduce unnecessary complexity? Anyway we are heading in a different direction now so I will finish of my discussion here with this comment. 
I don't have specific example which would show that in some situation using async set state could cause bug - or would not be suitable for some situation. If such cases don't exist, good for react. But I am not sure.

TrySound:
Here's the thing. It's necessary complexity for future react features.

taion:
The author of this RFC is apparently now a @ghost. Does this have any bearing on the RFC process?

gaearon:
It shouldn’t affect the initial review but if we don’t reject it and ask for a revisions then we’d need somebody to take over (assuming the author doesn’t want to).

TrySound:
@gaearon I guess there's nobody agreed with the author since there's a proposal from you and other react maintainers about async rendering.

gaearon:
I wrote a bit about why React works this way: https://github.com/facebook/react/issues/11527#issuecomment-360199710

theKashey:
We have used something like React and something like Redux, but with sync API. What was wrong?
 - You are calling setState, 
 - then it will call render with the new props, 
 - then children's componentWillReceiveProps(or mount/unmount) dispatches an event,
 - then a parent of the initial component will propagate new props down, reflecting the change,
 - then you may call setState once again. Why not?
Sooner or later executing point will reach the end of the initial setState call, and run some code after it. The only problem - you just time-travelled back and doing something inside a component which does not reflect the reality. And you may face 2-3-4-5 past versions of yourself. Loopers, you know.
That is the first rule of a time traveller - don't meet yourself, or the universe will collapse.

So please don't destroy the universe, and use async version of a setState.

gaearon:
To sum this up:

1. React state is intentionally set asynchronous: https://github.com/facebook/react/issues/11527#issuecomment-360199710. This is part of the reason React is fast enough for most use cases. We don't want to regress on that.
2. For corner cases where you need to force a sync render due to some browser API, there is `ReactDOM.flushSync(() => { /* call setState here */ })`.
3. Regardless, with Hooks setting state synchronously still wouldn’t update the captured values, negating the original motivation.

Therefore, closing.

PR #10, opened 2018-01-03
Promise-returning setState and forceUpdate

jamiebuilds:
[[Rendered](https://github.com/thejameskyle/rfcs-2/blob/patch-1/text/0000-promise-returning-set-state-and-force-update.md)]

This RFC specifies `setState` and `forceUpdate` to return a promise when no callback is provided.

```js
class DataFetcher extends React.Component {
  state = {
    loading: false,
    error: null,
    data: null,
  };

  async componentDidMount() {
    await this.setState({ loading: true });
    try {
      let response = await fetch(this.props.url);
      let data = await response.json();
      await this.setState({ data, loading: false });
    } catch (error) {
      await this.setState({ error, loading: false });
    }
  };
  
  render() {
    return this.props.render(this.state);
  }
}
```

> Note: This example doesn't rely so much on `setState` being applied before continuing. I can come up with a better example if people want it better demonstrated. I figured the data fetching example was easiest to understand.

### DIFF
--- a/text/0000-promise-returning-set-state-and-force-update.md
+++ b/text/0000-promise-returning-set-state-and-force-update.md
@@ -1,0 +1,113 @@
+- Start Date: 2017-01-03
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This RFC specifies `setState` and `forceUpdate` to return a promise when no
+callback is provided.
+
+# Basic example
+
+```js
+class DataFetcher extends React.Component {
+  state = {
+    loading: false,
+    error: null,
+    data: null,
+  };
+
+  async componentDidMount() {
+    await this.setState({ loading: true });
+    try {
+      let response = await fetch(this.props.url);
+      let data = await response.json();
+      await this.setState({ data, loading: false });
+    } catch (error) {
+      await this.setState({ error, loading: false });
+    }
+  };
+  
+  render() {
+    return this.props.render(this.state);
+  }
+}
+```
+
+Type definition:
+
+```js
+type Updater<State> = Partial<State> | (prevState: State) => Partial<State>;
+
+declare class React.Component<Props, State> {
+  setState(updater: Updater<State>): Promise<void>;
+  setState(updater: Updater<State>, callback: () => mixed): void;
+  forceUpdate(): Promise<void>;
+  forceUpdate(callback: () => mixed): void;
+  // ...
+}
+```
+
+# Motivation
+
+- Promises are JavaScript's primitive for representing async activity
+- Async-await is significantly easier to use than callback-based APIs
+- Lots of new and existing APIs in the ecosystem and standards are moving towards
+  promises (ex: `fetch()`)
+- As async-await becomes more available it will make more and more sense build with
+- Developers are using `setState` as if it were sync today in lots of places which
+  end up needing to be refactored later on when new code needs the state to be applied before running
+
+# Detailed design
+
+When called with a callback, `setState` and `forceUpdate` will continue returning
+`undefined`, when called without a callback it will return a promise which resolves
+at the same time as the existing callback API.
+
+```js
+this.setState(updater, callback); // >> undefined
+this.setState(updater); // >> promise
+```
+
+The reason for different return values is to avoid messy semantics supporting
+the callback and promise simultaneously:
+
+1. Would the callback be called before the promise resolves?
+2. Does the promise wait for a promise returned by the callback?
+3. What happens when the callback throws synchronously?
+
+Instead, by only supporting promises when a callback is not provided, React
+side-steps the problem entirely. There doesn't seem to be a use case where someone
+would want both anyways.
+
+# Drawbacks
+
+- If `setState` without the callback is (or could be) optimized in anyway (due to not
+  needing the schedule the callback or something), it wouldn't be able to anymore because
+  the promise would always have to be returned.
+- More API surface area (even if only temporarily)
+
+# Alternatives
+
+Unknown
+
+# Adoption strategy
+
+By supporting both the existing callback API and the promise-returning API, this
+feature can be introduced in a minor version and supported indefinitely. If desired
+the callback API can eventually start logging warnings and eventually removed in a
+major version.
+
+# How we teach this
+
+(I believe) this would be the first place where promises are used within (major?)
+React APIs, so teaching promises in the documentation might be necessary. Otherwise,
+this API change doesn't change much so it can be taught the same way as it is today.
+
+If React does eventually want to remove the callback API, that can be communicated
+through deprecation warnings, blog posts, Dan Abramov tweets, etc.
+
+# Unresolved questions
+
+- Is `setState()` without the callback optimizable in any way that always having to
+  always return a promise would prevent?


text/0000-promise-returning-set-state-and-force-update.md
```diff
@@ -0,0 +1,113 @@
...
+
+Instead, by only supporting promises when a callback is not provided, React
+side-steps the problem entirely. There doesn't seem to be a use case where someone
+would want both anyways.
+
+# Drawbacks
+
+- If `setState` without the callback is (or could be) optimized in anyway (due to not
+  needing the schedule the callback or something), it wouldn't be able to anymore because
+  the promise would always have to be returned.
```
thejameskyle: > "Yes, we’d have to tag it as a node to visit when committing changes. setState() without callback doesn’t need that." – [Dan](https://twitter.com/dan_abramov/status/948552271264210949)

thejameskyle: Depending on how big of an impact this makes, I would consider this a big enough reason to reject this proposal outright.

ruyasan: It may be possible to detect if `setState` is declared async (which is not the same as returning a promise, true, but perhaps a good enough work-around?) - e.g. https://stackoverflow.com/questions/38508420/how-to-know-if-a-function-is-async

alexkrolick: Some promisified callback APIs (like the AWS SDK) use an explicit call  to `.promise()` to know whether to return a Promise or not. It's a little uglier, but more explicit.

```js
class DataFetcher extends React.Component {
  state = {
    loading: false,
    error: null,
    data: null,
  };

  async componentDidMount() {
    await this.setState({ loading: true }).promise(); // 🆕 pseudo-blocking with await
    try {
      let response = await fetch(this.props.url);
      let data = await response.json();
      this.setState({ data, loading: false }); // normal, optimized
    } catch (error) {
      this.setState({ error, loading: false }, handleErrorSomehow); // callback
    }
  };
  
  render() {
    return this.props.render(this.state);
  }
}
```

Kovensky: That still means that you'd have to return an object that has a `promise` method, and `setState` always has to call a hidden callback to notify the object just in case `promise` was called. It's a lot of extra bookkeeping that you have to do that likely won't gain you anything over just returning an unused `Promise` to begin with.

thejameskyle: The only option really would be to add a second method:

```js
await this.setStateAsync({ loading: true });
```

But that name sounds bad because it implies the other is sync

acdlite: Returning a thenable instead of an actual promise solves the problem Dan is alluding to in that tweet. Here's psuedocode to illustrate:

```js
// What we do now:
setState(update, callback) {
  scheduleUpdate(update)
  // scheduleCallback marks the internal node with a "callback" effect.
  // We can skip it if a callback is not provided.
  if (callback) {
    scheduleCallback(callback)
  }
}

// Using a promise:
setState(update) {
  scheduleUpdate(update)
  return new Promise(resolve => {
    // This is called regardless of whether the promise is ever used :(
    scheduleCallback(resolve)
  })
}

// Using a thenable:
setState(update) {
  scheduleUpdate(update)
  return {
    then(cb) {
      // Only called if `then` is called :)
      scheduleCallback(cb)
    }
  }
}
```

If you await a thenable, it's automatically wrapped in a promise and just works. If you don't await it, then the callback is never scheduled.

acdlite: You still have to allocate the `thenable` object. But that's likely not a huge deal.

j-f1: How would that work if I did something like this?

```js
this.thenable = this.setState({ foo: 'bar' })

// later (maybe in an event)
this.setState({ foo: 'baz' }, () => {
  this.thenable.then(({ foo }) => {
    // what is `foo`?
  })
})
```

alexkrolick: @j-f1 the callback doesn't take arguments, so presumably the Promise wouldn't resolve it either. You use it to schedule work after setState completes, not to get the new value (a reference to `this.state` is sufficient for that).

acdlite: @j-f1 What @alexkrolick said. Callbacks are like `componentDidUpdate` and `componentDidMount`. They fire during React's synchronization phase, so `this.state` inside a callback always points to the most current state.

thejameskyle: I think the concern is more:

```js
let thenable = this.setState({ ... });
setImmediate(() => {
  thenable.then(() => {
    // ...
  });
});
```

What if `.then` isn't called immediately (but is potentially called before it would have resolved)?

acdlite: Same as calling `this.setState(null, callback)`.

taion: Is that going to work as expected (i.e. the promise will resolve after the explicit `setState` finishes)? If so, could `.then()` always have those semantics?

If so, you could potentially only allocate at most a single thenable per component instance.

thejameskyle: > Same as calling `this.setState(null, callback)`

I don't see how that's possible unless `scheduleCallback` can be called after the update is already applied

acdlite: > unless scheduleCallback can be called after the update is already applied

Bingo :D

bpartridge:
This seems easy enough to provide in an external library, and therefore be entirely opt-in.

```js
export const setStatePromise = (o, state) => new Promise(resolve => o.setState(state, resolve));

await setStatePromise(this, { loading: true })

// or if you want syntax similar to the RFC, but still opt-in for performance

this.setStateAsync = setStatePromise.bind(null, this) // in constructor

await this.setStateAsync({ loading: true })
```

We should be careful with naming conventions as well, now that `this.deferSetState` is coming down the pipeline as well... if someone wants to have deferSetStateAsync (or, for what it's worth, an AbstractDeferSetStateAsyncFactoryFactoryBean), that's a naming decision that shouldn't be made in the core library :)

cloudever:
I think that proposal will cause a perfomance problems.

sebmarkbage:
One data point here. In our concurrent mode API where you want to do one setState first at a higher priority followed by a setState at lower priority we originally proposed something like deferring a callback (or just requestIdleCallback or setTimeout). However, that introduced an unfortunate async seam that could now introduce new imperative effects (like other setStates). That in turn causes the sequence to be scheduled in a non-predictable way.

The original example in the beginning of this RFC is actually a good example of this. The classic problem with code like this is that it's not resilient to new updates coming in - in the meantime. E.g. what if props updates? How do fetches that happen during update interleave with other updates happening during the async sequence?

For data fetching in general, we have the whole Suspense thing which can solve most of the canonical use cases for this. For the more general problem, I think we need something that defines a sequence up front instead of triggering imperative code along the way. That way if other imperative code adds a sequence, it will by default be added in full after the first sequence instead of some operations becoming interleaved. Then those can be combined to be executed in different priorities or collapse like our update queue generally works.

We've learned a lot since this RFC was originally posted. I'm now pretty skeptical that this exact pattern is the right way forward. It's (deceiving) simplicity is aspirational for any API to compete with though.

Niryo:
Can someone update about this RFC? This is a fairly easy and straight-forward suggestion, why is it still open? 
It should have been merged already (or closed) ages ago, is there anything left to discuss here?
I am trying to figure out if this whole RFC mechanism is really working, or is it just a time waster for the community, because it looks like that not even a single RFC had been merged, that was not suggested by a member of the core team itself. All the community suggestions are just left stale. 
Can someone from the React team confirm to us that this RFC process is still active and we are not wasting our time? (not this specific RFC, I am talking about the whole RFC process now).

jamiebuilds:
This rfc predates hooks, and I don't really think it's worth improving the class api anymore, I'm gonna close it

PR #17, opened 2018-01-25
RFC: New createRef() API for ref objects

TrySound:
A proposal for a new createRef api as a replacement for existing string refs.

Collected messages from this refs:

https://github.com/facebook/react/pull/11555
https://github.com/facebook/react/issues/11973
https://github.com/facebook/react/issues/10581

[Rendered](https://github.com/TrySound/rfcs/blob/2647d5da3781eacb57579001fdea3091d5f5f76d/text/0000-new-create-ref.md)

### DIFF
--- a/text/0000-new-create-ref.md
+++ b/text/0000-new-create-ref.md
@@ -1,0 +1,104 @@
+- Start Date: (fill me in with today's date, 2018-01-25)
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Replace string refs with similar functionality to how they currently work but without many of the issues that string refs currently have (a list of them can be found at [#1373](https://github.com/facebook/react/issues/1373)).  
+
+# Basic example
+
+The React.createRef() API will create an immutable object ref (where it's value is a mutable object referencing the actual ref). Accessing the ref value can be done via ref.value. An example of how this works is below:
+
+```js
+class MyComponent extends React.Component {
+  divRef = React.createRef();
+
+  render() {
+    return <div><input type="text" ref={this.divRef} /></div>;
+  }
+
+  componentDidMount() {
+    this.divRef.value.focus();
+  }
+}
+```
+
+# Motivation
+
+This alternative API shouldn't provide any big real wins over callback refs - other than being a nice convenience feature. There might be some small wins in performance - as a common pattern is to assign a ref value in a closure created in the render phase of a component - this avoids that (even more so when a ref property is assigned to a non-existent component instance property).
+
+One benefit would be more correct Flow types. People tend to type refs incorrectly because Flow doesn't enforce uninitialized class properties correctly:
+
+```js
+class Foo { neverSet: boolean; }
+let foo = new Foo();
+(foo.neverSet: boolean); // works
+```
+
+# Detailed design
+
+Introduces `React.createRef()`:
+
+```js
+type RefObject<T> = {
+  value: T | null
+};
+
+interface React {
+  createRef<T>(): RefObject<T>;
+}
+```
+
+`createRef` requires a explicit type annotation to ensure type correctness.
+
+After `componentDidMount` lifecycle `divRef.value` will be filled with element/component reference:
+
+```js
+componentDidMount() {
+  this.divRef.value instanceof HTMLDivElement === true
+}
+
+render() {
+  return (
+    <div ref={this.divRef}>
+      {this.props.children}
+    </div>
+  );
+}
+```
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people React
+- integration of this feature with other existing and planned features
+- cost of migrating existing React applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Adoption strategy
+
+If we implement this proposal, how will existing React developers adopt it? Is
+this a breaking change? Can we write a codemod? Should we coordinate with
+other projects or libraries?
+
+# How we teach this
+
+Callback and object refs solve different problems and crossing over in what they offer too - much like how components in React do currently.
+
+Someone would likely choose an object ref when they only care about binding the object reference to the component instance/state they're in. Then the ref can be used for caching, or being passed to another component/handler in a lifecycle event.
+
+Callback refs are great when someone likes to have find grain control over the node at the point where it gets provides and removed (like a subscription) to react to it.
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?


text/0000-new-create-ref.md
```diff
@@ -0,0 +1,104 @@
...
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Replace string refs with similar functionality to how they currently work but without many of the issues that string refs currently have (a list of them can be found at [#1373](https://github.com/facebook/react/issues/1373)).  
+
+# Basic example
+
+The React.createRef() API will create an immutable object ref (where it's value is a mutable object referencing the actual ref). Accessing the ref value can be done via ref.value. An example of how this works is below:
```
aoc: change `it's` to `its`

aulisius:
Hi,

How would this tie in with `this.refs` which is an iterable object? Would both ref APIs be supported? From what I gathered, the idea is to gradually replace string refs. Then, will this API provide such a method to implement the same?

Faizaan.

j-f1:
Alternate proposal:

```jsx
class Component extends React.Component {
  onClear() {
    this.refs.input.value = ''
  }
  render() {
    return <div>
      <input ref={this.ref('input')}/>
      <a onClick={this.onClear}>Clear</a>
    </div>
  }
}
```

`this.ref` has the same behavior as a string ref, but without many of the caveats:

```js
class Component {
  ref(name) {
    if (!this.ref._cache) {
      this.ref._cache = new WeakMap()
    }

    const { _cache: cache } = this.ref
    if (!cache.has(name)) {
      cache.set(name, ref => {
        this.refs[name] = ref
      })
    }

    return cache.get(name)
  }
}
```

trueadm:
@j-f1 I'd prefer if `createRef()` was not coupled to class components, plus your example doesn't look type safe as you're having to pass in a string to a function rather than access a property reference (which has the same problems as string refs).

Furthermore, I see a valid use case of `createRef()` to store refs on the `state` object for places where you might currently use instance variables.

TrySound:
@aulisius `createRef` produces a single value which will be mutated with single reference. So iteration is not the case anymore.

j-f1:
> I'd prefer if `createRef()` was not coupled to class components

That’s a good idea, but this could be syntactic sugar only available to classes.

Also, how would a functional component use refs?

> plus your example doesn't look type safe as you're having to pass in a string to a function rather than access a property reference

TypeScript:
```tsx
// P = props
// S = state
// R = refs
class Component<P, S, R = {}> {
  // ...
  protected ref<K extends keyof R>(name: K): (ref: R[K]) => void
  protected refs: R
}

// Usage:

interface FooRefs {
  input: HTMLInputElement
  dialog: Dialog
}

class Foo extends React.Component<{}, {}, FooRefs> {
  renderInput() {
    return <input ref={this.ref('input')} />
  }
  renderDialog() {
    return <Dialog ref={this.ref('dialog')} />
  }
}
```

(of course, this needs some more typings at the JSX level)

TrySound:
> That’s a good idea, but this could be syntactic sugar only available to classes.

I wouldn't add feature to remove it later just because it nice. Also react API always was small and simple. Feature duplicating is not good practice.

> Also, how would a functional component use refs?

Nothing magical. Just pass ref as any prop and child will patch it. In `componentDidMount` you will always have filled reference.

aulisius:
@j-f1 refs are not supported on SFC even now. Any reason that should change?

@TrySound I understood that. My question is, will there be an alternative to still be using the same pattern. As in,

```js
class ComponentUsingRefs extends React.Component {
  componentWillReceiveProps(nextProps) {
    if (nextProps.isDataLoaded) {
      for (let refKey in this.refs) {
        const ref = this.refs[refKey];
        // Do stuff with it
      }
    }
  }

  render() {
    return this.props.list.map(values => (
      <ListItem key={values.id} ref={`${values.id}`} {...values} />
    ));
  }
}
```

How would do this code be written with the `createRef` API?


Faizaan.

j-f1:
> Nothing magical. Just pass ref as any prop and child will patch it. In `componentDidMount` you will always have filled reference.

My proposal would work with a functional component, too:

```js
function FunctionalComponent({ myRef }) {
  return <div ref={myRef} />
}

class ClassComponent extends React.Component {
  render() {
    return <FunctionalComponent myRef={this.ref('child')} />
  }
}
```

If this isn’t what you meant, can you give an example using the API in this RFC?

streamich:
I kinda of like it, because it is kinda nice and convenient.

And I kinda dislike it, because it creates a 3rd interface for the thing we can already do in 2 ways.

TrySound:
@streamich It's a replacement for one those things. Also it's possible this feature can eliminate callback refs in the future.

sebmarkbage:
It doesn't fully replace callback refs since these refs don't have a way to respond to a child being added and removed by a container.

They're only useful if you don't need to perform any effect and can lazily handle the case when a ref has been unmounted and remounted. That's most cases, but not all.

There are also cases where you can safely assume that a child shares the same life-time as the parent but sometimes that isn't fully safe. I suspect that with new types of containers like "expiration boundaries" this type of thing will happen more often than it does today.

So I don't see this replacing callback refs.

gaearon:
In other words the idea is to relegate callback refs to be a "power user" pattern for less common cases. This API would be a bit less powerful than callback refs but less cumbersome to use for the common case. It would replace string refs but stay alongside callback refs.

TryingToImprove:
Would it make sense to also add a prop-type?

```
- PropTypes.Ref
- PropTypes.Ref.isRequired
```

TrySound:
@TryingToImprove Quite react specific. One of the case of splitting prop-types is reusing it with other libs. Refs can be achieved with simple `PropTypes.object.isRequired`. And you can always write custom prop type.

TryingToImprove:
@TrySound There are already `PropTypes.element`. If the ref, is becoming a DOM element could also be `PropTypes.DOMElement`

TrySound:
@TryingToImprove Legacy)

j-f1:
There’s also `PropTypes.node`.

j-f1:
How about `React.refPropType`?

TrySound:
@j-f1 Of course not. prop-types is "external" feature. A lot of users don't need this api since they are use static typing.

streamich:
Don't you need to use `prop-types` when using current context API.

TrySound:
@streamich I won't need it in the next week.

j-f1:
> prop-types is "external" feature

React still checks `propTypes` when you provide them.

gaearon:
PropTypes is an external project to React now. This RFC doesn't need to concern itself with PropTypes. We can discuss it separately in PropTypes repo.

gaearon:
We’re happy with this proposal and have an experimental implementation that will likely make it into the 16.3.0 release. Per the process, this enters a final comment period.

@trueadm is the core champion for this feature so he will fill in the missing fields (alternatives, drawbacks) based on our internal discussions of this for the past few months.

aweary:
@trueadm if it's an experimental implementation, should this land as `React._unstableCreateRef` in 16.3?

trueadm:
@aweary I don't believe there's any intention to release this as experimental given how little surface area it touches in the codebase and how it's primary intention is to replace a deprecated feature (string refs). What would making it `unstable_` help with vs without for 16.3?

aweary:
Nothing specific, just that APIs prefixed with `_unstable` are subject to breaking changes in minors, so it gives you a little more wiggle room if the API ends up not being sufficient. If this _isn't_ experimental (I was going off what @gaearon just said) then there's probably no need 🙂

gaearon:
It's not really experimental (we've been discussing this design for a few months by now). I only mean "experimental" in the sense that the RFC is still going through the final stage so there's a few days for extra feedback. 16.3 won't be released by then so it doesn't matter what's in master.

strayiker:
Alternate proposal:
```javascript
class Field extends Component {
  refs = React.createRefs();

  componentDidMount() {
    this.refs.input.focus();
  }

  render() {
    return <input ref={this.refs('input')} />;
  }
}
```
```javascript
function createRefs() {
  return refs.bind({});
}

function refs(name) {
  // ...some cache
  return node => {
    this[name] = node;
  };
}
```

One benefit of this is that no need to use React.createRef() for each ref. So this will even less cumbersome against callback refs.

trueadm:
@strayiker I feel that you'd lose the explicitness and typed power in exchange for passing around strings. For example, what if I named a ref "name" as `prototype` or `valueOf`, wouldn't it cause side-effects? You're safer using a `Map` but then you lose a lot of the magic you've described above (accessing `this.refs.input` but setting it via `this.refs('input')`.)

strayiker:
Yes you are right. I wanted to give the idea. With Map it can be done with something like this: setting `this.refs.create('input')` / `accessing this.refs.get('input')`.
In my opinion, the initial proposal have no benefits against callback refs at all.

```javascript
  inputRef = ref => { this.input = ref };
  inputRef = React.createRef();
```
-10 characters on defining
```javascript
  this.input.focus();
  this.inputRef.value.focus();
```
+9 characters on accessing
Or am I missing something?

text/0000-new-create-ref.md
```diff
@@ -0,0 +1,104 @@
...
+  );
+}
+```
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
```
streamich: Here is a user space implementation

```js
const createRef = () => {
  const ref = el => ref.value = el;
  return ref;
};
```

  - https://github.com/MailOnline/libreact/pull/19

TryingToImprove:
Is this something that is going to be allowed?

```js
class Parent extends Component {
    constructor() {
        this.childRef = React.createRef()
    }

    componentDidMount() {
        //////////---------------- NOTE: that the childRef come from another component
        /// this.childRef === document.getElementById('HELLO_WORLD') // true
    }

    render() {
        return <Child passedRef={this.childRef} />
    }
}

class Child extends Component {
    constructor() {
        this.childRef = React.createRef()
    }

    render() {
        return ( 
            <div>
                <span id="HELLO_WORLD" ref={this.props.passedRef}
            </div>
        )
    }
}
```

Seems like it is not possible with `string`-ref and `callback`-refs?

TrySound:
Nope, reference value will be set to null by that child

gaearon:
You can pass an object ref down the tree (just like a callback ref), but it doesn’t create memory leaks because React clears its value when the ref is detached.

text/0000-new-create-ref.md
gaearon: This is true. But in practice this pattern hasn’t caught on, and people keep using string refs in many places because they perceive functional refs as inconvenient. By adding this into the core we encourage people to adopt this pattern more strongly.

text/0000-new-create-ref.md
```diff
@@ -0,0 +1,104 @@
...
+```js
+class MyComponent extends React.Component {
+  divRef = React.createRef();
+
+  render() {
+    return <div><input type="text" ref={this.divRef} /></div>;
+  }
+
+  componentDidMount() {
+    this.divRef.value.focus();
```
j-f1: How about `this.divRef()` as a shortcut? People are going to be typing this a lot, and that extra `.value` is 4 characters longer than `()`. You could define `React.createRef` as something like this:

```js
function createRef() {
  const ref = () => ref.value;
  ref.value = null;
  ref.__THIS_IS_A_REACT_REF_DO_NOT_TOUCH_THIS_OR_YOU_WILL_BE_FIRED__ = true;
  return ref;
}
```

gaearon: If object ref is also a function? There’s no easy way to distinguish which is which. Sure, React could use a field for this, but it is pretty confusing for third party libraries.

gaearon: Also refs are an escape hatch. People shouldn’t be typing this *so* much.

text/0000-new-create-ref.md
strayiker: @gaearon Deprecation warning with link on [documentation](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) should have a similar effect.

gaearon: If we deprecate string refs before adding this helper people will get very annoyed. Because function refs are more cumbersome to deal with. Yes, we can suggest this helper. But at that point we’re suggesting everybody to copy the same thing into their projects. What’s the point? It might as well be provided by React then.

text/0000-new-create-ref.md
streamich: 👍  for `__THIS_IS_A_REACT_REF_DO_NOT_TOUCH_THIS_OR_YOU_WILL_BE_FIRED__` almost spilled my hot chocolate.

trueadm:
In regards to comparing the amount of characters to type – I think you're missing the point with this PR. `createRef` isn't meant to replace callback refs, if you like using callbacks refs, please continue to use them. The issue is that there are plenty of people out there who still use string refs because they dislike using callback refs and this RFC aims to address that target audience.

I've had the pleasure of attending React bootcamp classes and introduction sessions where newcomers to React start out. One of the common complaints is that callback refs are far too confusing when learning React compared to string refs. 

When I explain to them they can simply do `(divRef) => this._divRef = divRef` or something similar, they go on to ask questions like:

- "How do I add a debugger into the arrow function?"
- "I thought creating closures in the render method was expensive?"
- "Why is `divRef` called multiple times with different values?"

Obviously, you can tackle all those points and address them, but most people then just use string refs without the hassle.

TrySound:
@trueadm Are you ok to chat in twitter PM?

text/0000-new-create-ref.md
```diff
@@ -26,6 +26,25 @@ class MyComponent extends React.Component {
 
 # Motivation
 
+Strings refs bind to the React's component's `currentOwner` rather than the parent. That's something that isn't statical analysable and leads to most of bugs.
```
gaearon: Most people don’t know what “current owner” is. Can you please explain the concept here?

text/0000-new-create-ref.md
```diff
@@ -0,0 +1,171 @@
...
+          // ref foo2 would bind to ComponentB
+          // even though it's all in ComponentA's render
+          () => <div ref="foo2" />
+        }</ComponentB>
+      </div>
+    );
+  }
+}
+```
+
```
gaearon: Can you mention other problems with string refs? For example lack of composability, or that they break “multiple Reacts” case.

text/0000-new-create-ref.md
```diff
@@ -0,0 +1,171 @@
...
+This alternative API shouldn't provide any big real wins over callback refs - other than being a nice convenience feature. There might be some small wins in performance - as a common pattern is to assign a ref value in a closure created in the render phase of a component - this avoids that (even more so when a ref property is assigned to a non-existent component instance property).
```
gaearon: I would add here that the primary motivation is to encourage people to migrate off string refs. Callback refs meet some resistance because they are a bit harder to understand. We want to introduce this API primarily for people who love string refs today.

text/0000-new-create-ref.md
TrySound: What do you mean by "lack of composability" and how createRef fixes it?

TrySound: And what is the symptom of breaking with multiple reacts. Is there an error or just refs are not filled?

text/0000-new-create-ref.md
TrySound: I don't know how to describe it. Doesn't example below help?

text/0000-new-create-ref.md
```diff
@@ -49,6 +49,12 @@ class ComponentA {
 }
 ```
 
+It is not statical analysable and leads to most of bugs.
```
j-f1: > statically-analyzable

?

text/0000-new-create-ref.md
```diff
@@ -0,0 +1,181 @@
...
+
+The common practice will be using ref objects (aka createRef), unless people have already been using callback refs and are happy with them.
+
+Also there's no point in changing from callback refs to createRef right now unless people see value in doing so, otherwise it's just tech debt with no real additional value other than maybe some perf wins from not creating closures in the render method.
+
+# Adoption strategy
+
+Initially, we will release the object refs alongside the existing string refs as a minor update. Also this update will include `StrictMode` which enables deprecation messages for its subtree, so people be able to catch using string refs in their components and replace with the new api incrementally.
+
+String refs will be removed in upcoming major release, so there will be enought time for migration.
```
vkrol: enought -> enough?

meteorplus:
Oh my God I wasted so much time because of this because it is returning null half of the time in componentDidMount and is accessible only after a few seconds not right away!

What can't we just use the normal document.getElementById ? it was first and simple to do!

This is returning null most of the time! (sometimes I this.myRef.current to be defined)

Please test before approving something!

TrySound:
@meteorplus Everything is tested and works perfectly. You probably missed the point that refs inside conditions are available only when conditions are thruthy. To cover all cases with this api you should use both `componentDidMount` and `componentDidUpdate`.

```js
class A extends React.Component {
  div = React.createRef();
  state = {
    show: false
  }

  componentDidMount() {
    this.div.current // null because on component mount div is not rendered
  }

  componentDidUpdate() {
   // after click on button div element will be shown
    this.div.current // inside this hook it will be filled
  }

  render() {
    return (
      <>
        <button onClick={() => this.setState({ show: true })}>Click open</button>
        {this.state.show &&
          <div ref={this.div}></div>
        }
      </>
    )
  }
}
```

`getElementById` is a bad option because it requires to introduce unique identifiers which breaks component isolation.

gaearon:
@meteorplus If you experience a bug in React please file a new issue with reproducing example. An RFC is not an appropriate place to express your frustration. Thanks!

PR #20, opened 2018-02-05
getInitialState support for ES6 classes

dryganets:


TrySound:
@dryganets Classes are es6

TrySound:
Also there's new [static getDerivedStateFromProps](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md) which may handle similar case and even more handle all props updates.

dryganets:
Yep, I mentioned getDerivedStateFromProps.
It is static, so it is impossible to override the behavior in in subclasses.

j-f1:
@dryganets 

```js
  static getDerivedStateFromProps(props) {
    return { /* ... */ }
  }
}

class Bar extends Foo {
  static getDerivedStateFromProps(props) {
    const state = Foo.getDerivedStateFromProps(props)
    /* ... */
    return state
  }
}

**Comment by dryganets** [comment]
@j-f1, this might be good enough solution.

The drawbacks are: 
- In this case, only static context of the class is available.
- No inheritance. Need to mention exact class instead of super call everywhere

**Comment by j-f1** [comment]
> No inheritance. Need to mention exact class instead of super call everywhere

Turns out this [isn’t necessary](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=PQKgBAZg9gTmAmBTAtlMJgCgDGUB2AzgC5gBKiAhtiQLxgDeYAwlMgA76J5EBcY2AGwoECDAL5gxmHEJFgAYlDSIAHkS7xR5KkQB0Ldp24NMYMMQpEAltjABzREQAiiGFYBuieAGUilxPIwrAAKQWwEABRsYQQAlCZmZjCOAK4weAyQSnxEMCmIkqaFUjLCogBCFHCq6niaCkoJ5n7Wtg7Orh5evv6BITFRMfH0RWa4hCQW6mB0BClsrrrtLm6ePi0BQcihUOGDu3Gjzf66AEZVM2C5-UfJRGkZU4hFUiUUAq5EEQBS3gDyADldMQ3Hg7FYIABPCKVGBLRwrLrrXpbHZ7ehiWJYoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=stage-2&prettier=false&targets=&version=6.26.0&envVersion=). You can use `super` just fine in class methods, and it executes the class method of the parent class.

**Comment by kingdaro** [comment]
Convenient, but doesn't seem all that necessary when it's only one extra line to accomplish this on your own:
```js
state = this.getInitialState()

getInitialState() {
 // ...
}
```

dryganets:
And to fix the inheritance problem just add:

`this` to the state in constructor.
Later on just ignore this field of state in `shouldComponentUpdate` and you could access `this` like this:
```state.this.blablabla```

PR #30, opened 2018-03-07
ref-forwarding API

bvaughn:
[View formatted RFC](https://github.com/bvaughn/rfcs/blob/ref-forwarding/text/0000-ref-forwarding.md)

Implementation at facebook/react/pull/12346

### DIFF
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -90,7 +90,7 @@ function injectLegacyRelayContext(Component) {
   // Create a special React component type that exposes the external 'ref'.
   // This lets us pass it along as a regular prop,
   // And attach it as a regular React 'ref' on the component we choose.
-  return React.useRef((props, ref) => (
+  return React.forwardRef((props, ref) => (
     <LegacyRelayContextConsumer {...props} forwardedRef={ref} />
   ));
 }
@@ -113,7 +113,7 @@ function withTheme(ThemedComponent) {
   }
 
   // Forward refs through to the inner, "themed" component:
-  return React.useRef((props, ref) => (
+  return React.forwardRef((props, ref) => (
     <ThemeContextInjector {...props} forwardedRef={ref} />
   ));
 }
@@ -123,7 +123,7 @@ function withTheme(ThemedComponent) {
 
 Hopefully the usage of this API is clear from the [above examples](#basic-example), so in this section I'll outline a possible implementation strategy.
 
-The `useRef` function could return a wrapper object (similar to what the context API uses) with a `$$typeof` indicating that it's a ref-forwarding component. When React encounters this type, it can assign a new type-of-work to React (e.g. `UseRef`). The "begin" phase could then invoke the render prop argument, passing it `workInProgress.pendingProps` and `workInProgress.ref` to create the children, and then continue reconciliation.
+The `forwardRef` function could return a wrapper object (similar to what the context API uses) with a `$$typeof` indicating that it's a ref-forwarding component. When React encounters this type, it can assign a new type-of-work to React (e.g. `UseRef`). The "begin" phase could then invoke the render prop argument, passing it `workInProgress.pendingProps` and `workInProgress.ref` to create the children, and then continue reconciliation.
 
 # Drawbacks
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -114,11 +114,15 @@ Specify a ["high-level" flag on the component](https://github.com/facebook/react
 
 # Adoption strategy
 
-Coming soon...
+This is a new feature. Since there are no backwards-compatibility concerns, adoption can be organic.
+
+One candidate that would immediately benefit from this feature would be [`create-subscription`](https://github.com/facebook/react/pull/12325).
 
 # How we teach this
 
-Coming soon...
+Add a section to the ["Refs and the DOM" documentation page](https://reactjs.org/docs/refs-and-the-dom.html) about ref-forwarding.
+
+Write a blog post about the new feature and when/why you might want to use it. Highlight examples using the new context API.
 
 # Unresolved questions
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -90,7 +90,7 @@ function injectLegacyRelayContext(Component) {
   // Create a special React component type that exposes the external 'ref'.
   // This lets us pass it along as a regular prop,
   // And attach it as a regular React 'ref' on the component we choose.
-  return React.withRef((props, ref) => (
+  return React.useRef((props, ref) => (
     <LegacyRelayContextConsumer {...props} forwardedRef={ref} />
   ));
 }
@@ -98,7 +98,7 @@ function injectLegacyRelayContext(Component) {
 
 # Detailed design
 
-Coming soon...
+TODO
 
 # Drawbacks
 
@@ -106,7 +106,7 @@ This API increases the surface area of React slightly, and may complicate compat
 
 # Alternatives
 
-1. Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that coudl be used to get the inner ref. (Some drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
+1. Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that could be used to get the inner ref. (Some drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
 
 2. Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it. This appraoch could enable refs to be forwarded one level (to the immediate child) but would not enable forwarding to deeper child, e.g.:
 
@@ -127,7 +127,7 @@ function withTheme(ThemedComponent) {
 }
 ```
 
-3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) This appraoch would not enable class components to forward refs, and so would not be sufficient, since wrapper components often require class lifecycles.
+3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) This appraoch would not enable class components to forward refs, and so would not be sufficient, since wrapper components often require class lifecycles. It would also have the same child-depth limitations as the above option.
 
 # Adoption strategy
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -96,6 +96,29 @@ function injectLegacyRelayContext(Component) {
 }
 ```
 
+The theme context wrapper could as well:
+
+```js
+const ThemeContext = React.createContext("light");
+
+function withTheme(ThemedComponent) {
+  function ThemeContextInjector(props) {
+    return (
+      <ThemeContext.Consumer>
+        {value => (
+          <ThemedComponent {...props} ref={props.forwardedRef} theme={value} />
+        )}
+      </ThemeContext.Consumer>
+    );
+  }
+
+  // Forward refs through to the inner, "themed" component:
+  return React.useRef((props, ref) => (
+    <ThemeContextInjector {...props} forwardedRef={ref} />
+  ));
+}
+```
+
 # Detailed design
 
 TODO
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -170,7 +170,7 @@ function withTheme(ThemedComponent) {
 }
 ```
 
-3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) This appraoch would not enable class components to forward refs, and so would not be sufficient, since wrapper components often require class lifecycles. It would also have the same child-depth limitations as the above option.
+3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) This approach would not enable class components to forward refs, and so would not be sufficient, since wrapper components often require class lifecycles. It would also have the same child-depth limitations as the above option.
 
 # Adoption strategy
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -121,7 +121,9 @@ function withTheme(ThemedComponent) {
 
 # Detailed design
 
-TODO
+Hopefully the usage of this API is clear from the [above examples](#basic-example), so in this section I'll outline a possible implementation strategy.
+
+The `useRef` function could return a wrapper object (similar to what the context API uses) with a `$$typeof` indicating that it's a ref-forwarding component. When React encounters this type, it can assign a new type-of-work to React (e.g. `UseRef`). The "begin" phase could then invoke the render prop argument, passing it `workInProgress.pendingProps` and `workInProgress.ref` to create the children, and then continue reconciliation.
 
 # Drawbacks
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -10,7 +10,7 @@ Provide an API that enables [`refs`](https://reactjs.org/docs/refs-and-the-dom.h
 
 I recently began contributing to [`react-relay` modern](https://github.com/facebook/relay/tree/master/packages/react-relay/modern) in order to prepare it for the [upcoming React async-rendering feature](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html).
 
-One of the first challenges I encountered was that `react-relay` depends on the legacy, unstable context API, and values passed through legacy `context` aren't accessible in the new, [static `getDerivedStateFromProps` lifecycle](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md#static-getderivedstatefrompropsnextprops-props-prevstate-state-shapestate--null). The long-term plan for `react-relay` is to use the [new and improved contex API](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md) but this can't be done without dropping support for older versions of React- (something `react-relay` may not be able to do yet).
+One of the first challenges I encountered was that `react-relay` depends on the legacy, unstable context API, and values passed through legacy `context` aren't accessible in the new, [static `getDerivedStateFromProps` lifecycle](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md#static-getderivedstatefrompropsnextprops-props-prevstate-state-shapestate--null). The long-term plan for `react-relay` is to use the [new and improved context API](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md) but this can't be done without dropping support for older versions of React- (something `react-relay` may not be able to do yet).
 
 In an effort to work around this, I created a smaller wrapper component to convert the necessary `context` value to a `prop` so that it could be accessed within the static lifecycle:
 
@@ -151,7 +151,7 @@ This API increases the surface area of React slightly, and may complicate compat
 
 1. Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that could be used to get the inner ref. (Some drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
 
-2. Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it. This appraoch could enable refs to be forwarded one level (to the immediate child) but would not enable forwarding to deeper child, e.g.:
+2. Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it. This approach could enable refs to be forwarded one level (to the immediate child) but would not enable forwarding to deeper child, e.g.:
 
 ```js
 const ThemeContext = React.createContext("light");
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -1,0 +1,121 @@
+- Start Date: 2018-03-07
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Provide an API that enables [`refs`](https://reactjs.org/docs/refs-and-the-dom.html) to be forwarded to a child component.
+
+# Motivation
+
+I recently began contributing to [`react-relay` modern](https://github.com/facebook/relay/tree/master/packages/react-relay/modern) in order to prepare it for the [upcoming React async-rendering feature](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html).
+
+One of the first challenges I encountered was that `react-relay` depends on the legacy, unstable context API, and values passed through legacy `context` aren't accessible in the new, [static `getDerivedStateFromProps` lifecycle](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md#static-getderivedstatefrompropsnextprops-props-prevstate-state-shapestate--null). The long-term plan for `react-relay` is to use the [new and improved contex API](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md) but this can't be done without dropping support for older versions of React- (something `react-relay` may not be able to do yet).
+
+In an effort to work around this, I created a smaller wrapper component to convert the necessary `context` value to a `prop` so that it could be accessed within the static lifecycle:
+
+```js
+function injectLegacyRelayContext(Component) {
+  function LegacyRelayContextConsumer(props, legacyContext) {
+    return <Component {...props} relay={legacyContext.relay} />;
+  }
+
+  LegacyRelayContextConsumer.contextTypes = {
+    relay: RelayPropTypes.Relay
+  };
+
+  return LegacyRelayContextConsumer;
+}
+```
+
+Unfortunately, this change broke several projects within Facebook that depended on refs to access the inner Relay components.
+
+As I considered this, I became convinced that the new Context API naturally lends itself to similar wrapper components, e.g.:
+```js
+const ThemeContext = React.createContext("light");
+
+function withTheme(ThemedComponent) {
+  return function ThemeContextInjector(props) {
+    return (
+      <ThemeContext.Consumer>
+        {value => <ThemedComponent {...props} theme={value} />}
+      </ThemeContext.Consumer>
+    );
+  };
+}
+```
+
+Unfortunately, within the current limitations of React, there is no (transparent) way for a 
+a `ref` to be attached to the above `ThemedComponent`. Common workarounds typically use a special prop (e.g. `componentRef`) like so:
+```js
+const ThemeContext = React.createContext("light");
+
+function withTheme(ThemedComponent) {
+  return function ThemeContextInjector(props) {
+    return (
+      <ThemeContext.Consumer>
+        {value => (
+          <ThemedComponent {...props} ref={props.componentRef} theme={value} />
+        )}
+      </ThemeContext.Consumer>
+    );
+  };
+}
+```
+
+This convention varies from project to project, and requires users of the `ThemedComponent` to be aware of the fact that it's wrapped by a higher-order component. This detail should not be important. It should be possible to use a standard React `ref` that the `ThemeContextInjector` (in this case) could forward to its child `ThemedComponent`.
+
+The idea of `ref`-forwarding [is not new](https://github.com/facebook/react/issues/4213), but the new context and other efforts like [`create-subscription`](https://github.com/facebook/react/pull/12325) greatly increase the importance of this feature.
+
+# Basic example
+
+The Relay `context` injector component ([shown above](#motivation)) could use the ref-forwarding API proposed by this RFC as follows:
+
+```js
+function injectLegacyRelayContext(Component) {
+  function LegacyRelayContextConsumer(props, legacyContext) {
+    return (
+      <Component
+        {...props}
+        relay={legacyContext.relay}
+        ref={props.forwardedRef}
+      />
+    );
+  }
+
+  LegacyRelayContextConsumer.contextTypes = {
+    relay: RelayPropTypes.Relay
+  };
+
+  // Create a special React component type that exposes the external 'ref'.
+  // This lets us pass it along as a regular prop,
+  // And attach it as a regular React 'ref' on the component we choose.
+  return Ref.forward((props, ref) => (
+    <LegacyRelayContextConsumer {...props} forwardedRef={ref} />
+  ));
+}
+```
+
+# Detailed design
+
+Coming soon...
+
+# Drawbacks
+
+Coming soon...
+
+# Alternatives
+
+Coming soon...
+
+# Adoption strategy
+
+Coming soon...
+
+# How we teach this
+
+Coming soon...
+
+# Unresolved questions
+
+None presently.
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Provide an API that enables [`refs`](https://reactjs.org/docs/refs-and-the-dom.html) to be forwarded to a child component.
+Provide an API that enables [`refs`](https://reactjs.org/docs/refs-and-the-dom.html) to be forwarded to a descendant (child or grandchild component).
 
 # Motivation
 
@@ -106,11 +106,28 @@ This API increases the surface area of React slightly, and may complicate compat
 
 # Alternatives
 
-Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that coudl be used to get the inner ref. (Drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
+1. Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that coudl be used to get the inner ref. (Drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
 
-Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it.
+2. Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it. This appraoch could enable refs to be forwarded one level (to the immediate child) but would not enable forwarding to deeper child, e.g.:
 
-[Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) I don't believe this approach would be sufficient, since wrapper components often require class lifecycles.
+```js
+const ThemeContext = React.createContext("light");
+
+function withTheme(ThemedComponent) {
+  return function ThemeContextInjector(props) {
+    return (
+      <ThemeContext.Consumer>
+        {value => (
+          // ref belongs here
+          <ThemedComponent {...props} ref={props.componentRef} theme={value} />
+        )}
+      </ThemeContext.Consumer>
+    );
+  };
+}
+```
+
+3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) I don't believe this approach would be sufficient, since wrapper components often require class lifecycles.
 
 # Adoption strategy
 
--- a/text/0000-ref-forwarding.md
+++ b/text/0000-ref-forwarding.md
@@ -90,7 +90,7 @@ function injectLegacyRelayContext(Component) {
   // Create a special React component type that exposes the external 'ref'.
   // This lets us pass it along as a regular prop,
   // And attach it as a regular React 'ref' on the component we choose.
-  return Ref.forward((props, ref) => (
+  return React.withRef((props, ref) => (
     <LegacyRelayContextConsumer {...props} forwardedRef={ref} />
   ));
 }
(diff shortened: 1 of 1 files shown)


trueadm:
I really like this approach. Makes a lot of sense for existing applications and should be something fairly simple to build into the compiler to handle too.

acdlite:
Maybe this is out of scope, but could be worth mentioning that this same strategy can be used for all HOCs. For example:

```js
function ThemedButton(props, theme) {
  return <button className={'theme-' + theme} {...props} />
}

export default withTheme(Movie);
```

where `withTheme` internally uses a context consumer.

acdlite:
@streamich This is an HOC, except it only works for functional components. The difference is that instead of injecting additional props, it passes additional arguments.

The larger idea behind this proposal is that we could use the same pattern for other APIs, too. Like, say, context:

```js
const ThemeContext = React.createContext('light');

function ThemedButton(props, theme) {
  return <button className={'theme-' + theme} {...props} />;
}

export default ThemeContext.withContext(ThemedButton);
```

acdlite:
@sebmarkbage suggested we use the `use` prefix instead of `with` since some HOCs may add more than one argument:

```js
function Counter(props, count, updateCount) {
  return (
    <>
      Count: {count}
      <button onClick={() => updateCount(c => c + 1)}>+</button>
    </>
  );
}

export default useState(Counter, 0);
```

Or no arguments at all:

```js
function ComputationallyExpensiveComponent(props) {
  const thing = somethingThatTakesALongTime(props);
  return <Whatever thing={thing} />;
}

const MemoizedComponent = usePure(ComputationallyExpensiveComponent);
```

bvaughn:
Thanks @trueadm!

> Maybe this is out of scope, but could be worth mentioning that this same strategy can be used for all HOCs.

~~Yeah, this was what I was thinking, but maybe didn't make it clear enough. I'll try to make it clearer in the code example.~~

Sorry, I misread this message initially. I still agree with the sentiment of your comment, but I think mentioning it might be a bit out of scope for this RFC. 😄

> @sebmarkbage suggested we use the use prefix instead of with since some HOCs may add more than one argument

@sebmarkbage always has good ideas. 😄I'll update!

text/0000-ref-forwarding.md
```diff
@@ -0,0 +1,169 @@
...
+      <ThemeContext.Consumer>
+        {value => <ThemedComponent {...props} theme={value} />}
+      </ThemeContext.Consumer>
+    );
+  };
+}
+```
+
+Unfortunately, within the current limitations of React, there is no (transparent) way for a 
+a `ref` to be attached to the above `ThemedComponent`. Common workarounds typically use a special prop (e.g. `componentRef`) like so:
```
mxstbr: We use `innerRef` for styled-components so users can get access to the underlying DOM node, see https://www.styled-components.com/docs/advanced#refs

jacobwgillespie:
How does this handle static class properties?  It would seem to swallow them, if you did something like:

```javascript
class DataContainer extends Component {
  static fragments = {widget: gql`...`}
}

export default React.useRef((props, ref) => (
  <DataContainer {...props} forwardedRef={ref} />
));
```

Perhaps the solution / challenge is the same as other higher-order components, in that it would either require buy-in on the original component itself so that it attached static properties to the wrapper or would require a workaround like `hoist-non-react-statics`.  

That said, it would be nice if however this worked did not require knowledge of static properties, especially in the case where you want to wrap a third-party component with `useRef` without needing to know the full implementation details of the wrapped component.

This very well may be a non-issue.

bvaughn:
> This would work 100% for styled-components.

That's great to hear, Max!

> How does this handle static class properties?

> Perhaps the solution / challenge is the same as other higher-order components, in that it would either require buy-in on the original component itself **so that it attached static properties to the wrapper**

I don't think static properties are a problem here (but please correct me if I'm overlooking something). I think your observation about attaching/forwarding them to the exported ref-type is a reasonable solution.

The key part is that you would be able to do this in a way that's transparent to users of the thing you'r wrapping. (That isn't currently possible, due to refs.)

dantman:
> Perhaps the solution / challenge is the same as other higher-order components, in that it would either require buy-in on the original component itself so that it attached static properties to the wrapper or would require a workaround like hoist-non-react-statics.

I'd be wary of this at the library level. When I write a HOC I actually don't use `hoist-non-react-statics`, I use `hoist-non-react-methods` because often when wrapping a component in a HOC not only do statics get shadowed but also instance methods like `focus`/`show`/`hide` and React Native's `measure`.

So if we create something at the library level that shadows static properties and the library has to workaround it, I'm worried that if no-one remembers to take them into account we'll end up with something baked into the library (where we can't just switch to a different implementation) that handles statics but forgets about instance methods on the prototype.

elado:
Thanks for writing this!

I'm having a similar issue where I have some HOCs over an <input> and need to expose it's `focus()` method. it's only possible with some custom `innerRef` prop and a recursive `getInnerMostRef` helper. But this will work only if my HOCs are in my control and know how to keep track of inner refs.

Not really related, but if there's any API change suggested, could React specific props be more distinct?

e.g.

```jsx
<Comp $ref={...} $key={...} />
<Comp react-ref={...} react-key={...} />
<Comp react.ref={...} react.key={...} />
```

this will allow more explicit way to define react props and prevent breaking changes when new react-specific props are added. Also will allow proper spread of objects that may contain `ref` or `key` attributes.

probablyup:
@elado XHTML does have a concept of namespaces, could do something like `react:key`, etc.

mxstbr:
Namespaces in JSX have been discussed before: https://github.com/facebook/jsx/issues/81 We would love to have them for styled-components but it doesn't seem like that's happening.

bvaughn:
> I'd be wary of this at the library level. When I write a HOC I actually don't use hoist-non-react-statics, I use hoist-non-react-methods because often when wrapping a component in a HOC not only do statics get shadowed but also instance methods like focus/show/hide and React Native's measure.

I might be misunderstanding your concern, but this ref-forwarding proposal is specifically intended to _help_ with the shadowing of instance methods like `focus`/`show`/`hide`. If you decorate/wrap a component with those methods, so long as you forward the `ref` to it, then the owner of the `ref` can use those methods (e.g. `ref.value.focus()`).

> So if we create something at the library level that shadows static properties and the library has to workaround it

I don't think static properties is a common concern. Maybe I'm wrong. But if this is a case for a given component, there is a workaround (discussed above).

WRT namespaced attributes, that's outside of the scope of this RFC. :smile:

dantman:
> I might be misunderstanding your concern, but this ref-forwarding proposal is specifically intended to help with the shadowing of instance methods like focus/show/hide. If you decorate/wrap a component with those methods, so long as you forward the ref to it, then the owner of the ref can use those methods (e.g. ref.value.focus()).

True, I could just forward the ref to the top-level component. But I am worried that: A) sometimes I want to forward refs for something deep but I still want to offer direct methods on my component for the top level component (Say a `<PromptDialog>` that forwards a ref to the text input for `inputRef.focus()` but directly exposes it's own `show`/`hide` instead of forwarding a `dialogRef` to the `<Dialog>` it happens to use internally). B) Sometimes I may have a good reason to expose an instance method as part of my Component's API, after all here and there even React Native has reasons to do so. And I don't believe forwarding refs help that, but it sounded like they interfere with hoisting if it turns out I need to forward refs for one of my component's children.

bvaughn:
> sometimes I want to forward refs for something deep but I still want to offer direct methods on my component for the top level component

This comment doesn't make sense to me.

I think the `ref` should always point to the conceptual "top-level component". So in the case of a `Button` component that's wrapped by a `withTheme` HOC, the `ref` attribute would point to the `Button` component.

In the case you describe, the top-level component sounds like it would be the `PromptDialog`, which may look something like this:

```js
class PromptDialog extends React.Component {
  inputRef = React.createRef();

  show() {
    // Show...
  }

  hide() {
    // Hide...
  }

  focus() {
    if (this.inputRef.value != null) {
      this.inputRef.value.focus();
    }
  }

  render() {
    return isShown ? <input ref={this.inputRef} /> : null;
  }
}
```

External users of your dialog would use it like so:

```js
class Example extends React.Component {
  dialogRef = React.createRef();

  componentDidMount() {
    this.dialogRef.value.show();
    this.dialogRef.value.focus();
  }

  render() {
    return <PromptDialog ref={dialogRef} />;
  }
}
```

It doesn't make sense for a single `ref` to point to multiple things. If you want to be able to have multiple refs, you still could:

```js
class PromptDialog extends React.Component {
  show() {
    // Show...
  }

  hide() {
    // Hide...
  }

  render() {
    // Forward props.inputRef so it points to the inner <input>
    return isShown ? <input ref={this.props.inputRef} /> : null;
  }
}

function wrapForSomeReason(Component) {
  function WrapperComponent(props) {
    // (This wrapper intentionally left blank to simplify the example.)
    return <Component {...props} ref={props.forwardedRef} />;
  }

  // Forward top-level `ref` so it points to Component
  return React.useRef((props, ref) => (
    <WrapperComponent {...props} forwardedRef={ref} />
  ));
}

// Top-level `ref` will point to PromptDialog
const WrappedDialog = wrapForSomeReason(PromptDialog);
```

> Sometimes I may have a good reason to expose an instance method as part of my Component's API, after all here and there even React Native has reasons to do so. And I don't believe forwarding refs help that, but it sounded like they interfere with hoisting if it turns out I need to forward refs for one of my component's children.

I don't really understand this comment, which makes me wonder if you're thinking of this proposal as something fundamentally different than I am. (Or maybe I'm just...misunderstanding your comment.) 😄

Do my above examples clarify any? If not, could you write a small example component that illustrates your point? I don't see how this proposal complicates or prevents you from doing what you're describing.

dantman:
@bvaughn Sorry, I read the RFC but I'll admit I found it hard to understand. Your last example makes sense and fits. I was just going off another comment talking about the interactions with hoisting and I worried because HOC usage is fairly common for me. For `wrapForSomeReason` I often have a bunch of reason to use some HOC, a theme context providing HOC, an API providing HOC, etc. And some of them are used on something like the `PromptDialog` that may be a component provided for other components to use and have a reason to have instance methods.

bvaughn:
No problem. If you have suggestions for ways I can clarify the RFC wording, let me know.

This RFC is intended to help with the specific case you're describing - so that you can wrap your `PromptDialog` with as many HOCs as you need to, without users of `PromptDialog` needing to know or care (because `ref` would just get passed through to your `PromptDialog`).

jacobwgillespie:
> I don't think static properties are a problem here (but please correct me if I'm overlooking something).

This is the case I'm thinking of:

```javascript
import SomeComponent from 'third-party-library';

// SomeComponent has static properties, but I don't necessarily know
// this because it's a third-party NPM module that's not in my control.

const RefWrappedSomeComponent = React.useRef((props, ref) => (
  <SomeComponent {...props} forwardedRef={ref} />
));

export default RefWrappedSomeComponent;
```

In that case, I would need to know about the implementation of `SomeComponent` in order to know whether or not static properties would need to be hoisted, but it would be nice if there was a way to wrap the ref on a component without necessitating static hoisting.

But like I said before, there may not be, given that this is a common concern with HOCs.

The "fix" using this API would be something like this:

```javascript
import hoistNonReactStatics from 'hoist-non-react-statics';
import SomeComponent from 'third-party-library';

const RefWrappedSomeComponent = React.useRef((props, ref) => (
  <SomeComponent {...props} forwardedRef={ref} />
));

hoistNonReactStatics(RefWrappedSomeComponent, SomeComponent);

export default RefWrappedSomeComponent;
```

bvaughn:
> In that case, I would need to know about the implementation of `SomeComponent` in order to know whether or not static properties would need to be hoisted, but it would be nice if there was a way to wrap the ref on a component without necessitating static hoisting.

That's fair. My thinking about this is that most times, if you're wrapping a component in this way, you fall into one of two categories:
1. You're working in a library that defines the component (e.g. react-select does this) in which case you know if the component has static methods, or...
2. You're working in an application that _uses_ the component, in which case you can search to see if you're using any static methods within your codebase.

I don't imagine many ambiguous cases where you import a third party component, wrap it, and re-export it for external use. Although I haven't personally used `hoist-non-react-statics` before, it seems like it could be a reasonable workaround for that case though.

I don't have any data to back this up, but my anecdotal experience is that instance methods on components far outweigh static methods.

bvaughn:
Renamed `React.useRef` to `React.forwardRef`

bvaughn:
This RFC is now in the final comment period. A version of this will likely be released in an upcoming 16.3 alpha as well.

btnwtn:
Hey @bvaughn how could I use this with a Stateful Component?

```js
render() {
  return React.forwardRef((...?) => (
    null
  ))
}
```

gaearon:
```js
  render() {
    // Use this.props.myRef (or whatever you called it)
  }
}

// Export this instead
export default React.forwardRef((props, ref) => (
  <Button {...props} myRef={ref} />
));
```

bvaughn:
I'm not sure I understand that example, @btnwtn.

It doesn't really make sense to create a ref-forwarding component wrapper if you have access to the thing inside of it (which you would have to if you were rendering it).

It's meant to be used when you want to export one type of component for rendering, but let refs pass through to another type- inside of what's rendered.
```js
const ThemeContext = React.createContext("light");

function withTheme(ThemedComponent) {
  function ThemeContextInjector(props) {
    return (
      <ThemeContext.Consumer>
        {value => (
          <ThemedComponent {...props} ref={props.forwardedRef} theme={value} />
        )}
      </ThemeContext.Consumer>
    );
  }

  // Forward refs through to the inner, "themed" component:
  return React.forwardRef((props, ref) => (
    <ThemeContextInjector {...props} forwardedRef={ref} />
  ));
}

// We have a handle on ButtonComponent here,
// We want to export a themed wrapper (so the button is auto-themed).
// But we want users to be able to get a ref to button.
export const ThemedButtonComponent = withTheme(ButtonComponent);

// This ref will point to the inner ButtonComponent,
// Even though we are rendering the "themed" wrapper.
<ThemedButtonComponent ref={buttonRef} />
```

Or to borrow Dan's example from Twitter:
```js
// Tell React I want to be able to put a ref on LikeButton.
const LikeButton = React.forwardRef((props, ref) => (
  <div className="LikeButton">
    {/* Forward LikeButton's ref to the button inside. */}
    <button ref={ref} onClick={props.onClick}>
      Like
    </button>
  </div>
));

// I can put a ref on it as if it was a DOM node or a class!
// (In this example, the ref points directly to the DOM node.)
<LikeButton ref={myRef} />;
```

gaearon:
@bvaughn Minor feedback on the presentation of these examples—I think it requires some additional mental effort to understand how this plays together with context. While this is useful as one practical case we should mention, it's probably easier to explain this feature alone.

satya164:
I wonder why not do something like the following and support both class and function components:

```js
const Header = props => React.forwardRef(ref => (
  <div ref="ref">{props.message}</div>
));

class Header extends React.Component {
  render() {
    return React.forwardRef(ref => (
      <div ref="ref">{this.props.message}</div>
    ));
  }
}
```

I'm not clear on the implementation, so I might be mis-interpreting how this works.

bvaughn:
This does support both class and functional components 😄 

```js
// Functional example...
const Header = ({ children, forwardedRef }) => (
  <header ref={forwardedRef}>{children}</header>
);

// Class example...
class Header extends React.Component {
  render() {
    const { children, forwardedRef } = this.props;
    return <header ref={forwardedRef}>{children}</header>;
  }
}

// You would use either the same way...
const ForwardingHeader = React.forwardRef((props, ref) => (
  <Header {...props} forwardedRef={ref} />
));
```

satya164:
> This does support both class and functional components 

Yeah, I saw that. Was wondering an alternative approach so you can do this in render directly rather than one extra wrapping.

gaearon:
>support both class and function components

The RFC as it is already supports both functional and class components. The thing you pass to `forwardRef` is not exactly a component. It's more like a render prop. Inside that render prop, you can render either a function or a class component (or any kind of tree really).

The API you propose seems aesthetically nicer to me but I don’t think it works conceptually. By the time you’re in the render method it’s already “too late”. For example consider what happens if you make the return value of `render` conditional, with one branch returning a plain div. It would be surprising if the component’s ref pointer “jumped” from a forwarded ref to a class instance (when we don’t render a `forwardRef` return value) depending on that branch at runtime. Especially considering this could happen deep in the tree.

satya164:
> It would be surprising if the component’s ref pointer “jumped” from a forwarded ref to a class instance

I wouldn't really be surprised if it jumps to the class instance when don't return the stuff wrapped in `React.forwardRef`, assuming you're talking about this:

```js
class Header extends React.Component {
  render() {
    if (this.props.isLoading) {
      return <div>Loading</div>;
    }

    return React.forwardRef(ref => (
      <div ref={ref}>{props.message}</div>
    ));
  }
}
```

This seems like the intended behaviour, since for the element I returned, I didn't signal it to forward the ref by wrapping in `React.forwardRef`.

However, in the following case, I'd expect the `ref` to become `null`:

```js
class Header extends React.Component {
  render() {
    return React.forwardRef(ref => {
      if (this.props.isLoading) {
        return <div>Loading</div>;
      }

      return <div ref={ref}>{props.message}</div>;
    });
  }
}
```

Same in this case:

```js
const Header = React.forwardRef(props, ref => {
  if (props.isLoading) {
    return <div>Loading</div>;
  }

  return <div ref={ref}>{props.message}</div>;
});
```

gaearon:
>I wouldn't really be surprised if it jumps to the class instance when don't return the stuff wrapped in React.forwardRef, assuming you're talking about this

It looks nice and tidy if it’s top level return value. What if it’s nested? Maybe even in a chain of helper methods called from render. Then it influencing “own” ref of the component seems quite confusing to me. You also can’t safely “extract” a component that contains this. It would change the meaning. So this violates existing expectations.

If something is only allowed at top level then we should enforce it. That’s what the API does.

gaearon:
In general our constraint (which is partly responsible for React’s success) is that changing what a component returns from render doesn’t change its own API. That’s what allows easy refactorings. There are a few APIs that violate this but they have other issues on their own (findDOMNode). I think we shouldn’t add more APIs that pierce through component abstraction.

jquense:
super excited about this. I'm going to give it a whirl in some of the react-bootstrap v4 upgrades and see how it feels.

rickbeerendonk:
Aren't we widening the gap between functional and class components?

In functional components we can just use `ref` in our render method, but in class components we don't have `this.ref` (or `this.props.ref`).

bvaughn:
> In functional components we can just use `ref` in our render method, but in class components we don't have `this.ref` (or `this.props.ref`).

You don't have `ref` in your render method for a functional component either. (Refs can't be attached to functional components.)

AlmeroSteyn:
Hey everyone, this is really awesome, thank you!

As this features very strongly in focus control to achieve accessibility, I'd like to update the a11y doc to reflect this at the right moment.

Is it prudent to already open a separate PR for that at this stage?

bvaughn:
Hi @AlmeroSteyn. Are you talking about [this section](https://reactjs.org/docs/accessibility.html#programmatically-managing-focus) of the React docs? If so, feel free to submit a PR at any point if you have ideas to improve the docs. 😄

rickbeerendonk:
@bvaughn 

So this is not a functional component:
```const LikeButton = React.forwardRef((props, ref) => (
  <div className="LikeButton">
    {/* Forward LikeButton's ref to the button inside. */}
    <button ref={ref} onClick={props.onClick}>
      Like
    </button>
  </div>
));```

But what's a proper name?

**Comment by bvaughn** [comment]
Ah. I see.

The like button example isn't a functional component, although it looks similar to one.

`forwardRef` accepts a render function that returns a react node. It's [invoked directly](https://github.com/facebook/react/blob/d38616d6936542cbf8ca34d3c75cd3b9e44945c9/packages/react-reconciler/src/ReactFiberBeginWork.js#L159-L162), rather than being wrapped up in a React element. This is why its second parameter is `ref` (instead of the legacy `context`).

**Comment by gaearon** [comment]
@rickbeerendonk Perhaps the example in https://github.com/reactjs/rfcs/pull/30#issuecomment-373170677 will mark more sense to you? It shows it works the same way for functional components and classes.

**Comment by rickbeerendonk** [comment]
@gaearon I'd read that comment before reacting. A first reaction might be "why do we need this if we can just add a forwardRef prop to our component?". So this feature cannot be explained all by itself, but needs context and requires knowledge of other features.

**Comment by bvaughn** [comment]
> So this feature cannot be explained all by itself, but needs context and requires knowledge of other features.

Hopefully the ["Motivation" section of the RFC](https://github.com/bvaughn/rfcs/blob/ref-forwarding/text/0000-ref-forwarding.md#motivation) provides sufficient context for at least one use case 😄 the comments here are partial examples, assuming shared context of the RFC content.

**Comment by rickbeerendonk** [comment]
@bvaughn Not really. If react-relay has a problem, why should React "solve" this?

> ...requires users of the ThemedComponent to be aware of the fact that it's wrapped by a higher-order component. This detail should not be important.

But in https://github.com/reactjs/rfcs/pull/30#issuecomment-373170677 we end up with 2 "components" `Header` and `ForwardHeader` and a different meaning of `ref`. Maybe it improves backwards compatibility, but future users are required to know the difference between the two.

**Comment by gaearon** [comment]
> If react-relay has a problem, why should React "solve" this?

It's not isolated to React Relay, the exact same problem will affect any other library that will want to use the context API and has some imperative hooks. The only reason we mentioned React Relay is *we're the ones migrating it* so we're first to experience this problem.

This is also not strictly limited to the context API (although that's what tipped the scales). The problem has existed for years for any HOCs and, in general, React DOM component libraries. You can see it solves longstanding pain points for [React Bootstrap](https://mobile.twitter.com/monasticpanic/status/974077689400385536) and [Styled Components](https://mobile.twitter.com/mxstbr/status/974014564630237189) maintainers. This problem is familiar to other HOC authors (e.g. React Redux) too how currently have to tell users to chain `getWrappedInstance()` calls any time a component needs to be connected but is already used as a ref.

>But in #30 (comment) we end up with 2 "components" Header and ForwardHeader and a different meaning of ref. Maybe it improves backwards compatibility, but future users are required to know the difference between the two.

You’d export just one of them. Just like when you use HOCs, you don’t export all intermediate versions of the component. This is about better encapsulation and exposing *less* internals (HOCs can be more “transparent”), not exposing *more* internals.

>Okay, the longer I think about this, the more it becomes clear that this should not be taught in a React class. It is clearly a corner case feature, best kept hidden from beginners.

This is a power user feature for library authors. We didn’t say it’s appropriate for beginners. Even the article on *refs themselves* is outside the “beginner-friendly” Quick Start section in the React docs.

**Comment by rickbeerendonk** [comment]
@gaearon 

Thank you for the extensive answer.

Did you consider making `ref`-prop available inside components instead?

Two things to close (no answer needed):
- I do teach `ref` in a beginner course, since students are used to DOM manipulation and I don't want to make them feel like React is just one huge abstraction far away from what they know.
- HOC in the React world are hard to avoid. Beginners will be confronted by HOC in the projects they join. In my course I don't do a deep dive into HOC to explain them though.

**Comment by gaearon** [comment]
>Did you consider making ref-prop available inside components instead? -- I see now appearing your RFC that you did.

Yeah, that doesn’t really work because it’s an all-or-nothing approach. This RFC doesn’t have that problem.

>I do teach ref in a beginner course, since students are used to DOM manipulation and I don't want to make them feel like React is just one huge abstraction far away from what they know.

That sounds good. I still think ref forwarding is an even more advanced topic and primarily is useful to library developers.

>HOC in the React world are hard to avoid. Beginners will be confronted by HOC in the projects they join. In my course I don't do a deep dive into HOC to explain them though.

Right, but they’re unlikely to *write* them before they know React well. Ref forwarding is primarily useful to people *writing* HOCs. By the point they need to write one, you have to explain the problem with refs anyway, but at least now we’re offering a solution to that problem.

**Comment by bvaughn** [comment]
> Did you consider making ref-prop available inside components instead?

Assuming you mean `props.ref`- we did briefly talk about this. The main motivation _against_ it is that it would not work with the very common case of spreading props. (It would be too easy for the same `ref` to end up being passed to multiple components, unintentionally. That would be a very backwards-breaking change.)

**Comment by rickbeerendonk** [comment]
@bvaughn @gaearon 

I can see the issues with `this.ref`.

Component could also overwrite what `ref` is being returned by introducing a lifecycle method:

```js
function logProps(WrappedComponent) {
  return class extends React.Component {

    // New Lifecycle Method:
    // Allows user to overwrite the ref that is being returned from this component.
    componentReturnsRef(topRef) {
      // Defaulf would be: return topRef
      return this.wrappedRef;
    }

    componentWillReceiveProps(nextProps) {
      console.log('Current props: ', this.props);
      console.log('Next props: ', nextProps);
    }
    render() {
      return <WrappedComponent {...this.props} ref={ref => this.wrappedRef = ref} />;
    }
  }
}
```

bvaughn:
I'm not sure I understand what's being proposed/suggested above, but the proposal in this RFC also works for functional components (which don't have lifecycles- at least not at the moment). 😄

rickbeerendonk:
@bvaughn 

I was responding to @gaearon's 

> wrapping a component in a HOC always breaks refs

Seems like it is a problem that `ref` is not being exposed inside components (or the meganism not being open enough). I was just wondering (playing around to see) if you could expose `ref` just enough to fix the problem, instead of introducing external constructs to "fix" the limited visibility.

gaearon:
>I was just wondering (playing around to see) if you could expose ref just enough to fix the problem, instead of introducing external constructs to "fix" the limited visibility.

The approach you’re suggesting doesn’t work with functional components because they have no lifecycle methods. This limitation is unfortunate because we are considering moving away from class-based APIs in the future. So any APIs we’re adding today shouldn’t ideally be bound to classes unless they only extend class-only features (e.g. this is not a concern for `getDerivedStateFromProps` because it only makes sense for components with `state` anyway). Forwarding refs isn’t necessarily tied to classes.

Another problem with adding a lifecycle method is that we have to check each class component for it. We already do quite a bit of checks for lifecycle methods, and they chip away at performance for large updates. We’d like to avoid adding more such checks. For comparison, with the API in this RFC you only “pay” for this feature only when you actually use it.

Finally, I’m not sure I see how the API you proposed would work in practice (it’s a bit hard to think about). How would React know *when* to call `componentReturnsRef` and “propagate” the update upwards? You could pass a ref several layers deep and then give it to a leaf that mounts and unmounts independently of its parent (e.g. in response to a deep state change). In your example code, `this.wrappedRef = ref` assignment wouldn’t be sufficient for React to guess that the outer’s class *own* ref has changed, and now needs to also be updated. With the API in this RFC, this is not a problem because we are literally injecting the ref “from above” right into the component. So the ref given to the nested child is the same ref provided to the parent, not a wrapper.

jamesreggio:
Forgive me if this has already been covered — did a quick read-through, and didn't see anything.

I just want to confirm that the `ref` supplied as the callback argument within `React.forwardRef` is of the `RefObject<T>` type returned by `React.createRef()`.

For reference, `RefObject<T>` has the following signature:

```typescript
type RefObject<T> = {
  value: T | null
};
```

In some cases, its necessary for a wrapped component to both forward a ref and access it itself. I'm looking for assurance that the following code would work:

```js
class FancyTextbox extends Component {
  state = {
    value: '',
  };

  // In this contrived method, I'd like to be able to access
  // the same <input> ref that I'm forwarding.
  reset() {
    this.setState({ value: '' }, () => {
      if (this.props.forwardedRef.value) {
        this.props.forwardedRef.value.focus();
      }
    });
  }

  render() {
    return (
      <div>
        <input
          type="text"
          ref={this.props.forwardedRef}
          value={this.state.value}
          onChange={(value) => this.setState({ value })}
        />
      </div>
    );
  }
}

export default React.useRef((props, ref) => (
  <FancyTextbox {...props} forwardedRef={ref} />
));
```

The above example is _exceedingly_ contrived, but the actual usage I have in mind concerns binding native `Animated` events in React Native, which uses an imperative API that needs to be invoked every time the `ref` changes. (Such code is best tucked away in a HOC, hence the need for both ref forwarding and internal access.)

bvaughn:
> I just want to confirm that the `ref` supplied as the callback argument within `React.forwardRef` is of the `RefObject<T>` type returned by `React.createRef()`.

No. The `ref` argument could also be a [callback ref](https://deploy-preview-587--reactjs.netlify.com/docs/refs-and-the-dom.html#callback-refs).

jamesreggio:
Why is setting `.current` legal? That's not part of the `createRef()` spec, right?

How resistant would you be to normalizing the type of `ref` to either a callback ref or `RefObject<T>`? If this API is borne of a wish to simplify and normalize nested ref handling, it'd be a shame if it ended up trading old boilerplate for new (i.e., trading manual, proprietary ref unwrapping with the code above, which will need to be aware of all possibly-legal ref values, and is thus not future proof).

I feel like this RFC + the `createRef()` RFC gets us _so_ close to avoiding all this unnecessary ref munging.

dantman:
Forcing refs to be a `RefObject<T>` would mean complex use cases that we tell people to use callback refs for would not work with this RFC.

Rather than forcing a single type of ref, if it's important for user code be able to essentially set a ref, we should expose a helper for that instead of restricting to a single type.

i.e. A `React.???(ref:RefObject<T>|CallbackRef<T>, T)` that would either set the ref object's value or call the callback ref.

jamesreggio:
I'm perfectly fine with @dantman's approach, too.

I missed the boat on the `createRef()` RFC, but I think it was an oversight for the `RefObject<T>` to not contain a `setValue` method.

Regarding the use cases served by a callback `ref`, they can still be accomplished via a `RefObject<T>` (albeit in a convoluted manner) if you work off the assumption that `componentDidUpdate` will always be invoked after a change in ref.

```js
class Wrapper extends Component {

  lastRef = this.props.wrappedRef.value;

  componentDidUpdate() {
    const { lastRef } = this;
    const { value: nextRef } = this.props.wrappedRef.value;

    if (lastRef !== nextRef) {
      // Do whatever you would've done in your callback ref.
      this.lastRef = nextRef;
    }
  }

  // ...
}

// ref is of type RefObject<T>
export default React.useRef((props, ref) => (
  <Wrapper {...props} forwardedRef={ref} />
));
```

bvaughn:
> Why is setting `.current` legal? That's not part of the `createRef()` spec, right?

`createRef` has been updated (in alpha 2) to use `.current` instead of `.value` because the team had concerns that cases like `inputRef.value.value` would be confusing.

The [updated example](https://github.com/reactjs/rfcs/pull/30#issuecomment-373556227) by @jamesreggio does not handle the callback ref case, so it would not work. Setting that aside, using the property initializer case for extracting a `createRef` ref wouldn't work either, as the ref wouldn't be available until after the component mounted.

gaearon:
I agree it would be nice to have a built-in bulletproof to compose refs (the result would always have to be a functional ref). I don’t feel like it has to be a part of this proposal though. It is also not a fundamentally new capability so perhaps it could live outside of React (I know at least one library that does this although it’ll need to be updated for object refs.) Maybe a separate RFC?

jamesreggio:
> The updated example by @jamesreggio does not handle the callback ref case, so it would not work. 

The point of this example was to show how to attain the flexibility of a callback ref if `React.forwardRef` _only_  provided instances of `RefObject<T>`. It hinges upon that assumption.

> createRef has been updated (in alpha 2) to use .current instead of .value because the team had concerns that cases like inputRef.value.value would be confusing.

Name change aside, it sounds like you're saying that it's safe and supported to write to the `RefObject<T>.current` property. 👍 

----

I'm fine punting the normalization concerns to user-space or a separate RFC, but in order to do so, I want to explicitly confirm what has been implicit to this conversation. Specifically:

**The `ref` value supplied by `React.forwardRef` to the callback is the exact value of `ref` passed into the outer component returned by `React.forwardRef`.**

This guarantee means that we can assume that the value of `ref` is one of the known, valid ref formats, and we can behave accordingly (e.g., by invoking callback refs or by updating `.current` on boxed refs).

(I feel like there may be certain internal implementation conveniences or optimizations that could be leveraged if the `ref` provided by the `forwardRef` function were opaque, but those would have to be deferred to a future major release.)

gaearon:
>it sounds like you're saying that it's safe and supported to write to the RefObject<T>.current property

Yes. (When you want to manage it manually, such as the case when composing refs.)

>The ref value supplied by React.forwardRef to the callback is the exact value of ref passed into the outer component returned by React.forwardRef.

Yes. (Intentionally because always “normalizing” it to some form might be an unnecessary cost in most cases.)

bvaughn:
As I think about the use-case you raise, @jamesreggio, I find myself warming up to the idea of a new package that would help compose both ref-types. I'll add it to my list of things to write an RFC for (unless you beat me to it).

text/0000-ref-forwarding.md
```diff
@@ -0,0 +1,189 @@
...
+
+# Summary
+
+Provide an API that enables [`refs`](https://reactjs.org/docs/refs-and-the-dom.html) to be forwarded to a descendant (child or grandchild component).
+
+# Motivation
+
+I recently began contributing to [`react-relay` modern](https://github.com/facebook/relay/tree/master/packages/react-relay/modern) in order to prepare it for the [upcoming React async-rendering feature](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html).
+
+One of the first challenges I encountered was that `react-relay` depends on the legacy, unstable context API, and values passed through legacy `context` aren't accessible in the new, [static `getDerivedStateFromProps` lifecycle](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md#static-getderivedstatefrompropsnextprops-props-prevstate-state-shapestate--null). The long-term plan for `react-relay` is to use the [new and improved contex API](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md) but this can't be done without dropping support for older versions of React- (something `react-relay` may not be able to do yet).
```
ivikash: new and improved `contex` API -> new and improved `context` API

text/0000-ref-forwarding.md
```diff
@@ -0,0 +1,189 @@
...
+
+# Drawbacks
+
+This API increases the surface area of React slightly, and may complicate compatibility efforts for react-like frameworks (e.g. `preact-compat`). I believe this is worth the benefit of having a standardized, transparent way to forward refs.
+
+# Alternatives
+
+1. Add a new class method, e.g. `getPublicInstance` or `getWrappedInstance` that could be used to get the inner ref. (Some drawbacks listed [here](https://github.com/facebook/react/issues/4213#issuecomment-115019321).)
+
+2. Specify a ["high-level" flag on the component](https://github.com/facebook/react/issues/4213#issuecomment-115048260) that instructs React to forward refs past it. This appraoch could enable refs to be forwarded one level (to the immediate child) but would not enable forwarding to deeper child, e.g.:
```
ivikash: `appraoch` -> `approach`

text/0000-ref-forwarding.md
```diff
@@ -0,0 +1,189 @@
...
+          // ref belongs here
+          <ThemedComponent {...props} ref={props.componentRef} theme={value} />
+        )}
+      </ThemeContext.Consumer>
+    );
+  };
+}
+```
+
+3. [Automatically forward refs for stateless functions components](https://github.com/facebook/react/issues/4213#issuecomment-115051991). (React currently warns if you try attaching a `ref` to a functional component, since there is no backing instance to reference.) This appraoch would not enable class components to forward refs, and so would not be sufficient, since wrapper components often require class lifecycles. It would also have the same child-depth limitations as the above option.
```
ivikash: appraoch -> approach

AlmeroSteyn:
> Hi @AlmeroSteyn. Are you talking about this section of the React docs? If so, feel free to submit a PR at any point if you have ideas to improve the docs. 😄

@bvaughn Yes that is it exactly! I will definitely allign that doc in the near future.

elado:
I don't know if there are plans on making ref forwarding a part of React, but I just wrote a quick library that allows hoisting instance methods from inner refs. It uses context (old API), and some JS magic.

First draft here:
https://github.com/elado/react-ref-method-forwarder

bvaughn:
> I don't know if there are plans on making ref forwarding a part of React

That's what this RFC was for 😄and the feature is part of the upcoming 16.3 (currently available as an alpha on NPM).

bvaughn:
I think the final comment period is done for this PR.

elado:
@bvaughn @gaearon there's one use case I'm still not sure how `React.forwardRef` solves:

```jsx
class Field extends Component {
  focusInput() {
    this.input.focus()
  }

  render() {
    return (
      <div>
        <input ref={i => (this.input = i)} />
      </div>
    )
  }
}

class App extends Component {
  render() {
    return (
      <div>
        <Field ref={i => (this.field = i)} />
        <button onClick={() => this.field.focusInput()}>Focus</button>
      </div>
    )
  }
}
```

when Field is wrapped with some HOC from a different library:

```jsx
class Field extends Component {
  focusInput() {
    this.input.focus()
  }

  render() {
    return (
      <div>
        <label>{this.props.label}</label>
        <input ref={i => (this.input = i)} />
      </div>
    )
  }
}

// one or more HOCs from my code/other libraries
const WrappedField = someHOC(Field)

class App extends Component {
  render() {
    return (
      <div>
        <WrappedField ref={i => (this.field = i)} />
        {/* this will fail */}
        <button onClick={() => this.field.focusInput()}>Focus</button>
      </div>
    )
  }
}
```

In `App` I want `this.field` to point to the `Field` instance, not the `someHOC` instance. I can't change `someHOC` to forward refs.

Or even with Redux `connect` - I want to "skip" HOCs and access methods on my wrapped component. Specifically `focus`, `scrollTo` and methods that change UI state that is not controlled by React (e.g. setting `isFocused` on an element is not the solution because this UI state is ephemeral and user may blur it).

satya164:
> In App I want this.field to point to the Field instance, not the someHOC instance. I can't change someHOC to forward refs.

This is meant for library developers writing the HOC, not for consumers.

But you could write your own HOC which renames forwards the ref as some prop (say `forwardedRef`, and then in your component, use that when assigning the ref for `input`.

bvaughn:
@satya164 is right. `forwardRef` is meant more for use by the developer writing the HOC than one using it. Ideally, `someHOC` would use `forwardRef` to do this for you automatically.

Worst case, assuming `someHOC ` passes through `props` to your inner component, you can just intercept the ref and convert it to a prop, and maybe pass it through using something like this?

```js
// Extracts a forwared ref-prop and applies as a ref
function applyForwardedRefTo(Component) {
  return function ApplyForwardedRef({ forwardedRef, ...props }) {
    return <Component {...props} ref={forwardedRef} />;
  };
}

// Intercepts an incoming ref and forwards it as a prop
function forwardRefTo(Component) {
  function ForwardRefTo(props, ref) {
    return <Component {...props} forwardedRef={ref} />;
  }
  return React.forwardRef(ForwardRefTo);
}

const WrappedField = forwardRefTo(someHOC(applyForwardedRefTo(Field)));
```

Here's a CodeSandbox example: https://codesandbox.io/s/p7xnpvvpoj

elado:
@satya164 

> This is meant for library developers writing the HOC, not for consumers

This means that current non-maintained libraries will probably never get this feature

Instead of waiting for updates/submitting PRs, I think there should be a global, bulletproof solution.

That's why I came up with https://github.com/elado/react-ref-method-forwarder in the first place. 

`React.forwardRef` actually helps here, and still needs 2 HOCs (like `react-ref-method-forwarder`): 

```jsx
import React, { Component } from 'react'
// just a 3rd party library that exports a HOC
import dimensions from 'react-dimensions'

class Field extends Component {
  focusInput() {
    this.input.focus()
  }

  render() {
    return (
      <div>
        <input ref={i => (this.input = i)} />
      </div>
    )
  }
}

const forwardRefOuter = Wrapped => {
  return React.forwardRef((props, ref) => <Wrapped {...props} innerRef={ref} />)
}

// this *WORKS*, but throws a warning:
// Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
// const forwardRefInner = Wrapped => props => <Wrapped {...props} ref={props.innerRef} />


const forwardRefInner = Wrapped => {
  return class extends Component {
    render() {
      return <Wrapped {...this.props} ref={this.props.innerRef} />
    }
  }
}

export default forwardRefOuter(dimensions()(forwardRefInner(Field)))

/* or
@forwardRefOuter
@dimensions()
@forwardRefInner
class Field...
*/
```

gaearon:
>This means that current non-maintained libraries will probably never get this feature

This goes for any RFC. We’ve always been improving the way React apps and libraries are written in the future rather than adding retroactive hacks that get around an abstraction. If something is unmaintained it will likely bump into breaking changes between majors in any case. I understand it’s not always convenient, but it’s a larger discussion and unrelated to this particular API.

PR #31, opened 2018-03-08
[WIP] ES Decorators and React

AlicanC:
A WIP RFC to discuss how ES Decorators can benefit React ecosystem.

### DIFF
--- a/text/0000-decorators.md
+++ b/text/0000-decorators.md
@@ -1,0 +1,155 @@
+- Start Date: 2018-02-21
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This is a WIP RFC to discuss how we can develop the React ecosystem with the upcoming ECMA262 feature "Decorators".
+
+[Decorators](https://github.com/tc39/proposal-decorators) is an ECMA262 proposal at stage 2 and yet to be standardised.
+
+**Please note that decorators proposal have been through breaking changes. Please read the current version if you were unaware: [Link to Decorators Proposal](https://github.com/tc39/proposal-decorators)**
+
+While it's not yet a standard, how they could be used in React ecosystem should be discussed for benefit of the React community and the standardisation process.
+
+Just like React uses JSX at its core, it could also make use of another language future, decorators to enable better development experiences.
+
+# Examples
+
+## React Use: Component Flags
+
+Instead of extending a separate class or nesting components for flagging a component, decorators can be used:
+```javascript
+// Before:
+export default class MyComponent extends React.PureComponent {
+  return (
+    <AsyncMode>
+      {/* ... */}
+    </AsyncMode>
+  )
+}
+
+// After:
+@React.pure
+@React.async
+export default class MyComponent extends React.Component {}
+```
+
+## React Use: Refs
+
+A field decorator can be implemented to allow easier use of Refs:
+```javascript
+class MyComponent extends Component {
+  @React.ref
+  myTextInput;
+
+  myMethod = () => {
+    foo(this.myTextInput.value);
+  }
+
+  render() {
+    return (
+      <TextInput ref={this.refs.myTextInput} />
+    )
+  }
+}
+```
+
+## Community Use: HOCs
+
+Many libraries like Redux and Relay provide additional Props to Components. They do this by wrapping the Component in another Component. Class decorators can be used to simplify consumption.
+
+```javascript
+// Before:
+import { withRouter, connect } from './hocs';
+
+class MyComponent extends Component { /* ... */ }
+
+const MyComponentWithRouter = withRouter(MyComponent);
+
+export default connect(/* ... */)(MyComponentWithRouter);
+
+// After
+import { withRouter, connect } from './hocs';
+
+@withRouter
+@connect(/* ... */)
+export default class MyComponent extends Component { /* ... */ }
+```
+
+## Community Use: Providing Mock Data
+
+A decorator can be implemented for attaching mock data to a Component to help out-of-context rendering:
+
+```javascript
+type Props = {
+  showDetails: string, // From Redux
+  user: User, // From Relay
+}
+
+@MyLib.mock({
+  showDetails: false,
+  user: MyLib.fakeUser,
+})
+class MyComponent { /* ... */ }
+```
+
+Out-of-context rendering enabled by using such a decorator can allow design tools (like Facebook Origami and Adobe XD) or IDEs (like Atom and VSCode) to work on individual files.
+
+# END OF DOCUMENT, THE REST WILL BE FILLED LATER
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+Please focus on explaining the motivation so that if this RFC is not accepted,
+the motivation could be used to develop alternative solutions. In other words,
+enumerate the constraints you are trying to solve without coupling them too
+closely to the solution you have in mind.
+
+# Detailed design
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with React to understand, and for somebody familiar with the
+implementation to implement. This should get into specifics and corner-cases,
+and include examples of how the feature is used. Any new terminology should be
+defined here.
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people React
+- integration of this feature with other existing and planned features
+- cost of migrating existing React applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Adoption strategy
+
+If we implement this proposal, how will existing React developers adopt it? Is
+this a breaking change? Can we write a codemod? Should we coordinate with
+other projects or libraries?
+
+# How we teach this
+
+What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing React patterns?
+
+Would the acceptance of this proposal mean the React documentation must be
+re-organized or altered? Does it change how React is taught to new developers
+at any level?
+
+How should this feature be taught to existing React developers?
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?


AlicanC:
A recent tweet by @acdlite ([read here](https://twitter.com/acdlite/status/971598256454098944)) discusses a new pattern. That pattern and this RFC both covers a way to provide extra data to components so whoever has interest in this RFC should also read that.

AlicanC:
Other ways of providing extra data are [HOCs](https://reactjs.org/docs/higher-order-components.html) and [Render Props](https://reactjs.org/docs/render-props.html)

There are old-ES-Decorators-style HOCs (`(Config) => (ComponentType) => ComponentType`):
- [react-redux](https://github.com/reactjs/react-redux) `connect()`
- [recompose](https://github.com/acdlite/recompose)

And also Relay HOCs (`(Config, ComponentType) => ComponentType`):
- [react-relay](https://github.com/facebook/relay/tree/master/packages/react-relay)

text/0000-decorators.md
```diff
@@ -0,0 +1,155 @@
...
+    <AsyncMode>
+      {/* ... */}
+    </AsyncMode>
+  )
+}
+
+// After:
+@React.pure
+@React.async
+export default class MyComponent extends React.Component {}
```
streamich: This can be done by

```js
export default React.pure(React.async(MyComponent));
```

text/0000-decorators.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+## React Use: Refs
+
+A field decorator can be implemented to allow easier use of Refs:
+```javascript
+class MyComponent extends Component {
+  @React.ref
+  myTextInput;
```
streamich: Isn't it better to just assign the value?

```js
myTextInput = React.ref();
``

File: text/0000-decorators.md [hunk 3]
```diff
@@ -0,0 +1,155 @@
...
+
+  myMethod = () => {
+    foo(this.myTextInput.value);
+  }
+
+  render() {
+    return (
+      <TextInput ref={this.refs.myTextInput} />
```
streamich: Do we need `.refs` here?

```js
<TextInput ref={this.myTextInput} />
```

text/0000-decorators.md
```diff
@@ -0,0 +1,155 @@
...
+const MyComponentWithRouter = withRouter(MyComponent);
+
+export default connect(/* ... */)(MyComponentWithRouter);
+
+// After
+import { withRouter, connect } from './hocs';
+
+@withRouter
+@connect(/* ... */)
+export default class MyComponent extends Component { /* ... */ }
```
streamich: Correct me if I'm wrong, but you can already do this.

text/0000-decorators.md
AlicanC: Decorators:
```javascript
@React.pure
@React.async
export default class MyComponent extends React.Component {
    // ...Maybe hundreds of lines...
}
```

Nested Calls:
```javascript
    // ...Maybe hundreds of lines...
}

export default React.pure(React.async(MyComponent));
```

Nested Calls are harder to write and read, produce worse diffs and requires you to scroll through possibly hundreds of lines to be seen.

Whatever can be done with decorators can already be done in one way or another. I just believe that decorators are easier to use. Consider it sugar.

yordis: @streamich everything can be done by functions calls at the end of the day 

```es6
const NavigationStyled = withStyles(styles)(NavigationComponent);
const NavigationStyledConnected = connect(mapStateToProps, mapDispatchToProps)(NavigationStyled);
const NavigationStyledConnectedWithRouter = withRouter(NavigationStyledConnected)

export { NavigationStyledConnectedWithRouter as Navigation }
```

vs

```es6
@withStyles(styles)
@connect(mapStateToProps, mapDispatchToProps)
@withRouter()
class Navigation extends React.Component { }
export Navigation
```

I would prefer to focus on how this would improve our code base and workflows. Everything could be a function call at the end of the day, but fundamentality having something like this would improve the coding.

P.S: despite the current decorator state and opinions about mutating things inside or not, decorators are not evil, engineers are evil.

text/0000-decorators.md
AlicanC: (Also correct me if I'm wrong, but) ES Decorators spec has changed and their current implementations are no longer valid decorators. Current decorators do not return a new class. (AFAIK, finalising the decoration with a new class is still possible though.)

text/0000-decorators.md
AlicanC: `myTextInput` is typed `?TextInput`. `this.refs` contains hints for every property decorated with `@React.ref`. React uses those hints to automatically build a ref callback like `(ref) => { instance[propertyName] = ref; }`.

Naming could be different or this could be a stupid idea after all.

text/0000-decorators.md
AlicanC: See the other comment for explanation.

TypeScript has support for old ES Decorators and it's under a flag. They jumped the gun and implemented a non-standard and fed it to people through Angular. TypeScript's mess doesn't and shouldn't affect ECMA262 or React ecosystems.

text/0000-decorators.md
streamich: @yordis 

You might be interested in [pipeline operator](https://github.com/tc39/proposal-pipeline-operator).

P.S. I like decorators and [use them all the time](https://github.com/MailOnline/libreact#contents), I just don't really understand what is the discussion about? If you like `@withRouter` as a decorator, you can already use it. Or is about bringing decorators into React core?

AlicanC:
I'd like to note that the main aim of the RFC is talk about not React's code but its ecosystem. I can imagine many conclusions for this RFC:
- React starts providing decorator APIs.
- Libraries start providing decorator APIs.
- We write documentation to tell users how to use decorators. (Like there is for HOCs and Render Props)
- All, some or none of the above.

While I follow many well-considered React people, I haven't seen anyone talking about decorators and how they can (or maybe can't) benefit React.

I just wanted to start a discussion and get people thinking with decorators. I will edit the RFC as we further the discussion.

I believe this will be an easier discussion after Stage 2 Decorators support for Babel 7 is complete: https://github.com/babel/proposals/issues/11

Until then it will not be possible to provide POCs.

text/0000-decorators.md
AlicanC: You are using Stage 0 Decorators. TypeScript doesn't have Stage 2 Decorators.

vcarl:
As you note, Decorators are stage 2. I believe the main reason they haven't been discussed more broadly is because they're still viewed as unstable and likely to cause churn. The confusion between stage 0 and stage 2 decorators in the comments above seems to be evidence to that point, it seems very possible that they'll undergo further changes if and when they progress to stage 3.

And if it's not currently possible to provide POCs because they're not supported anywhere, then that again seems like evidence that this conversation should be happening sometime in the future.

AlicanC:
> it seems very possible that they'll undergo further changes

That's the important part. We can either wait until TC39 finalises the spec, or shape the spec by thinking on valid use cases. It would be great if we could discuss how we could benefit from decorators and provide TC39 with constructive inputs.

text/0000-decorators.md
streamich: I use whatever decorators TypeScript supports, but thanks for the info, will look into the differences.

yordis: @streamich as an Elixir developer I definitely prefer the pipe operator but that is not my point.

yordis:
> As you note, Decorators are stage 2. I believe the main reason they haven't been discussed more broadly is because they're still viewed as unstable and likely to cause chur

@vcarl the reason why I would love React team to talk about this so the specification moves faster and we can benefit from this feature.

From my code example, either pipe operator or decorators but the current way to do is so ugly and easy to error prune to be honest.

AlicanC:
## HOCs with Decorators

**HOC (Note: This is NOT a valid Stage 2 Decorator)**

```javascript
const injectFoo = (value) => (Base) =>
  class Wrapper extends Component {
    render() {
      return <Base foo={value} />;
    }
  };
const injectBar = /* ... */;
const injectBaz = /* ... */;

  render() {
    const { foo, bar, baz } = this.props;
  }
}

export default injectFoo('foo')(injectBar('bar')(injectBaz('baz')(MyComponent)));
```

**HOC w/ Stage 2 Decorators**
```javascript
const injectFoo = (value) => (classDescriptor) => ({
  ...classDescriptor,
  // Finisher is a HOC
  finisher: (Base) =>
    class Wrapper extends Component {
      render() {
        return <Base foo={value} />;
      }
    },
});
const injectBar = /* ... */;
const injectBaz = /* ... */;

@injectFoo('foo')
@injectBar('bar')
@injectBaz('baz')
export default class MyComponent extends Component {
  render() {
    const { foo, bar, baz } = this.props;
  }
}
```

# Preventing Prop Name Collision with Property Decorators 

Apollo has a HOC named "[graphql](https://www.apollographql.com/docs/react/basics/queries.html#basics)" and it injects a prop named "data". To prevent collisions you have to pass an options object for each use:

```javascript
import { graphql } from 'react-apollo';

  render() {
    const { foo, bar, baz } = this.props;
  }
}

export default compose(
  graphql(foo, { name: 'foo' }),
  graphql(bar, { name: 'bar' }),
  graphql(baz, { name: 'baz' }),
)(MyComponent);
```

This could be simplified with property decorators:

```javascript
import { graphql } from 'react-apollo';

const graphqlProp = (query, options) => (elementDescriptor) => {
  const propName = elementDescriptor.key;
  
  return {
    ...elementDescriptor,
    propertyDescriptor: {
      ...elementDescriptor.propertyDescriptor,
      writable: false,
      get() {
        return this.props[propName];
      },
    },
    // Finisher is a wrapped HOC
    finisher: (Base) => graphql(query, {
      ...options,
      name: propName,
    })(Base),
  }
};

export default class MyComponent extends Component {
  @graphqlProp(foo) foo;
  @graphqlProp(bar) bar;
  @graphqlProp(baz) baz;

  render() {
    // Both work
    const { foo, bar, baz } = this.props;
    const { foo, bar, baz } = this;
  }
}
```

text/0000-decorators.md
```diff
@@ -0,0 +1,155 @@
...
+# Examples
+
+## React Use: Component Flags
+
+Instead of extending a separate class or nesting components for flagging a component, decorators can be used:
+```javascript
+// Before:
+export default class MyComponent extends React.PureComponent {
+  return (
+    <AsyncMode>
```
gaearon: Note that AsyncMode *applies to the whole subtree*. It's not specific to a component. You'd normally only use it once at the top. So I think it's not very relevant to this proposal.

text/0000-decorators.md
AjayPoshak: Pipe operators are nice, but still,  I would prefer to use decorators at class level. I think that pipe operators are more useful for minor operations such as helper function calls, but they would look horrible at class level.

adrianhelvik:
I use decorators a lot for providing access to some context, such as confirm dialogues and status messages. A nice bonus is that they can render differently in different contexts. Also, I prefer it to using render props. With render props I don't have access to the received state from `this`. Would love to have
```
@withContext('foo', MyContext)
class Foo extends Component {
  render() {
    return (
      <div>{this.props.foo.bar}</div>
   )
  }
}
```

gaearon:
https://github.com/reactjs/rfcs/pull/68

PR #32, opened 2018-03-08
Introducing an RFC for isomorphic IDs

n8mellis:
This pull request introduces the first draft of an RFC to address the ability for React to support isomorphic IDs; a feature which is highly desirable for building WCAG 2.0 compliant web sites and applications.

For more information about the motivation behind this pull request, please see the following issue:

  https://github.com/facebook/react/issues/5867

### DIFF
--- a/text/0000-isomorphic-ids.md
+++ b/text/0000-isomorphic-ids.md
@@ -1,0 +1,315 @@
+- Start Date: 2018-03-08
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Many accessibility features of HTML require linking elements using IDs. There is
+currently no easy and consistent way to programmatically handle these IDs in a 
+way that survives the client hydration process when server-side rendering is 
+involved. 
+
+The key issue to solve here is a way to express relationships between 
+HTML elements that are identified by the ID attribute in a way that allows that 
+relationship to be valid and deterministic on both the server-side and 
+client-side render of the tree.
+
+# Basic example
+
+Consider a simple Checkbox component.
+
+```js
+// Checkbox.js
+class Checkbox extends React.Component {
+  
+  render() {
+    const { label, value, name, disabled } = this.props;
+    const id = generateUniqueId();
+    
+    return (
+      <div>
+        <input
+          id={id}
+          type="checkbox"
+          name={name}
+          value={value ? value : label}
+          disabled={disabled || false}
+        />
+        <label htmlFor={id}>{label}</label>
+      </div>
+    );
+  }
+  
+}
+
+// App.js
+<Checkbox label="My Label" value="my-label" name="MyLabel" disabled={false} />
+```
+
+The current state of the isomorphic world would result in something to the 
+effect of:
+
+```
+// Browser console.
+Warning: prop `id` did not match. Server: "Checkbox:..." Client: "Checkbox:...."
+```
+
+```html
+<!-- Server output -->
+<div>
+  <input
+    id="unique-id-1"
+    type="checkbox"
+    name="MyLabel"
+    value="my-label"
+  />
+  <label htmlFor="unique-id-1">My Label</label>
+</div>
+
+<!-- Client output -->
+<div>
+  <input
+    id="unique-id-3"
+    type="checkbox"
+    name="MyLabel"
+    value="my-label"
+  />
+  <label htmlFor="unique-id-3">My Label</label>
+</div>
+```
+
+**The goal of this RFC is to ensure that there is a way to guarantee that the 
+IDs used by the server and client are the same.**
+
+# Motivation
+
+In order to successfully link the `label` with the `input`, the `input` element
+must have an ID and that same ID must be supplied to the `label` in the `html-for`
+attribute.
+
+From this component's perspective, the relationship between these two elements 
+is always encapsulated inside the component. If only considering client-side 
+rendering, the simple solution would be to generate a random, unique ID, store 
+it as an instance variable, and assign that value to both elements.
+
+This procedure breaks down when introducing server-side rendering because the 
+value will be different between the server's execution and the client's execution
+resulting in an invariance violation (see the Basic Example above).
+
+To work around this, the developer must break encapsulation and supply a unique
+value for every instance. This introduces additional human input and puts an 
+unnecessary strain on developers to consider and coordinate uniqueness of IDs 
+for these types of components.
+
+There should be no need for users of this component to consider global uniqueness
+when the component itself should know everything it needs to know to render 
+correctly.
+
+# Detailed design
+
+There are probably several ways that this could be addressed. The simplest from a 
+conceptual standpoint would be to ensure that the render path is deterministic so
+that an auto-increment ID generation approach would result in the same number 
+being assigned to each ID.
+
+This is not a great solution. First off, it will probably result in a significant
+performance penalty which would be unacceptable. Secondly, it does not take into 
+account potential (intentional) differences in what is rendered server-side and 
+client-side.
+
+A more robust solution would be to take advantage of the hydration process to 
+absorb attributes included in the server-rendered output into the hydrated DOM 
+before it is merged into the browser's DOM.
+
+Let us suppose that we had a way to reserve an identifier from React. Something
+akin to the following:
+
+```js
+// Checkbox.js
+class Checkbox extends React.Component {
+  
+  constructor() {
+    this.reserveUniqueIdentifier(); // <------------------------------------NEW
+  }
+  
+  render() {
+    const { label, value, name, disabled } = this.props;
+    const id = this.reservedIdentifier(); // <----------------------------- NEW
+    
+    return (
+      <div>
+        <input
+          id={id}
+          type="checkbox"
+          name={name}
+          value={value ? value : label}
+          disabled={disabled || false}
+        />
+        <label htmlFor={id}>{label}</label>
+      </div>
+    );
+  }
+  
+}
+
+// App.js
+<Checkbox label="My Label" value="my-label" name="MyLabel" disabled={false} />
+```
+
+On the server, this component would now be rendered as:
+
+```html
+<div data-react-reserved-id="6CB96365-C679-415A-8FDE-8CD52C2AA678">
+  <input
+    id="6CB96365-C679-415A-8FDE-8CD52C2AA678"
+    type="checkbox"
+    name="MyLabel"
+    value="my-label"
+  />
+  <label htmlFor="6CB96365-C679-415A-8FDE-8CD52C2AA678">My Label</label>
+</div>
+```
+
+> NOTE: Calling `this.reservedIdentifier` without (or before) calling 
+> `this.reserveUniqueIdentifier` should result in an exception being thrown.
+
+By calling `this.reserveUniqueIdentifier` in the constructor, an additional 
+property can be added to the object to tag it as having a reserved ID. This will
+be useful later for quickly determining if additional steps are needed during 
+hydration.
+
+Example:
+```js
+const checkbox = <Checkbox label="My Label" value="my-label" name="MyLabel" disabled={false} />;
+console.log(checkbox.hasReservedUniqueIdentifier);
+// => true
+```
+
+During hydration, whenever a component is encountered that has a reserved ID, 
+it is scheduled for an update before the shadow DOM is merged with the browser's
+DOM. That element's position in the tree is determined and the corresponding 
+element in the browser's DOM is retrieved, the `data-react-reserved-id` attribute
+is read, and its value is set as the reserved ID for the hydrated component. 
+Then that component is enqueued for an update. Finally, after all components with
+reserved IDs have been updated, the hydrated DOM can be merged with the browser's 
+DOM in the usual fashion.
+
+In the case where a corresponding element is not found in the browser's DOM 
+(meaning that is was omitted during server-side render), then a new unique ID 
+is generated and that new ID is used by the hydrated component.
+
+Furthermore, since this process is part of the hydration process then these 
+additional checks will not be performed if not intentionally hydrating SSR output
+ensuring that there is no impact on the `render` path.
+
+### Step-by-step procedure
+
+#### Server procedure
+
+1. `ReactDOM.renderToStaticMarkup` is called.
+1. A `Checkbox` component is encountered.
+1. The `Checkbox` constructor calls `this.reserveUniqueIdentifier()`.
+1. `reserveUniqueIdentifier` determines that it is running on the server using
+   `fbjs/lib/ExecutionEnvironment`.
+1. A unique ID is generated (for example using `UUIDv4`) and set to a private
+   property on the instance. (for example `this.__reservedIdentifier`)
+1. Property `hasReservedUniqueIdentifier` is set to `true` on the instance.
+1. `Checkbox` component's `render` method is called.
+1. The reserved ID is retrieved by the `this.reservedIdentifier()` method
+   that retrieves the value of the private instance variable.
+1. The `data-react-reserved-id` HTML attribute is added to the root element 
+   of the component with the value of `this.reservedIdentifier()`.
+1. HTML is sent to the client.
+
+#### Client procedure
+
+1. `ReactDOM.hydrate` is called.
+1. A data structure is created to store references to elements that have reserved
+   identifiers.
+1. A component where `hasReservedUniqueIdentifier` is `true`.
+1. A reference to the Component instance is stored in the above data structure.
+1. After the whole client DOM has been processed:
+  1. Iterate over instances stored in the above data structure.
+  1. Determine the DOM path of the element in the shadow DOM.
+  1. Look up the equivalent DOM path in the browser's DOM.
+  1. If an element is found:
+    1. Retrieve the value of `data-react-reserved-id`.
+  1. If an element is not found:
+    1. Generate a new ID.
+  1. Set the value of the instance's `__reservedIdentifier` private property.
+  1. Enqueue that instance for an update.
+  1. Perform queued in the shadow DOM.
+1. Merge shadow DOM with browser DOM in normal fashion.
+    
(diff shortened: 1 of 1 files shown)


streamich:
Very nice, this is something very much needed, but can this be achieved by a single call to `React.createId()?

```js
class Checkbox extends React.Component {
  id = React.createId();

  render () {
    return [
      <input id={this.id} />,
      <label htmlFor={this.id} />,
    ];
  }
}
```

n8mellis:
@streamich I don't see `React.createId()` anywhere in the documentation or source code. Is this something that already exists? If so, it's hiding very well. :)

If this is a reference to `generateUniqueId()` in my RFC, then I should have been more clear that that was just a placeholder to a generic UID-generating function. I wasn't proposing that as an actual function name. I'll push an update that makes that more clear.

Generating the UID isn't the hard part. It's getting the UID that the server used into the client's render context.

streamich:
@n8mellis Sorry, I did not write it up well, what I was asking if the two methods `reserveUniqueIdentifier()` and `reservedIdentifier()` could be simplified to one method and if it could be done in constructor, to avoid calling `reservedIdentifier()` on every re-render.

n8mellis:
@streamich Ah. That makes much more sense. :) My thinking was that this ID be treated differently than passing in an ID prop and thus the different approach.

I split it into two so that all the heavy lifting happens in the constructor.  The `reservedIdentifier` method is just a property getter so it's no more expensive than any property access.  I made it a function rather than a straight property accessor to make sure that the private property is obscured and that there is little risk in it accidentally being overwritten or changed in unexpected ways.

That however is a personal preference and would be open to arguments in favor of something different.

lostpebble:
Thanks for this RFC, I think it would be extremely useful.

The main reason I need this is for deterministic incremented ids of my React Components which are rendered - so the same content is displayed both server and client side.

For example, I have "filler" images for some ad space when it's not used and I'd like to display different random images when the user loads the page. If I only had one on a page, it would be easier (as I could just use a single random seed generated in my state stores on server and hydrated in client). But I would generally have multiple Components which look something like this:

```
const fillerImageUrls = [
   "https://www.somewhere.com/filler-image-1.jpg",
   "https://www.somewhere.com/filler-image-2.jpg",
   "https://www.somewhere.com/filler-image-3.jpg",
   "https://www.somewhere.com/filler-image-4.jpg",
];

Component {
   render() {
     const uniqueIncrementedId = this.getUniqueIncrementedId();

     return <img src={fillerImageUrls[randomSeedFromStores + uniqueIncrementedId) % fillerImageUrls.length]}/>
   }
}
```

So, whereas this RFC is pushing for implementing a unique ID _string hash_ of sorts - I'd really also like the capability of getting an incremented integer number which is unique to each component as well.

I would hope this to be incremented according to the amount of a **single type** of component on the page. In other words, it would not be unique between different types of components (the count should start at zero for each new type of component on the page).

manuelbieh:
This is highly desirable! I dropped automatic id generation for `aria-labelledby` attributes in [a component](https://github.com/manuelbieh/react-minimal-dropdown/commit/22e288344954e56839a8a1109bda8cbb3e689d7b#diff-17f754083466fd12638e891008f502b2) recently because creating the IDs based on `Math.random()` on a per render basis of course caused React to show a warning that the client HTML does not match the server HTML. 

I couldn't figure out an easy solution to deterministically generate unique IDs without spending too much effort on this. Would be super nice to have a small helper function built-in.

I remember the good ol' `data-reactid` attribute from 0.14 which were unique within each mounted component. Maybe they could be revitalized for that, maybe optionally prefixed with either a custom string to increase the chance of being _really_ unique. 

Spontaneous idea:
```
class MyComp extends Component {
  id = this.dataReactId('optionalPrefix');
  render() {
    return (
      <div id={this.id}>...</div>
    }
  }
}
```

Result:
```
<div id="optionalPrefix_.0.0.1">...</div>
```

ryanflorence:
Could definitely use something like this.

For now I use a context with a `genId()` on it. The default context generates random IDs, but if you wrap the whole app in a Provider then it increments each time it's called.

```jsx
import React, { createContext } from "react";

// most things that we auto-id aren't server rendered, and are rendered into
// portals anyway, so we can get away with random ids in a default context. If
// people need to server render with auto-ids, they can wrap their app in an
// IdProvider
let genId = () =>
  Math.random()
    .toString(32)
    .subsr(2, 8);

let IdContext = createContext(genId);

// Apps can wrap their app in this to get the same IDs on the server and the
// client
class Provider extends React.Component {
  id = 0;

  genId = () => {
    return ++this.id;
  };

  render() {
    return (
      <IdContext.Provider value={this.genId}>
        {this.props.children}
      </IdContext.Provider>
    );
  }
}

let { Consumer } = IdContext;

export { Provider, Consumer };
```

gaearon:
What does building this into React achieves compared to @ryanflorence's solution?

AjayPoshak:
I think that React itself should unopinionated about generating and handling `id`.  It should be a community standard to generate and handle `id`.  The feature to generate unique `id` might be abused, as people would like to use it to fix unique keys error in a component list.

probablyup:
@ryanflorence if the component render order differs between server and client, won't this fail to match up? I think that's a possible situation with async mode on?

maranomynet:
@ryanflorence: @probablyup is right that auto-incrementers break when rendering is async or for some reason happens in different order on server and client.

diegohaz:
It would be really useful to have something like this:

```jsx
import { useId } from "react";

function Component() {
  const id = useId();
  return <div id={id} />;
}
```

Merri:
Last summer I started a project for providing a HOC to deal with this uniqueness issue. It hasn't had much use and I haven't pushed it as I didn't have time to finish it before today.

The issue with most proposed or shown solutions is that they often either work only client side, keep increasing forever on server side, or even leak memory. To account these issues:

- ID service must create same IDs for the same render.
- IDs that are no longer used must be cleared.
- Must expose extra method so that all generated IDs can be cleared after each server render (so that IDs don't keep piling up in server).

Here is a short sample on how `react-maria` works:

```jsx
// MyComponent.jsx
import React from 'react'
import { withUniqueId } from 'react-maria'

function MyComponent(props) {
    // uidTools is optional, but useful when you render a lot of elements that require unique IDs
    const uidTools = props.getMariaIdTools()
    const uids = uidTools.make(['hello', 'world'])

    // note: id is provided by withUniqueId; also, <MyComponent id="hi"> results into props.id = 'hi1'
    return (
        <ul id={props.id}>
            {uids.map(uid => <li key={uid}>My unique id is {uid}</li>)}
        </ul>
    )
}

export default withUniqueId({ identifier: 'my-component' })(MyComponent)
```

You can find [source and docs at GitHub](https://github.com/Merri/react-maria). I just hope I didn't brainfart anywhere since I haven't tested this in a production environment yet :)

As far as I can understand the good thing about this solution is that there is no need for a Provider so end user of a third party component that uses React Maria needs to do nothing extra - only use the component!

theKashey:
This feature is not possible. You just can't use _just_ use `generateUniqueId` in async or concurrent environments. __And SSR__.
In short - you should be able to generate the same "set" of IDs regardless of component/chunk load order. That order could be different - data for components arrive unpredictable, chunks got loaded unpredictable, async in a client, and usually sync on a server.

Solution:
1. Store counter in a context.
2. Read-on-mount and increment-on-read. That would provide a unique ID for a component during its lifetime. It will generate ids like __1, 2, 3, 4__.
3. __Nest__ counter for any async children. It will generate ids like __1.1, 1.2, 1.3__.
4. As a result - every component would get _stable_ id, regardless of its nature.

```js
<UIDReset> // sets `counter` to 0
     <UIDFork> // `forks` counter
       <AsyncLoadedComponent> // will render something in a second
          <UIDConsumer>
            { uid => <span>{uid} is unique </span>} // will render 1.1
          </UIDConsumer>
      </UIDFork>    
    </AsyncLoadedComponent>
     <UIDFork> // `forks` counter.
         <UIDConsumer>
           { uid => <span>{uid} is unique </span>} // would render 2.1
     </UIDFork>
 </UIDReset>
```
So, even if the render order of `AsyncLoadedComponent` is not known - Id is "stable". These "forks" are the missing parts.

Meanwhile it still could be "amost" as this RFC proposes
```js
const Form = () => {
  const uid = useUID(); // this is hook
  return (
    <>
     <label for={uid}>Email: </label>
     <input id={uid} name="email" />
    </>
  )
}
```

It's not a pseudo code, but https://github.com/thearnica/react-uid, I am using it for a while to pair labels and input on frontend, or generate consistent Ids between server and client.

r00dY:
I'd love this feature. Generating unique ID really leaves a lot to be desired from developer experience point of view right now.

My use case:

I'm working on a simple library with UI component. Its CSS is dependent on props. And the CSS uses media queries which means I can't use inline styles. So I generate random ID inside the component, add class names based on this random name and add inline `<style>` tag.

And ofc random id breaks with SSR.

I believe there's a strong motivation behind the approach I take:
1. I want simplest developer experience from my library. Just use the component and it should work. Without SSR or with it (like in next.js / gatsby).
2. I don't want to enforce any specific CSS-in-JS solution for user of my lib. If someone uses emotion, I don't want him to add styled-components only to use my simple lib. Library should be definitely independent of such things.

And now:

1. Solution with providers doesn't sound good imo, why would I force my user to wrap the app with my provider only because she wants to use my simple small library? We're going to get into a Provider hell. And developer experience (onboarding) of using my library will be very bad in SSR environments (like next.js / Gatsby).
2. Solutions based on providers and counters suffer from async issues.
3. Solution with using the fact that `componentDidMount` is called only on client side suffers from FOUC.

There's definitely sth missing here. Solution could be `useUID` hook or some more generic mechanism of passing data from server-side-rendered component to client-side-rendered version.

I work a lot with React in next.js / Gatsby environments (out-of-the-box SSR, SEO-critical stuff like e-commerce / websites) and I'd like to take a small lib and simply make it work without any extra hassle. And any solution that requires sth more than import component and use it is not an optimal developer experience.

theKashey:
@r00dY please don't do it. You don't need it.
What you need:
- generate any `hash` from a media query you want to apply. And that `hash` would be __stable__, and would stand anything, async stuff included.
- use `hash` as className.
- magic! you can deduplicate your styles - similar hashes mean similar rules.
- PS: styled-components are doing exactly the same. It's a proven solution.
- PPS: And it's much simpler, and does not require almost anything, like a _Provider Hell_.

eps1lon:
It looks like an implementation has been proposed: https://github.com/facebook/react/pull/17322

gaearon:
Thank you so much for writing this up!

We have merged an experimental implementation inspired by this proposal in https://github.com/facebook/react/pull/17322 and are going to try it at Facebook. It's not using *exactly* the same heuristics as this proposal but should address some of the issues that would come up when combining it with other features we care about (like Progressive Hydration).

We'll report back with our findings.

theKashey:
@gaearon - could you provide some clarification, please.
A new hook is based on a [single counter variable](https://github.com/facebook/react/pull/17322/files#diff-b116dadddb2aa3db16234e5b5e1f8047R1159), which is:
- would be not reset from one server render to another
- would be not incremented if some pieces would be not hydrated, opening the possibility to use the same ids twice on the page.
- "micro-frontends" gonna blow up, ie two Reacts bundled separately on one page would fight with each other.

gaearon:
> would be not reset from one server render to another

Would you mind opening an issue about this? 

> would be not incremented if some pieces would be not hydrated, opening the possibility to use the same ids twice on the page.

This doesn’t sound right to me. The client and the server intentionally use different prefixes so I don’t think they could clash. (We don’t attempt to match the IDs between client and server — instead, we correct them during hydration.)

> "micro-frontends" gonna blow up, ie two Reacts bundled separately on one page would fight with each other.

Also worth discussing in an issue.

theKashey:
Ok. So having different counters on the server and the client solves my first two questions. My bad, I've missed this point in the PR.
However, it would be still a good idea to reset the counter in test-renderer to support snapshots.

So the only problem left is with two reacts on one page. 
For example we have some legacy projects, some pieces of which are being replaced by "third-party" React components, and we could not guarantee that react versions between those pieces would be the same, and share the same counter. Even if they would - modules resolution could duplicate react per-dependency.

The _standard_ solution for this is `randomize`, as an explicit seed initialization, or `<Randomize>` component in our case, which might prefix all `id` inside it by some UID. It really does not matter how it does it - stuff should be just unique behind it, no matter what.
Such component would solve multi-frontend issues, as well as multi-backend or theoretical fragment cache, which is the same as multi-backend - results from different renders are somehow combined together.

gaearon:
@theKashey Let's discuss current implementation in a React issue. This would be a bit easier for us to trakc.

samvv:
Sorry for the noise, but is there any update on this 'experiment'? I can definitely see use-cases that would benefit from this RFC. For instance, I'm currently developing a system that could use unique IDs for every `<form />`, such that the server can automatically reverse-lookup the submission of an arbitrary React-generated form.

theKashey:
Thus we have to talk about `unique-unique-ids` as per this RFC and _(sequentially)_`uniquie-ids` implemented in facebook/react#17322

gaearon:
@samv We haven't even checked it into the internal codebase yet. :D I think we'll have something more concrete in the coming months.

text/0000-isomorphic-ids.md
```diff
@@ -0,0 +1,315 @@
...
+This RFC represents a high-level proposal for how the goal might be achieved. 
+Due to my unfamiliarity with the internals of React, there are undoubtedly still 
+many questions that would need to be answered. Some examples off the top of my 
+head:
+
+- Can the methodology described above for absorbing HTML attributes from the
+  server-rendered markup be accomplished during the hydration phase before the 
+  first DOM merge?
+
+- Would this approach cover all use cases?
```
eps1lon, line 311: This RFC does not mention how (dynamic) children in composite widgets would communicate their IDs. An example for this would be the [tabs widget](https://www.w3.org/TR/wai-aria-practices/#tabpanel) where a `[role="tabpanel"]` is labelled by its corresponding `[role="tab"]`.

Given the current implementation of `useOpaqueIdentifier` and the [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), we would need to pass the ids to the widget container:

```jsx
function App() {
  const tabIdOne = React.unstable_useOpaqueIdentifier();
  const panelIdOne = React.unstable_useOpaqueIdentifier();
  const tabIdTwo = React.unstable_useOpaqueIdentifier();
  const panelIdTwo = React.unstable_useOpaqueIdentifier();

  return (
    <React.Fragment>
      <Tabs defaultValue="one">
        <div role="tablist">
          <Tab id={tabIdOne} panelId={panelIdOne} value="one">
            One
          </Tab>
          <Tab id={tabIdTwo} panelId={panelIdTwo} value="one">
            One
          </Tab>
        </div>
        <TabPanel id={panelIdOne} tabId={tabIdOne} value="one">
          Content One
        </TabPanel>
        <TabPanel id={panelIdTwo} tabId={tabIdTwo} value="two">
          Content Two
        </TabPanel>
      </Tabs>
    </React.Fragment>
  );
}
```
-- [Full example](https://codesandbox.io/s/useopaqueidentifier-for-tabs-tbjgu)

If we could use `useOpaqueIdentifier` as prefix we could avoid boilerplate by creating a single opaque identifier per widget that is suffixed by the string representation of the tab value.

The following hack only works with ReactDOM.render:

```jsx
class SuffixedOpaqueIdentifier {
  constructor(opaqueIdentifier, suffix) {
    // private
    this.opaqueIdentifier = opaqueIdentifier;
    this.suffix = suffix;
  }

  toString() {
    return `${this.opaqueIdentifier}-${this.suffix}`;
  }
} 

function Tabs({ children }) {
  const id = React.unstable_useOpaqueIdentifier();
  const getTabsId = React.useCallback(
    value => {
      return new SuffixedOpaqueIdentifier(id, value);
    },
    [id]
  );

  return (
    <TabsContext.Provider value={getTabsId}>
      {children}
    </TabsContext.Provider>
  );
}

...
function useTabId(tabValue) {
  const getTabsId = React.useContext(TabsContext);
  return getTabsId(`tab-${tabValue}`);
}
```
-- https://codesandbox.io/s/useopaqueidentifier-as-a-prefix-for-tabs-kke2c?file=/src/App.js:1029-1596

eps1lon, line 311: Didn't think about using this RFC for threading before I openened https://github.com/facebook/react/issues/18594.

TL;DR: What about attributes that allow multiple ids (i.e. attributes allowing IDREFS) e.g. `aria-labelledby` or `aria-describedby`?

gaearon:
Hey all, there's been some concerns so I wanted to give a small update.

We *are* experimenting with a Hook for this because this space intersects with some of our ongoing work on progressive hydration. However, that Hook's API is not final in any way, and we definitely want to take another read through all the concerns posted so far before sending an RFC for our version. Initially, we planned to experiment with it way earlier, but due to the pandemic people who planned to integrate it had to switch to different projects, so the work has frozen for a while and we switched to other ongoing projects.

We expect to start integrating and testing our version soon, and after gaining insights from it, we will return to this RFC and any concerns raised in it.

(Our Hook shipped in `experimental` release with an `unstable_` prefix but as usual, it only means we're planning to test it internally, and doesn't mean that specific version would become stable.)

gaearon:
A small update that there hasn't been much changes since the last update. It's still on our list to continue experimenting with, but React 18 is still [in alpha](https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html). We expect to revisit this closer to the beta or after it.

gaearon:
An update: https://github.com/reactwg/react-18/discussions/111

gaearon:
https://github.com/facebook/react/pull/22644

gaearon:
Thank you for the proposal and discussion.

We've merged #212 which includes [`useId`](https://github.com/reactjs/rfcs/blob/main/text/0212-react-18.md#useid).

It fulfills the same role as this RFC, so let's close this one.

PR #33, opened 2018-03-10
New commit phase lifecycle getSnapshotBeforeUpdate()

bvaughn:
[View formatted RFC](https://github.com/bvaughn/rfcs/blob/new-commit-phase-lifecycles/text/0000-new-commit-phase-lifecycles.md)

### DIFF
--- a/text/0000-new-commit-phase-lifecycles.md
+++ b/text/0000-new-commit-phase-lifecycles.md
@@ -151,11 +151,11 @@ class ScrollingList extends React.Component<Props, State> {
 
 # Adoption strategy
 
-Since this lifecycle- and async rendering in general- is new functionality, adoption will be organic. Documentation and dev-mode warnings have already been created to encourage people to move away from render phase lifecycles like `componentWillUpdate` to commit phase lifecycles.
+Since this lifecycle- and async rendering in general- is new functionality, adoption will be organic. Documentation and dev-mode warnings have already been created to encourage people to move away from render phase lifecycles like `componentWillUpdate` in favor of commit phase lifecycles.
 
 # How we teach this
 
-Lifecycle documentation on the website. Add a before an after example (like above) to the [Update on Async Rendering](https://github.com/reactjs/reactjs.org/pull/596) blog post "recipes".
+Lifecycle documentation on the website. Add a before an after example (like [the one above](#basic-example)) to the [Update on Async Rendering](https://github.com/reactjs/reactjs.org/pull/596) blog post "recipes".
 
 # Unresolved questions
 
--- a/text/0000-new-commit-phase-lifecycles.md
+++ b/text/0000-new-commit-phase-lifecycles.md
@@ -4,26 +4,29 @@
 
 # Summary
 
-Add new "commit" phase lifecycle, `getSnapshotBeforeUpdate`, that fires during an update, before mutations are made. The return value from this lifecycle will be passed as a third parameter to `componentDidUpdate`.
+Add new "commit" phase lifecycle, `getSnapshotBeforeUpdate`, that gets called _before_ mutations are made. Any value returned by this lifecycle will be passed as a third parameter to `componentDidUpdate`.
 
 This lifecycle will be important for [async rendering](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html), where there may be delays between "render" phase lifecycles (e.g. `componentWillUpdate` and `render`) and "commit" phase lifecycles (e.g. `componentDidUpdate`).
 
 # Basic example
 
-Consider the use-case of preserving scroll position within a list as its contents are updated. The way this is typically done is to read `scrollHeight` during render (using `componentWillUpdate`) and then adjust it after the update has been committed (using `componentDidUpdate`).
+Consider the use-case of preserving scroll position within a list as its contents are updated. The way this is typically done is to read `scrollHeight` during render (`componentWillUpdate`) and then adjust it after the update has been committed (`componentDidUpdate`).
 
-This approach does not work with async rendering, since there might be a delay between these lifecycles during which the user might continue scrolling. The only way to safely do this with current lifecycles would be to _force a synchronous render_.
+Unfortunately this approach **does not work with async rendering**, because there might be a delay between these lifecycles during which the user continues scrolling. The only way to ensure an accurate scroll position was read would be to _force a synchronous render_.
 
-The solution to this is to introduce a new lifecycle that gets called during the commit phase _before_ mutations have been made to e.g. the DOM. For example:
+The solution is to introduce a new lifecycle that gets called during the commit phase _before_ mutations have been made to e.g. the DOM. For example:
 
 ```js
 type Snapshot = number;
 
-class ScrollingList extends React.Component {
+class ScrollingList extends React.Component<Props, State> {
   listRef = React.createRef();
 
-  getSnapshotBeforeUpdate(prevProps: Props, prevState: State): Snapshot | null {
-    // We are adding new items to the list.
+  getSnapshotBeforeUpdate(
+    prevProps: Props,
+    prevState: State
+  ): Snapshot | null {
+    // Are we adding new items to the list?
     // Capture the current height of the list so we can adjust scroll later.
     if (prevProps.list.length < this.props.list.length) {
       return this.listRef.value.scrollHeight;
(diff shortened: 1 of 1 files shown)


bvaughn:
componentWillUpdate and componentWillMount are called before render, and may be called for a render that is discarded because of a higher priority update. The newly proposed lifecycles would be called during the "commit" phase.

This draft docs page does a nice job in explaining the difference and why it matters to distinguish between the two: https://deploy-preview-587--reactjs.netlify.com/docs/strict-mode.html#detecting-unexpected-side-effects

Reusing the old lifecycles would not be a good option for a couple of reasons. We currently plan to support the old lifecycle functionality with an "UNSAFE_" prefix. Having similarly named methods with significant different behavior would cause confusion and could make updating from 16 to 17 significant more difficult.

We would also want these new lifecycles to be available in a minor release of 16- so people can begin upgrading to use async

gaearon:
Specifically an important distinction is that people expect to be able to use `componentWillMount` for setting the initial state, but by the point these lifecycles run it's already too late. So changing what `componentWillMount` would be too confusing given all the existing code and examples that assume it happens before render (and on the server).

gaearon:
I *do* wish we could recycle the existing names though since "will mount" and "will update" indeed seem like ideal names for these APIs 😢

gaearon:
I guess we can add them under new names and then eventually find better names for them when/if we ever introduce a new lifecycle API.

bvaughn:
> I do wish we could recycle the existing names though since "will mount" and "will update" indeed seem like ideal names for these APIs 😢

Yeah ☹️I had the same thought. It's unfortunate that those names are tainted.

I'm definitely open for bike-shedding on better names than the ones I proposed initially though. 😄

probablyup:
I like the idea of `componentWillFlush`, since that implies a stronger correlation with the "flush to DOM" part of the rendering flow.

Could we even perhaps just have a single event for both cases of mounting vs updating? I'm curious if the extra API surface is necessary in practice.

bvaughn:
I don't think a single lifecycle is a good idea, because `prevProps` and `prevState` would be null the first time it is called. I think the ergonomics of that might be a bit awkward.

gaearon:
Maybe

`componentWillFlushMount`
`componentWillFlushUpdate`

bvaughn:
I don't object too strongly to that, but I did intentionally avoid using "will" or "did" in the name because of the potential to cause confusion. With the (kind of awkward) caveat of `componentWillUnmount`, we're training people to be cautions of the "will" lifecycles, so I didn't feel great about adding new caveats.

acdlite:
Internally, we use the `prepare` prefix to refer to this phase

trueadm:
`componentDidRender`?

bvaughn:
`componentDidRender` was my first idea, actually 😁**but** I think it's important to have unique names for the did-mount case and the did-update case.

bvaughn:
Renamed `componentIsUpdating` to `getSnapshotBeforeUpdate` and dropped the concept of the before-mount lifecycle. Doesn't seem useful for any case I can think of.

TrySound:
Can `getSnapshotBeforeUpdate` be static? Refs can be kept in state.

bvaughn:
Briefly considered making it static but decided against it because of refs. Felt odd to force refs to be stored in state.

Swizec:
What about `componentWillCommit`? You mention a couple times that this happens during the *commit* phase before the component commits to the DOM.

flarnie:
One potential use case is in Draft.js, where we have two values we track that are outside of state and get updated roughly at the time this instance method would be called.

The current, short-term work-around we are doing is to [update them in the `componentDidMount` and `componentDidUpdate` of a child component](https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditor.react.js#L419).

I think this could be a possible solution for the Draft.js use case, @sophiebits would be curious what you think when you are online again.

I also think that for Draft.js we also will not need the 'before-mount' call. Seems like Draft wasn't setting those flags in 'willMount'. 

To be more specific about what Draft does by tracking those values - In one case it is intended to block stray 'select' events that fire in IE after Draft has already started a selection update, and in the other case it's caching the latest rendered state, so for both of those I don't think the initial mount is a problem.

dantman:
Reparenting (RFC https://github.com/reactjs/rfcs/pull/28) could make use of this feature. `<video>`/`<audio>` elements stop playing in some browsers, inputs lose their focus, and scrollable elements lose their scroll position when you move them from one dom node to another dom node.

If these pieces of state are really important to a user, they could be snapshotted before reparenting using this RFC and restored after they are moved to a new dom node.

sophiebits:
@flarnie For Draft I don't think we need any new lifecycles, instead maybe just we should refactor our componentDidMount calls so that we don't need to depend on the ordering as strongly. (That likely looks like the parent component controlling the selection handling instead of the leaves.)

bvaughn:
Whoops, forgot to replace `this.props` in one of the places.

text/0000-new-commit-phase-lifecycles.md
```diff
@@ -0,0 +1,167 @@
...
+    return (
+      <div ref={this.listRef}>{/* ...contents... */}</div>
+    );
+  }
+}
+```
+
+# Motivation
+
+This lifecycle provides a way for asynchronously rendered components to accurately read values from the host envirnment (e.g. the DOM) before it is mutated.
```
ivikash: `envirnment` -> `environment`

text/0000-new-commit-phase-lifecycles.md
```diff
@@ -0,0 +1,167 @@
...
+    return null;
+  }
+
+  // ...
+}
+```
+
+### No return value
+
+The proposed lifecycle will be the first commit phase lifecycle with a meaningful return value and the first lifecyle whose return value is passed as a parameter to another lifecycle. Likewise, the new parameter for `componentDidUpdate` will be the first passed to a lifecycle that isn't some form of `Props` or `State`. This adds some complexity to the API, since it requires a more nuanced understanding the relationship between `getSnapshotBeforeUpdate` and `componentDidUpdate`.
```
ivikash: `lifecyle` -> `lifecycle`

text/0000-new-commit-phase-lifecycles.md
```diff
@@ -0,0 +1,167 @@
...
+
+An alternative would be to scrap the return value in favor of storing snapshot values on the instance.
```
ivikash: `favor` -> `favour`

text/0000-new-commit-phase-lifecycles.md
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  // ...
+}
+```
+
+# Adoption strategy
+
+Since this lifecycle- and async rendering in general- is new functionality, adoption will be organic. Documentation and dev-mode warnings have already been created to encourage people to move away from render phase lifecycles like `componentWillUpdate` in favor of commit phase lifecycles.
```
ivikash: `favor` -> `favour`

text/0000-new-commit-phase-lifecycles.md
bvaughn: Both are valid spellings.

text/0000-new-commit-phase-lifecycles.md
bvaughn: Both are valid spellings.

bvaughn:
> Should we also mention about `shouldComponentUpdate` and how will its return value true or false will impact this?

This lifecycle will be invoked if `shouldComponentUpdate` doesn't prevent a new `render`, just like `componentDidUpdate`. I don't know if this needs to be called out explicitly?

text/0000-new-commit-phase-lifecycles.md
ivikash: 👍

bvaughn:
I've created a potential polyfill for this method with reactjs/react-lifecycles-compat/pull/1 but I don't think it's worth merging because of the [reasons mentioned in the RFC](https://github.com/bvaughn/rfcs/blob/new-commit-phase-lifecycles/text/0000-new-commit-phase-lifecycles.md#polyfill-support.

gaearon:
Can you expand on what

>This would not work in all cases (e.g. Object.defineProperty syntax).

means?

bvaughn:
Sure, sure. I meant cases like this:
```js
class Test {
  constructor() {
    Object.defineProperty(this, "componentDidUpdate", {
      configurable: true,
      enumerable: true,
      writable: true,
      value: (prevProps, prevState) => {// ...
      }
    });
  }
}
```

Which is what Babel would generate for:
```js
class Test {
  componentDidUpdate = (prevProps, prevState) => {
    // ...
  }
}
```

gaearon:
I think I'd be fine with not supporting that pattern. It shouldn't ever be necessary. As long as we document that in the polyfill README.

bvaughn:
Gotcha.

It is at least possible to warn about this case (when cDU isn't on the prototype) which this PR does.

bvaughn:
Thanks for the suggestions and discussion, everyone.

PR #34, opened 2018-03-12
Reparenting API

dantman:
[View formatted RFC](https://github.com/dantman/rfcs/blob/reparenting/text/0000-reparenting.md)

Terminology and naming for the `createReparent` function are up for discussion.

### DIFF
--- a/text/0000-reparenting.md
+++ b/text/0000-reparenting.md
@@ -1,0 +1,351 @@
+- Start Date: 2018-03-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+When writing a component that contains a set of large subtrees that stay relatively the same, but are simply moved around such that React's virtual DOM diffing can't detect the movement, React will end up recreating huge trees it should simply be moving.
+
+The goal of this RFC is to introduce a "Reparent" API that allows portions of the React tree to be marked in a way that allows React to know when to move them from one part of the tree to another instead of deleting and recreating the DOM and component instances.
+
+# Basic example
+
+## Layout
+Using reparents to render a page layout that can change structure between desktop and mobile without causing the page contents and sidebar to be recreated from scratch.
+```js
+class Layout extends PureComponent {
+    header = React.createReparent(this);
+    content = React.createReparent(this);
+    sidebar = React.createReparent(this);
+
+    render() {
+        const {isMobile, children} = this.props;
+
+        const header = this.header(
+            <div className='header'>
+                <Logo />
+                <NavigationBar />
+            </div>
+        );
+
+        const sidebar = this.sidebar(
+            <div className='sidebar'>
+                <SidebarContent />
+            </div>
+        );
+
+        const content = this.content(
+            <Fragment>
+                children
+            </Fragment>
+        );
+
+        if ( isMobile ){
+            return (
+                <div>
+                    {header}
+                    {content}
+                    {sidebar}
+                </div>
+            );
+        } else {
+            return (
+                <div>
+                    {header}
+                    <div>
+                        {content}
+                        {sidebar}
+                    </div>
+                </div>
+            );
+        }
+    }
+}
+```
+
+## Dynamic template
+Using reparents in a dynamic template/widget structure to avoid recreating widgets from scratch when they are move from one section of the template to another.
+
+```js
+const TemplateWidgetReparentContext = React.createContext({});
+
+class ReparentableTemplateWidget extends PureComponent {
+    static getDerivedStateFromProps(nextProps, prevState) {
+        if ( !prevState.widget || prevState.widgetType !== nextProps.widget ) {
+            const WidgetClass = getWidgetClassOfType(nextProps.widget);
+            return {
+                widgetType: nextProps.widget,
+                widget: new WidgetClass(),
+            };
+        }
+    }
+
+    render() {
+        const {widget} = this.state;
+
+        return (
+            <div>
+                {widget.render()}
+            </div>
+        );
+    }
+}
+
+/**
+ * This wrapper uses reparents from Template and the widget's own id
+ * to ensure that when a widget is moved from one section to another
+ * it is moved by react instead of recreated.
+ *
+ * This could be turned into an HOC
+ */
+class TemplateWidget extends Component {
+    render() {
+        const {id} = this.props;
+
+        return (
+            <TemplateWidgetReparentContext.Consumer>
+                {templateWidgetReparents => templateWidgetReparents[id](
+                    <ReparentableTemplateWidget {...this.props} />
+                )}
+            </TemplateWidgetReparentContext.Consumer>
+        );
+    }
+}
+
+class Section extends PureComponent {
+    render() {
+        const {title, widgets} = this.props;
+
+        return (
+            <div>
+                <h2>{title}</h2>
+                {widgets.map(widget => (
+                    <TemplateWidget {...widget} key={widget.id} />
+                ))}
+            </div>
+        );
+    }
+}
+
+class Template extends PureComponent {
+    static getDerivedStateFromProps(nextProps, prevState) {
+        if ( nextProps.sections === prevState.sections ) return;
+
+        let templateWidgetReparents = prevState.templateWidgetReparents;
+
+        const widgetIds = new Set();
+        nextProps.sections.forEach(section => section.widgets.forEach(widget => widgetIds.add(widget.id)));
+
+        return {
+            widgetIds: widgetIds.toArray(),
+        };
+    }
+
+    templateWidgetReparents = {};
+
+    render() {
+        const {sections} = this.props;
+        const {widgetIds} = this.state;
+
+        for ( const id of widgetIds ) {
+            if ( !(id in templateWidgetReparents) ) {
+                this.templateWidgetReparents = Object.assign(this.templateWidgetReparents, {
+                    [id]: React.createReparent(this),
+                });
+            }
+        }
+
+        for ( const id in this.templateWidgetReparents ) {
+            if ( !widgetIds.has(id) ) {
+                this.templateWidgetReparents[id].unmount();
+                this.templateWidgetReparents = omit(this.templateWidgetReparents, id);
+            }
+        }
+
+        return (
+            <TemplateWidgetReparentContext.Provider value={this.templateWidgetReparents}>
+                {sections.map(section => {
+                    <Section {...section} key={section.id} />
+                })}
+            </TemplateWidgetReparentContext.Provider>
+        );
+    }
+}
+
+let templateData = {
+    sections: [
+        {
+            id: 's-1',
+            title: 'Untitled 1',
+            widgets: [
+                {
+                    id: 'w-1',
+                    widget: 'paragraph',
+                    props: {
+                        text: 'Lorem ipsum...'
+                    }
+                }
+            ]
+        },
+        {
+            id: 's-2',
+            title: 'Lorem ipsum',
+            widgets: []
+        }
+    ]
+};
+
+ReactDOM.render(<Template {...templateData} />, app);
+
+// Modify the template to move widget:w-1 to section s-2, it should not be necessary to re-create widget w-1 from scratch
+templateData = {
+    sections: [
+        {
+            ...templateData.sections[0],
+            widgets: []
+        },
+        {
+            ...templateData.sections[1],
+            widgets: [
+                templateData.sections[0].widgets[0]
+            ]
+        }
+    ]
+};
+ReactDOM.render(<Template {...templateData} />, app);
+```
+
+# Motivation
+
+Up till now the only way to handle reparenting has been using `ReactDOM.unstable_renderIntoContainer` or Portals to separate a dom node from react rendering and allow that node to be moved around. However this is a complex hack that only works with ReactDOM, it does not work with isomorphic React web apps or other environments like React Native.
+
+This reparenting issue have been left alone for awhile. We have an RFC process now. An RFC for a new context API has made an unstable feature of React stable. And accepted RFCs like createRef have given suggestions on what APIs fitting of React might look like. I think now is a good time to try tackling reparenting.
+
+More motivation can be found in past discussions on reparenting:
+
+- [Support for reparenting (facebook/react#3965)](https://github.com/facebook/react/issues/3965)
+- [A gist that has hosted a lot of reparenting discussion](https://gist.github.com/chenglou/34b155691a6f58091953)
+
+# Detailed design
+
+This proposal introduces a `React.createReparent(component)` function. It accepts a component instance (typically `this`) and returns a *Reparent*. The "Reparent" naming is open to better name proposals.
+
+A Reparent is tied to a component, when this component is unmounted the reparent is unmounted and the detached tree is discarded. This allows reparents to hold detached trees without doing so in a permanent way that would leak memory.
+
+A Reparent has an additional `.unmount()` method, this can be used to force the reparent to unmount and discard its tree like it would when the component it is tied to unmounts. This allows reparents to be hoisted upwards and used for dynamic content and explicitly unmounted when necessary.
+
+The Reparent itself is a function that accepts children for the reparent. The return value is a special Fragment used to render the reparent and its contents somewhere in the React tree.
+
+When React intends to remove the dom/native tree rendered from this special fragment because it is not present in a later render, React instead detaches the tree and holds a reference to it in the Reparent instead of purging it. If a fragment returned by the reparent later is rendered by a component the detached dom/native tree is re-used.
+
+In addition to retaining the dom/native tree for future use React Component instances present in a Reparent are not unmounted until the Reparent is rendered without them, the Reparent is unmounted via .unmount(), or the Reparent is implicitly unmounted when its owner unmounts.
+
+Unlike a normal fragment the fragment returned by a reparent may only be used once in a react tree. If a fragment returned by a Reparent is used multiple times only the latest one will be rendered (they will not be duplicated as normal in React) and during development React may emit warnings that a reparent is used in multiple spots like the warnings when a duplicate key is found.
+
+The fragment returned has an implicit key which is unique to the Reparent. A Reparent itself is like a key that is unique beyond just a single element's children so there is no need for the user to specify an additional key to use its contents in an array.
+
+Naively, createReparent without the reparenting and unmount behaviour behaves similar to the following implementation:
+```js
+React.createReparent = function() {
+    const key = generateUniqueKey();
+
+    const Reparent = function(children) {
+        return React.createElement(Fragment, {key}, children);
+    };
+    Reparent.unmount = () => {
+        throw new Error('Not implemented');
+    };
+
+    return Reparent;
+};
+```
+
+And conforms to the following types/interface:
+```js
+type ReparentFunction = (Component) => ReactFragment;
+type ReparentObject = {
+    unmount(): void;
+};
+type Reparent = ReparentFunction & ReparentObject;
+
+interface React {
+    createReparent(): Reparent;
+}
+```
+
+# Drawbacks
+
+The `.unmount()` method is not necessarily perfect in identifying when a tree is detached but kept or discarded and to be unmounted. If someone chooses to use createReparent dynamically and forgets to call `.unmount()` they will leak memory.
+
+# Alternatives
+
+## createKey
+
+@gaearon proposed [a createKey API](https://gist.github.com/chenglou/34b155691a6f58091953#gistcomment-1460942)
+
+```js
+class MyComponent extends Component {
+    contentKey = React.createKey(this);
+
+    render() {
+        const {isB, children} = this.props;
+
+        const content = <Fragment key={this.contentKey}>{children}</Fragment>;
+
+        return (
+            <div>
+                {isB && content}
+                <div>{!isB && content}</div>
+            </div>
+        );
+    }
+}
+```
+
+The createKey API could work, however it has some limitations that createReparent does not:
+
+- While createKey and createReparent shares the same advantage that unmounting of its host results in unmounting of the reparentable root. contentKey does not have a secondary method of unmounting/discarding the tree. React cannot differentiate between a detached tree and a discarded/unmountable tree. As a result it cannot be used for varying numbers of reparentable roots as trees for discarded keys remain in memory as leaks.
+- createKey adds alternative behaviour using just the `key`, createReparent gives React internals more control in how it they decide to handle the link between the tree and the reparent.
+
+## DetatchedTree
+
+If `.unmount()` proves to be too complex it may be possible to make detached trees work with createReparent, createKey, or other global key methods by using a Fragment-like `<DetatchedTree />` component which holds a reference to the detached tree in the React tree but omits it from the DOM. Then React knows it may unmount components and trees if it is not used as part of the live tree or in a DetatchedTree.
+
+```js
+class MyComponent extends Component {
+    contentKey = React.createKey();
+
+    render() {
+        const {show, children} = this.props;
+
+        const content = <Fragment key={this.contentKey}>{children}</Fragment>;
+
+        return (
+            <div>
+                <DetatchedTree>
+                    {!show && content}
+                </DetatchedTree>
+                {show && content}
+            </div>
+        );
+    }
+}
+```
+
+# Adoption strategy
+
+Reparents can be released in a feature release of React, there are no breaking changes.
+
+If people are still using some experimental 3rd party libraries like [react-teleporter](https://github.com/jaredly/react-teleporter) that use `ReactDOM.unstable_renderIntoContainer` or Portals to hack reparenting into React DOM, we may wish to provide migration guides to the Reparent API or update these libraries to make use of the Reparent API.
+
+# How we teach this
+
+"Reparents" may not be the best terminology, some terminology including "root" in the name to refer to the segment of the tree created by a reparent that can be moved.
+
+We may need a new documentation page to explain reparenting as one was added for portals. Like portals this is an advanced feature that does not need to be part of the main tutorials.
+
+# Unresolved questions
+
+- How do reparent instances interact with hydration?
+  - Perhaps when hydrating the dom tree, portions of the dom tree that match up with portions of the virtual dom belonging to a reparent will be given to the Reparent to hydrate as its dom tree.
+- Should the Reparent function just accept a single children argument, or should it accept ...children rest and pass it on to createElement to behave similar to createElement.


streamich:
Does this RFC suggest that React would move around the actual DOM nodes? Is it actually possible? For example, if one of the child node was a playing `<video>` element, by removing it from DOM and reinserting into a different parent, would it still be playing?

dantman:
@streamich Yes, DOM nodes would be detached with removeChild if the reparent tree is not rendered, and re-inserted later with appendChild/insertChild when added back.

If one of the children of the reparent is a video element this does imply that the video element inserted in the other parent would be the exact same video element.

I cannot make guarantees however about the video still playing. Even if you keep the same instance of the dom node, some DOM elements have quirks when you move them around the dom tree. I'm not proposing any special hacks to handle that. So the video will be the same, but if videos stop playing when you appendChild them into another parent then videos will stop playing when you reparent them.

Though on the topic of playing videos. Take a look at my other ["registered prop as ref"](https://github.com/reactjs/rfcs/pull/28) RFC, one of my examples is a custom [`Playing`](https://github.com/dantman/rfcs/blob/e69115dba77e96bd6838ff81db45c8ef46edd825/text/0000-registered-prop-as-ref.md#video-playback) implementation. Which would make the video element a controlled component in regards to playing (like how inputs are controlled when you pass a `value` to them). Theoretically if you reparented a video with a controlled `Playing` attribute then even if the browser stopped it the controlled nature would make it resume after it is reparented (though you might have to ignore a pause event if the browser fires one).

streamich:
> ...if you reparented a video with a controlled `Playing` attribute then even if the browser stopped it the controlled nature would make it resume after it is reparented...

I think it resets the `currentTime`, too.

dantman:
> I think it would reset the currentTime of the video, too.

Well. If you really wanted. You could build a `<Video>` component which would make all the state in a `<video>` tag controlled (`playing`, `currentTime`, etc...) in the normal way for react (register events to keep `state` up to date like firing setState every time the video time change, map things to props and change them when props change, etc...). Then this RFC would guarantee that if your `<Video>` was reparented the React component would stay intact with all its state and in `componentDidUpdate` you could ensure that any controlled state that the browser reset from the reparent would be restored by your Video component.

polkovnikov-ph:
@dantman The whole point of reparenting is to keep intact all and any of the states of element that are not controlled by React.

polkovnikov-ph:
@streamich Yes, it's as simple as `appendChild` on a component that is already appended somewhere else, without any `removeChild` calls.

streamich:
I guess, for React just re-parenting and leaving it up to the user to deal with consequence is fine as long it is opt-in. Here is a short list of consequences:

1. `<video>` will stop and lose its state
2. `<audio>` will stop and lose its state
3. `<iframe>` will be reloaded
4. Any focused elements will lose their focus?
5. Very important one: any `.swf` Flash players will reload 🤣
6. Pobably a bunch of other things will lose their state or reload.

vcarl:
Seems to have some overlap with #7? `dangerouslyRecycleNode`sounds like a similar concept to me, though I admit I haven't deeply read through either proposal.

text/0000-reparenting.md
```diff
@@ -0,0 +1,351 @@
...
+- [Support for reparenting (facebook/react#3965)](https://github.com/facebook/react/issues/3965)
+- [A gist that has hosted a lot of reparenting discussion](https://gist.github.com/chenglou/34b155691a6f58091953)
+
+# Detailed design
+
+This proposal introduces a `React.createReparent(component)` function. It accepts a component instance (typically `this`) and returns a *Reparent*. The "Reparent" naming is open to better name proposals.
+
+A Reparent is tied to a component, when this component is unmounted the reparent is unmounted and the detached tree is discarded. This allows reparents to hold detached trees without doing so in a permanent way that would leak memory.
+
+A Reparent has an additional `.unmount()` method, this can be used to force the reparent to unmount and discard its tree like it would when the component it is tied to unmounts. This allows reparents to be hoisted upwards and used for dynamic content and explicitly unmounted when necessary.
```
mAAdhaTTah: I think this conflicts w/ the previous paragraph. If a reparent can be hoisted upwards, and the component it was tied to is unmounted, then it gets unmounted and discarded. Is there a method for changing the owner of a reparent?

dantman: Sorry for not being clear. By hoisting for dynamic content, I mean you can move the `React.createReparent(this)` higher in the tree (hoisting), this higher node you use `this` from is the "component it is tied to"/owner, you can pass the reparent down to children using context/props and use it in descendants, then if the dynamic reparent is no longer needed you can call `.unmount()` (usually at the higher node) in order to unmount and discard whatever you put inside the reparent tree.

Perhaps I should define a glossary of things like 'state tree', 'dom/native tree', 'reparent owner', etc... in the RFC and use those terms to make things clear.

dantman:
@polkovnikov-ph
There are dom states that do not get discarded when you call `removeChild` on them. And the ability to retain React Components while reparenting is just as important.

But my intent with mentioning `removeChild` is for the situations where you intentionally detach a tree and hold on to a reference to later mount it.

```js
  content = React.createReparent(this);

  render() {
    const {show, children} = this.props;
    const content = this.content(<Fragment>{children}</Fragment>);

    if ( show ) {
      // Show the contents
      return content;
    } else {
      // Don't render the contents, the dom/component tree will not be
      // present in the page but it will be retained as a detached tree
      // and later re-used when `show` becomes true again.
      return null;
    }
  }
}
```

If you move a dom node from one parent to another in a render such that the reparent is never left out of the react component tree, I do think it would be ideal for React to just use `appendChild`/`insertChild` to move the node instead of using `removeChild` and then inserting it in the new parent.

If there are dom states that get discarded when you `removeChild`->`appendChild` in the same tick but not when you just `appendChild`. I can add a note to the design/implementation section noting that the implementation should not use `removeChild` unless the tree is actually being detached and should only use `appendChild`/`insertChild` calls when it is only necessary to move dom nodes.

I can also add a limitations section noting the dom nodes that lose state when you reparent them with appendChild.

Would that be sufficient?

text/0000-reparenting.md
mAAdhaTTah: Ah, so not at runtime, but during development. Got it.

dantman: Yeah, sorry. I thought it was clear by using "hoisting" because I normally when I hear "hoisting state" it's referring to moving the `state = {}` and `setState` definition from one Component to a parent component.

polkovnikov-ph:
I'm unsure, but it seems that it's possible to create a loop of reparenting components that locally looks as if it's possible to use only `appendChild` on them, but it actually requires at least one `removeChild` to break the loop. I'm too tired today to convert this idea of counterexample into code tho :/

dantman:
To dive more into this issue of removeChild/appendChild and state loss I've built a series of simple dom reparent tests that will reparent an element using just appendChild, removeChild + appendChild, and using a delay before re-insertion. Feel free to propose other elements

https://dantman.github.io/dom-reparent-tests/

## Summary

- Chrome is the only browser that pauses video/audio when you reparent something with appendChild
- Firefox and Safari won't even pause the video when you removeChild+appendChild, until you put a delay between the calls, they'll just keep playing the video at the new position
- IE11 was the only browser that had any difference between appendChild and removeChild+appendChild – pausing the video when you removeChild+appendChild. Though given that Chrome always pauses, this is actually better than Chrome!
- Absolutely none of the browsers I tested lost the current position of the video, even when the video was detached from the tree and a delay was used before putting it back. The suggestion that `currentTime` gets reset seems to be invalid.
- iframes do reload in all browsers, no matter what. This is just a limitation that will have to be accepted for iframes as even just using appendChild without removeChild makes no difference.
- Likewise scroll positions are the same, they always reset and you cannot avoid it by just using appendChild. Please see RFC #33, it suggests a new API that solves the scroll issue for normal content updates and potentially could also be used to retain scroll state when reparenting. It may even work for retaining other states like focus and selection.

## Conclusion

It would be nice to still note that React should try to use appendChild without first using removeChild if it does not need to. However this actually isn't a big deal. In practice it appears that almost nothing is lost during removeChild that is not lost during appendChild (as long as you re-insert within the same tick). The only difference I could find is that IE11 will pause the video when you removeChild+appendChild but not appendChild. But this is a pretty pointless difference because Chrome pauses the video on even a basic appendChild, so you are going to have to make the playing state a controlled behaviour anyways.

## Results

| Chrome | append                            | remove+append                     | delay                             |
|--------|-----------------------------------|-----------------------------------|-----------------------------------|
| video  | pauses, but doesn't lose position | pauses, but doesn't lose position | pauses, but doesn't lose position |
| audio  | pauses, but doesn't lose position | pauses, but doesn't lose position | pauses, but doesn't lose position |
| iframe | refreshes                         | refreshes                         | refreshes                         |
| scroll | loses scroll position             | loses scroll position             | loses scroll position             |

| Firefox | append                          | remove+append                   | delay                             |
|---------|---------------------------------|---------------------------------|-----------------------------------|
| video   | does not pause or lose position | does not pause or lose position | pauses, but doesn't lose position |
| audio   | does not pause or lose position | does not pause or lose position | pauses, but doesn't lose position |
| iframe  | refreshes                       | refreshes                       | refreshes                         |
| scroll  | loses scroll position           | loses scroll position           | loses scroll position             |

| Safari | append                          | remove+append                   | delay                             |
|--------|---------------------------------|---------------------------------|-----------------------------------|
| video  | does not pause or lose position | does not pause or lose position | pauses, but doesn't lose position |
| audio  | does not pause or lose position | does not pause or lose position | pauses, but doesn't lose position |
| iframe | refreshes                       | refreshes                       | refreshes                         |
| scroll | loses scroll position           | loses scroll position           | loses scroll position             |

| IE11   | append                          | remove+append                     | delay                             |
|--------|---------------------------------|-----------------------------------|-----------------------------------|
| video  | does not pause or lose position | pauses, but doesn't lose position | pauses, but doesn't lose position |
| audio  | could not test with .wav        | could not test with .wav          |  could not test with .wav         |
| iframe | refreshes                       | refreshes                         | refreshes                         |
| scroll | loses scroll position           | loses scroll position             | loses scroll position             |

dantman:
Important discussion points:

Is it possible to hydrate a server-rendered DOM tree and know, without needing deterministic IDs, and know what portion of it belongs to a Reparent and should be detached and held on to when unmounting instead of destroyed? (Or would we need to wait for the internals of https://github.com/reactjs/rfcs/pull/32 to solve this)

Should the reparent function (`this.header(children)`) accept multiple children with a spread, the same way `createElement` accepts multiple children? These would simply be passed on to `createElement`, so the Fragment could contain more than just one child. Or is this API confusing and we should just recommend people needing this pass in a Fragment as the child of the Reparent?

When a Reparent's tree is detached from the DOM (the DOM nodes are not part of the document, but the tree still exists so it can be re-inserted later) and the Reparent is given new children. Should it render those children into the detached tree; or, should it just store the new children and wait until the tree is re-mounted before rendering them.

I can see rationales for both:

- If we render to a detached tree it should mostly work out, but some dom operations that may be done in refs may not like it and expect there to be a parent. Perhaps something like `.closest` that tries to crawl up to the root or dom operations that calculate layout.
- If we don't render to a detached tree then React components inside that tree will not be updated. If this includes a props change that affects something like Redux/Flux subscriptions the Component may continue to work with out of date subscriptions, which could result in runtime errors as the Component tries to access data that doesn't exist anymore or is under a different name now.

dantman:
I've added a possible alternative to `.unmount()`, a [`.keep()`](https://github.com/dantman/rfcs/blob/reparenting/text/0000-reparenting.md#keep) signal.

I'd like some feedback on that idea. If people like it I may refactor the RFC to use `.keep()` and make `.unmount()` the alternative.

I've also clarified the goals and explained why reparenting is useful even if we can't retain native state like focus and iframe contents and even when on React Native where there's not necessarily any concept of reparenting at the native View level.

ackvf:
I like the idea of moving nodes around, but this seems as a needless overhead for the developer. Why not introduce new react-specific attribute, such as is the `key` that will tell React that this is the same node?

dantman:
> I like the idea of moving nodes around, but this seems as a needless overhead for the developer. Why not introduce new react-specific attribute, similar to the key that will tell React that this is the same node?

This is mostly explained in the Requirements section. Here are a few points:

- No other React API has the concept of a globally unique string.
- Using globally unique keys makes using reparents harder in libraries since you have to ensure that your keys will not conflict with any the user makes.
- Using globally unique keys makes it harder to use things like database IDs for the unique key. You can guarantee that this ID is unique within the set of data you want to be able to move around, but you cannot guarantee that the ID is 100% unique among all possible reparentable nodes. Thus you have to pad it with uniqueness that shouldn't be necessary for the type of reparenting you are doing.
- If we use keys it is not possible to hold on to detached nodes. There are legitimate reasons to want to be able to intentionally detach a tree temporarily. And there are legitimate cases where a tree momentarily may not be present in the tree during a move but you do not want it to be unmounted (and this type of async move may become more common after the future async rendering mode is introduced).

hiddentao:
I'm currently building a web browser-like app in react native. One of the requirements is to be able to switch from viewing a single tab to viewing all tabs in an animated scroller - at the moment, due to the inability re-parent, switching between views causes tabs to be re-mounted from scratch, loosing any and all state. So re-parenting would be useful for me.

erobwen:
I would love this feature as well, when doing a responsive app. I cannot understand why they did not have this from the very beginning, it is brilliant!

philipp-spiess:
Great work on the RFC @dantman. 🙌 That'd definitely be useful for us at @PSPDFKit as well!

The `.keep` API makes sense as it must be possible to retain elements even if they aren’t rendered it in the current tree. However it seems a bit counter-intuitive that the reperantable element will also be kept in the following example: 

```js
class DetachableTree extends Component {
  reparent = React.createReparent(this);

  render() {
    const content = this.reparent(this.props.children);
    return this.props.show && content;
  }
}
```

I think it's also against the design principles of React that the `render` method has side effects. In this case, the render method might return `null` but we still have to track that we want to keep the reperantable - thus we would need to save that information as a side effect. Maybe this can be improved by rendering a special component that describes this keeping state?

Another problem is that the lack of a context makes it possible to create a reparent deep in the tree and pass it upwards to render it in the root. That's something that's not possible with the context API for example and I'm not sure if this causes any implications.

I also fear that the id passing might be very confusing for developers that use it. Your examples make use of `getDerivedStateFromProps` twice which smells because that method should rarely be used at all. I personally prefer your `createReparentContext` to solve this. In this API, you create a context/repereantable pair similar to the new context API. This will drastically reduce the chance of an id clash since _you_ are in total control of the context. Third party APIs can never clash with your repereantable contexts as well.

Here’s one idea that I have in mind that tries to cope with the problems I outlined above. I decided to migrate your TemplateWidget example to show you how you can get rid of gDSFP:

<details>
  <summary>TemplateWidget using createReparentContext</summary>

  ```js
  const WidgetReparentContext = React.createReparentContext();

  function TemplateWidget({ id }) {
    return (
      <WidgetReparentContext.Reparentable id={id}>
        <p>Hello from widget {id}</p>
      </WidgetReparentContext.Reparentable>
    );
  }

  function Section({ title, widgets }) {
    return (
      <div>
        <h2>{title}</h2>
        {widgets.map(widget => <TemplateWidget {...widget} key={widget.id} />)}
      </div>
    );
  }

  function Template({ sections }) {
    return (
      <WidgetReparentContext.Provider>
        {sections.map(section => {
          <Section {...section} key={section.id} />;
        })}

        {keep}
      </WidgetReparentContext.Provider>
    );
  }
  ```
</details>
<br/>

Nothing special about this so far, this is basically your `createReparentContext` idea. One thing to notice here is that your solution does also keep widgets mounted that were once passed in as props but are no longer. I think that this use case can be further generalized into custom cache strategies which must know which items are added in order to know how to cache them.

To solve this, I assume that we could add a callback to the `ReparentContext.Provider` which gets called whenever a reparentable key is rendered. A regular React component could then collect all of those and render special "keep" placeholders so that they are always part of the render results:

<details>
  <summary>TemplateWidget using createReparentContext with onReparentableRendered</summary>

  ```js
  const WidgetReparentContext = React.createReparentContext();

  function TemplateWidget({ id }) {
    return (
      <WidgetReparentContext.Reparentable id={id}>
        <p>Hello from widget {id}</p>
      </WidgetReparentContext.Reparentable>
    );
  }

  function Section({ title, widgets }) {
    return (
      <div>
        <h2>{title}</h2>
        {widgets.map(widget => <TemplateWidget {...widget} key={widget.id} />)}
      </div>
    );
  }

  class Template extends React.Component {
    state = { keptReparentables: [] };

    // In this example we use the `onReparentableRendered` callback to be notified
    // whenever a new reparentable is rendered. This will easily allow more 
    // advanced caching systems (e.g timeout- or LRU-based) to be implemented 
    // as React components.
    addReparentable = ({ id }) => {
      this.setState(state => {
        // We already cache this reparentable so there's nothing to do.
        if (state.keptReparentables.includes(id)) {
          return null;
        }
        // We add this to our kept reparentable list.
        return { keptReparentables: [...state.keptReparentables, id] };
      });
    };

    render() {
      const { sections } = this.props;
      const { keptReparentables } = this.state;

      return (
        <WidgetReparentContext.Provider
          onReparentableRendered={this.addReparentable}
        >
          {sections.map(section => <Section {...section} key={section.id} />)}

          {/* A special kept flag will tell React to keep this reparentable around
              if it exists without updating properties. */}
          {keptReparentables.map(id => (
            <WidgetReparentContext.Reparentable id={id} keep />
          ))}
        </WidgetReparentContext.Provider>
      );
    }
  }
  ```
</details>

<br>

I'd love to hear your thoughts on this API 😊 I also think that this API can be implemented in userland if we have https://github.com/facebook/react/issues/13044 (well, for ReactDOM only at least as long as `createPortal` is DOM-only).

dantman:
> I think it's also against the design principles of React that the render method has side effects. In this case, the render method might return null but we still have to track that we want to keep the reperantable - thus we would need to save that information as a side effect. Maybe this can be improved by rendering a special component that describes this keeping state?

This is correct, we do have to accept that in this case not only is there an effect from a return from render but also a `.keep()`/`Reparent()` in render. I am half on/half off with this idea though, while technically true it being a side effect, in a way we are just tracking references in render() and GCing things when references disappear which doesn't feel as strange as other types of side effect. I did consider requiring that `this` be passed to `.keep()` if it turned out this behaviour couldn't be implemented in React.

Fragments do make it possible to, use some sort of internal reference tracking element where we pass the reparent to it to keep the reference without it being rendered, without it messing with the dom. But I'm very hesitant to use that pattern given how horrible it looks to use that type of pattern. How it limits how render() can be written, in some cases forcing you to rewrite what could be a conditional `if (...) { return ...; }` into an `if (...) { result = ...; } return <>{/* code to keep references to reparents */}{result}<>;`. And I also feel a little awkward about how `.keep()` is designed to allow you to keep a reference to something while still rendering it somewhere else, something that feels at odds with rendering a special component because you are returning a reparent in 2 places in JSX, it just happens that one of those places is actually a special type of component that doesn't actually render the reparent using a loophole to bypass the "a reparent may only be rendered in one place" rule.

> Another problem is that the lack of a context makes it possible to create a reparent deep in the tree and pass it upwards to render it in the root. That's something that's not possible with the context API for example and I'm not sure if this causes any implications.

I suppose someone ingenious *might* be able to find a way to pass a reparent upwards and find a way to use a few tricks to make a reparent render the component that first created the reparent as as child. Although I don't believe this is possible without the use of a second reparent that technically means the component wouldn't actually be a child of the reparent created in the component.

I'm also not sure this is even a problem, as with the new .keep() API I believe a Reparent is now pretty much disconnected from the component that creates it. We only opt-in to patterns like that because it's useful to have a component keep references to a series of reparents so they don't unmount and pass those reparents to its children.

Also worth noting, it's not directly within the React tree but the Portal API does allow you to render something in the root from a child component of the root.

> I also fear that the id passing might be very confusing for developers that use it. Your examples make use of getDerivedStateFromProps twice which smells because that method should rarely be used at all. I personally prefer your createReparentContext to solve this. In this API, you create a context/repereantable pair similar to the new context API. This will drastically reduce the chance of an id clash since you are in total control of the context. Third party APIs can never clash with your repereantable contexts as well.

The IDs are purely an implementation detail of the example, they aren't required of the API. If someone wants they are free to use a Map and object keys in theirs. Id clashes are also a non-issue, the IDs only matter within the scope of the context created by the user in the example. i.e. You can create another context containing reparents for something else and use the same ids, there will be no clash because they are separate contexts – the IDs do not need to be globally unique.

Honestly the only reason that example is ID based is because the application I am currently working on happens to use IDs. I needed an example of that kind of pattern of passing a dynamic set of reparents to descendents via context, so I just used what I had on hand. That example is basically just a pattern I'm actually using in the application I'm currently working on right now, just stripped down to the most basic/generic parts so I'm not sharing any client code.

Also I think the examples originally needed `getDerivedStateFromProps` a bit more in the pre-keep version of the rfc. Now that we have keep, I have a feeling that you could probably refactor that example to work using very very minor memoization in render(). Also that example is an example of one of the most extreme most advanced ways you can use a reparent, the example kind of is the "rare" situation where you do want gDSFP.

Speaking of userland things, I am aiming for an api that is as flexible and supporting of advanced cases as possible. As I expect that a simple API will not cover all the use cases people want. And if we implement an API that is advanced/flexible enough people will be free to write libraries that expose simple APIs for using reparents. This will let different use cases for reparents be supported by separate libraries that each have an API that is simplest to use for the use case they target. And will allow anyone with a use case too advanced for a library to handle to use the built in API.

dantman:
> However I'm still not keen on the idea that we're sacrificing the purity of the render method because we have to track keep() calls. I agree that your API looks more compact but would it make it clear that, for example, you'd have to call setState() and force a re-render to be able to "unkeep" reparentables? I think a React component would make that more obvious because people using React are already familiar with that concept.

Is this really something that strange?
- You are explicitly calling a method that says it will keep a reference to the reparent. And in some cases you are explicitly calling a `.keep()` method, whose sole purpose is to keep a reference and only reason you'd call it is if you knew what it did.
- I'm also not sure forcing a re-render with setState is that strange either. If you keep/render children for a reparent in a render you've essentially "used" the reparent in a render(), it doesn't seem that strange that something you used in a render() won't go away until that component is rendered again.

> That can easily be solved by returning three elements from the createReparentContext() API and not calling the special loophole reparentable at all:

Ok that covers the strangeness with a different API. Though I don't really like this id based pattern because there are also use cases for singular reparents.

> Another thing that is impossible with your API (which I just found out now that I sketched the above example): It would require the use of a class component for the this to be passed into createReparent().

Oh, whoops. I didn't realize that my examples and my protocol still listed `this`/`Component` in them. Ignore those, the rewrite to use `.keep()` was so big I missed things. `this` was required in the original `.unmount()` based API because we needed an "owner" so we could implicitly unmount when the owner unmounted. With the `.keep()` we do not need to know the owner because we only have to keep track of what components are using the reparent in order to know when to umount its contents.

Every time an example uses `React.createReparent(this)` it should actually be `React.createReparent()` now.

> Another example where the .keep() API is insufficient is when you think about unit testing using techniques like the shallow renderer. How would you write an assertion that verifies that .keep() is called indeed?

It's possible you may need a bit more than a shallow render, I'm not sure how shallow the shallow renderer is. But you can test the effects of keep by watching whether a component instance unmounts or not. I think this is justified because you are not testing the react tree but testing the effects of a mounting behaviour.

Something along the lines of:
```js
let mounted = false;
class Foo extends Component {
	componentDidMount() {
		mounted = true;
	}

	componentWillUnmount() {
		mounted = false;
	}
}

const reparent = React.createReparent();
function Bar({show, children}) {
	reparent.keep();
	return show ? reparent(children) : null;
}

renderer.render(<Bar show><Foo /></Bar>);
// wait
expect(mounted).toBe(true, 'Foo should be mounted when rendered through a reparent');

renderer.render(<Bar hide><Foo /></Bar>);
// wait
expect(mounted).toBe(true, 'Foo should not unmount even when not rendered, becase .keep() was called');

renderer.render(<Bar show></Bar>);
// wait
expect(mounted).toBe(false, "Foo should unmount when it is no longer part of the reparent's children");
```

aapoalas:
+1 to not particularly liking the id based approach. I find it confusing.

If the side-effect keep-function is seen as a problem, would making keep into a React Component be an acceptable alternative? I think this has been partially in discussion between you already, but I wasn't exactly sure if it was the same as I was thinking about.

So essentially something like a React.Fragment, but instead of allowing a grouped-together set it would act as the keep-call for one or more Reparents inside it. This admittedly adds complexity to the already relatively complex API but would remove the side-effects and would make the API a reasonably accurate rendition of the actual resulting DOM: The main DOM tree expands as the React tree does, while the Reparent components inside the Keep component are still in existence in the DOM, just in a detached state. The component could even be named with something to that effect, like Detached or I don't know.

Sorry if this is just rehashing the same ideas already given and shot down.

philipp-spiess:
@aapoalas No need to feel stupid, thank you for helping us out here 🙌 

So just to clarify, you have something like this in mind, right?

```js
const reparent = React.createReparent();

function SomeComponent() {
  return (
    <React.Fragment>
	  {reparent(<p>Some DOM</p>)}

      <React.Detached>
        {reparent()}
      </React.Detached>
    </React.Fragment>
  );
}
```

I think one problem here is the same that @dantman pointed out earlier: It's hard to see why we're not allowed to render the same `reparent()` twice inside the React tree but it's ok when one of them is wrapped inside a `<React.Detached>`. I'm quoting here because I found his words clearer:

> And I also feel a little awkward about how .keep() is designed to allow you to keep a reference to something while still rendering it somewhere else, something that feels at odds with rendering a special component because you are returning a reparent in 2 places in JSX, it just happens that one of those places is actually a special type of component that doesn't actually render the reparent using a loophole to bypass the "a reparent may only be rendered in one place" rule.

What I like about this idea though is that the return value of `createReparent()` can now easily be a component itself which makes it a mot more intuitive to use:

```js
const Reparent = React.createReparent();

function SomeComponent() {
  return (
    <React.Fragment>
	  <Reparent><p>Some DOM</p><Reparent />

      <React.Detached>
        <Reparent />
      </React.Detached>
    </React.Fragment>
  );
}
```

dantman:
> What I like about this idea though is that the return value of createReparent() can now easily be a component itself which makes it a mot more intuitive to use:

That's an interesting idea, one caveat though. How do we differentiate between rendering `<Reparent />` as a "keep" and rendering `<Reparent></Reparent>` with empty children? We do need to allow a reparent to be rendered with empty children. It would be really confusing if the answer to that was "`<Reparent />` behaves differently when it is a child of `<React.Detached />`.

I still don't quite like the idea of an element based API for reparents. But this could work for one:
```js
const Reparent = React.createReparent();

function SomeComponent() {
  return (
    <React.Fragment>
      <Reparent.Reference />

      <Reparent><p>Some DOM</p></Reparent>
    </React.Fragment>
  );
}
```

We can use a different name than `.Reference` if someone can come up with one. But this would also make `<React.Detached />` unnecessary. Rendering the Reparent itself would render it and its contents. But rendering the `<Reparent.Reference />` would just be a reference to the reparent and wouldn't change it's contents.

Though I still do not like the idea that a complex branching render() would have to be refactored so it can return elements that aren't actually elements and aren't actually rendered in order to keep references to something.

j-f1:
What if you could pass a component to `createReparent`:

```js
const reparent = React.createReparent(() => <p>Some DOM</p>)

function SomeComponent() {
  return (
    <React.Fragment>
      <Reparent />
    </React.Fragment>
  )
}
```

philipp-spiess:
@j-f1 Can you elaborate what the benefits of this approach is over using the regular React tree in the `render()` function as proposed? I do think it makes it harder for example to swap the contents.

j-f1:
@philipp-spiess What I was thinking is that if you stopped rendering the `<Reparent />`, it would cache the rendered node somewhere to reuse when you rendered it again.

aapoalas:
@philipp-spiess I was thinking something along the following lines (though thinking deeply on this I keep finding more and more problems so this isn't going to be a whole lot of good I think):

I would have `createReparent` return a unique React Element that acts similarly to a React Fragment; inside it you can render any Components and the contents is diffed normally for changes. Each time you render it, the contents must be rendered as well, no keeping of references that remember what the previous contents was and would nicely render that stuff for you or some such.

The React.Detached Component would again be similar to React Fragment; inside it you can render any Components and the contents is diffed normally for changes. Yes, I mean you could also render normal divs etc. inside this component. It would act the same as creating a DOM tree manually without appending it to be a child of the body. This would, however, be mostly useless without the Reparent Component, as the moment you "move" your normal DOM elements from within the React.Detached to some actual DOM tree, the Components of course unmount and are destroyed, and new ones created in the new location. The Reparent Component could however use the React.Detched to keep itself alive while detached without ill effects.

So the usage of these would be something like:
```javascript
const ReparentA = React.createReparent(); // This would more likely be created inside some parent component during mounting or updating based on some inflowing data

class MyPointOfForking extends React.Component {
  render() {
    const {
      children,
      condition,
    } = this.props;
    return (
      <ForkA>
        {condition === true && <ReparentA>{children}</ReparentA>}
      </ForkA>
      <ForkB content={condition === false && <ReparentA>{children}</ReparentA>} />
      <React.Detached>
        {typeof condition !== "boolean" && <ReparentA>{children}</ReparentA>}
      <React.Detached>
    );
```

So we have some React tree (children in this case, but could just as well be created inside the render function) that we want to render somewhere, but the location changes based on some condition, or it can be detached from the DOM entirely in some cases.

Thus we have two choices by and large: Either we place the ReparentA Component (remember, this is a unique individual) in a normal React DOM tree one way or another (ForkA and ForkB, where ForkA will render directly as children, while ForkB might likely pass the content prop onwards to somewhere even deeper, before it finally gets rendered as a child), OR we can render the component in a detached React DOM tree.

Doing more than two of these will cause an error. You cannot render a Reparent in two places, not even if one of those is a Detached DOM tree.

I can see some problems with this approach though:
- Keeping the Reparents straight (which one should render what content etc) can easily become a chore and might well end up detaching and remounting your contents even thought the whole point is to avoid that.
- Having a unique Component is heretofore unknown in React and will indubitably cause confusion.
- The need to keep the contents / children of each Reparent in order will likely lead to passing of the Reparent Components as props or children through React Component trees, and much deeper than is done currently. This is not necessarily a good thing, though both the Context API and Portals can be of help here.

Some good things:
- The lifecycle of a Reparent component is essentially identical to that of a normal component; if at some point the component fails to be returned from some render function, it will unmount.
- Although it may seem so, this approach does not mind rendering the pretty much anywhere; even though the Reparents are likely created within some parent component, f.ex. based on an array of ids, the parent component does not tie its own lifecycle methods to the Reparents. This should work fine as long as you do not attempt to render a Reparent as a parent of the component that originally created the Reparent, as that would lead to a constant creation of a new Reparent that would then become its own parent, triggering a new creation...
- A Reparent that can be rendered at present DOM node or in some deep child can quite easily be made using the Reparent and Context API. Use Context to pass on either null or the Reparent component, while rendering the other at present node.
- A Reparent that can be rendered at present DOM node or in eg. a modal container can quite easily be made using the Reparent and Portal API. Render the Reparent component either normally, or into your modal container using a Portal.
- A Reparent that can be detached from the DOM at will is as easy as the above Portal case. Just use React.Detached instead of React.Portal (which React.Detached is just a special case of, essentially, and could easily be created in userland)

dantman:
> Doing more than two of these will cause an error. You cannot render a Reparent in two places, not even if one of those is a Detached DOM tree.

The ability to "keep" a reference while still rendering it elsewhere is a requirement. This API won't work well when you have to wait for something async to finish when moving a reparent from one child to another.

aapoalas:
I don't quite understand what kind of reference do you want to keep, and what would you want to do with it.

Also, what do you mean by "wait[ing] for something async to finish when moving a reparent from one child to another"? Do you mean that part of the content of a reparent will move first, and some other part will be moved later once some async action has finished?

dantman:
By keep a reference I mean it should be possible for a component to use a reparent such that it will not unmount, without providing children, and while still allowing another component to actually render the reparent.

By waiting for async, I'm referring to a scenario where a parent with dynamic contents passes reparents to children (likely using context), the children render the reparents, then the dynamic content changes. If the dynamic content changes such that a reparent immediately moves from one child to another child then the reparent would normally move fine. However, if the second child happens to need to wait for say some API call to fetch before it can render then the reparent will not have a destination until that is fulfilled. If the parent cannot essentially say "this reparent is still in use, even if no-one is currently rendering it" then while we are waiting for the API call to finish the reparent will end up unmounting instead of detaching so it can be inserted into the second child when it renders.

aapoalas:
Reference: What does the first component "use" the reparent for if it does not provide content for the reparent, and does not render the reparent?

Waiting for async: In my opinion this is simply a question of keeping proper tabs on the reparent components, just as you would keep proper tabs on components you create in normal renderers. eg. The second child needs to wait for some API call before it is ready to actually render the reparent. Then in its renderer it should return `<React.Detached>{reparent}</React.Detached>` until the API call is done, after which it will render some other stuff along with the reparent. Or alternatively the parent should keep tabs (via Redux store, or just callback functions, or w/e) on which reparent components are actually in use, and itself return the rest inside a React.Detached Component.

If it really is thought necessary to allow for the parent to essentially call `keep()` on reparents it has created just "to be sure they don't disappear", then I believe we need to bring back the this-binding and maybe even explicit unmounts (I've reversed my stance on .keep()-function inside render :)).

philipp-spiess:
There will still be problems with Async React. In the future, React might render and commit one part of the tree before the other part renders (see [this tweet from Dan](https://twitter.com/dan_abramov/status/1007211886541180928)). This means that there might be a "hole" when no reparent is rendered although you're immediately moving it into a different subtree.

This is why we need an API where a shared controller can tell React to not GC certain reparents - No matter if they appear in a commit phase or not.

dantman:
> Reference: What does the first component "use" the reparent for if it does not provide content for the reparent, and does not render the reparent?

The parent knows that the item the reparent is associated with still exists. It does not know if it will be rendered.



> Waiting for async: In my opinion this is simply a question of keeping proper tabs on the reparent components, just as you would keep proper tabs on components you create in normal renderers. eg. The second child needs to wait for some API call before it is ready to actually render the reparent. Then in its renderer it should return <React.Detached>{reparent}</React.Detached> until the API call is done, after which it will render some other stuff along with the reparent. Or alternatively the parent should keep tabs (via Redux store, or just callback functions, or w/e) on which reparent components are actually in use, and itself return the rest inside a React.Detached Component.

- Assuming the parent knows enough about the async request, this just adds needless complexity trying to track this when there is no need to track it
- The child returning a detached container works if the child is the one doing the network request, but if the async request comes from an intermediate component "Parent > AsyncElement > Child" neither of them know when they should render a detached node in order to just avoid the component being unmounted.

> I've reversed my stance on .keep()-function inside render :)

How about the [alternative JSX based api](https://github.com/reactjs/rfcs/pull/34#issuecomment-397843490)? That uses a `<Reparent.Reference />`.

aapoalas:
> There will still be problems with Async React. In the future, React might render and commit one part of the tree before the other part renders (see this tweet from Dan). This means that there might be a "hole" when no reparent is rendered although you're immediately moving it into a different subtree.

I do not know much anything about Async React, having already forgotten all that was said in the presentation. I believe this is not an abyss we should look too deeply into. Dan mentions avoidance of "tearing" and if reparents were a current part of React, I'm sure reparents losing state / unmounting would be considered "tearing". He also makes a vague reference to the postponing of commission being an opt-in thing, so such a situation where a reparent gets unmounted due to postponed rendering could also be argued to be a user error.

> The child returning a detached container works if the child is the one doing the network request, but if the async request comes from an intermediate component "Parent > AsyncElement > Child" neither of them know when they should render a detached node in order to just avoid the component being unmounted.

Mmm... I'd want to say that the AsyncElement here should take the reparent from the context and either render it in React.Detached, or pass it as a prop to Child once it is done with the necessary request. Similar to lifting of state and all that. Of course, if AsyncElement here is some 3rd party component that you can't mess around with, you may be in quite some trouble.

> How about the alternative JSX based api? That uses a `<Reparent.Reference />`.

... It's okay, maybe. I'm not exactly thrilled with it, as I already managed to sell myself on my idea. (Surprise!) I've managed to convince myself that a <Reparent> as close to a normal element as possible is the prettiest and most elegant solution and thus the best. So having this extra Reference thing is weird as it doesn't really have any precedent and smells a bit of laziness ("I don't know if my stuff is actually needed, so I'll just keep it around just to be safe...").

The whole Reference / keep() -system feels a little like tying the `<Reparent>` into the parent's lifecycle methods without actually doing it. So in that sense if what is wanted is a type of reparent that is created and stays in existence until it is explicitly thrown away (leaving out the Reference is an explicit act), then I'd maybe rather just have `<Reparent>` automatically become detached and stop updating if it is not rendered anywhere, and bring back the explicit unmount(). This would remove the need for React.Detached and Reference and would still allow reparents to be relatively normal components.

Throwback:
> We do need to allow a reparent to be rendered with empty children.

Why? What would an empty reparent accomplish? Except work as a different kind of `null`. How I see it in my proposal, once a reparent is created it can be used, become unmounted and be reused later if that is wanted, for some reason or another. Of course all components inside it would likewise be gone, but that's a designed for and intended facet of my proposal.

Rendering the reparent as empty would, of course, work as well but it would simply not do anything nor would it be a necessary thing in order to keep the reparent usable for a future case. I can only see this being a reasonable situation when the only child (or all children) of a reparent return `null`s from their renderers, in which case the reparent is empty, DOM-wise, but not React-wise.

dantman:
> I do not know much anything about Async React, having already forgotten all that was said in the presentation. I believe this is not an abyss we should look too deeply into. Dan mentions avoidance of "tearing" and if reparents were a current part of React, I'm sure reparents losing state / unmounting would be considered "tearing". He also makes a vague reference to the postponing of commission being an opt-in thing, so such a situation where a reparent gets unmounted due to postponed rendering could also be argued to be a user error.

From what I've seen of the suspense api so far. It's encapsulated enough that forcing a parent to know enough about what it's children are rendering in order to know when it needs to render a detached node in some apps would break encapsulation. And a reparent unmounting because a placeholder was rendered temporarily is not user error.

Additionally while suspense is great, it's not the only way to do async react. Apps are currently already doing async react using componentDidMount and state. Even if suspense were to somehow solve the reparent issue internally and avoid unmounting something that may be used in the tree it is waiting for; it still wouldn't solve the issue for all the apps currently not using suspense for fetching data.

> Mmm... I'd want to say that the AsyncElement here should take the reparent from the context and either render it in React.Detached, or pass it as a prop to Child once it is done with the necessary request. Similar to lifting of state and all that. Of course, if AsyncElement here is some 3rd party component that you can't mess around with, you may be in quite some trouble.

Even if we assume AsyncElement wasn't a 3rd party component, pretend it was say a lightweight component tasked with downloading i18n data and then rendering its children once that data is available. It would be a pretty big encapsulation break to say that knowledge of Child's reparent should be hoisted into AsyncElement.

And frankly even the fix of having the Parent pass a "render this if you aren't going to render Child at the moment" down through the tree gets pretty messy as you try to add layers.

> Why? What would an empty reparent accomplish? Except work as a different kind of null. How I see it in my proposal, once a reparent is created it can be used, become unmounted and be reused later if that is wanted, for some reason or another. Of course all components inside it would likewise be gone, but that's a designed for and intended facet of my proposal.

My issue was primarily with the distinction between `<Reparent>Foo</Reparent>` being the way that one renders to a reparent. And `<Reparent></Reparent>` ending up being a reference that keeps the children and should usually be used in `<React.Detached>`.

A) React allows you to render an element with no children, and the only difference is the absence of the data in the `children` prop. It would be a pretty large break from React conventions if at a fundamental level `<Reparent>Foo</Reparent>` rendered 'Foo' and `<Reparent></Reparent>` kept rendering 'Foo' instead of removing it.
B) children can also be undefined as a result of a conditional, `<Reparent>{condition && <Bar />}</Reparent>`. You would expect this to render `<Bar />` when condition is true `<Foo condition />` and nothing when condition is undefined (remember `<Foo />` has an undefined `condition`, not `false`). Surprise! It actually renders `<Foo />` when true, and when undefined it's actually a reference which would suggest that the reparent is still rendered but its children are not modified.

> The whole Reference / keep() -system feels a little like tying the <Reparent> into the parent's lifecycle methods without actually doing it. So in that sense if what is wanted is a type of reparent that is created and stays in existence until it is explicitly thrown away (leaving out the Reference is an explicit act), then I'd maybe rather just have <Reparent> automatically become detached and stop updating if it is not rendered anywhere, and bring back the explicit unmount(). This would remove the need for React.Detached and Reference and would still allow reparents to be relatively normal components.

It actually feels like the reverse for me. I think of it similar to GCing. React doesn't unmount things as an explicit order from someone to unmount something, it unmounts things implicitly when they are no longer in use anywhere. Just like how a GC tracks what objects are being used and removes any objects that are no longer referenced.

Having to explicitly track and `.unmount()` reparents sounds like adding an imperative order, in a place where React doesn't work imperatively. While `.keep()` and `<Reparent.Reference />` are declarative "I'm still using this" declarations. And they fit in with the GC-like behaviour of unmounting by declaring when you are still using something (have a "reference" to it like the GC tracks references) and letting React unmount it when you no longer indicate it is in use.

Aside from that, people brought up some legitimate concerns about the explicit `.unmount()` API in that it's too easy to accidentally introduce a memory leak.

> ... It's okay, maybe. I'm not exactly thrilled with it, as I already managed to sell myself on my idea. (Surprise!) I've managed to convince myself that a as close to a normal element as possible is the prettiest and most elegant solution and thus the best. So having this extra Reference thing is weird as it doesn't really have any precedent and smells a bit of laziness ("I don't know if my stuff is actually needed, so I'll just keep it around just to be safe...").

I was going to agree and arguer further that the idea of a unique instance `<Reparent>` element where Reparent is a dynamically generated element with it's own context was strange and didn't behave like a normal element any more than adding `.Reference`. But then I realized I was wrong. Uniquely generated elements and use of `Foo.Bar` is **not** strange or without precedent.

```js
const FooContext = React.createContext();
// ...
<FooContext.Provider value='foo'>
  <FooContext.Consumer>
    {foo => <span>{foo}</span>}
  </FooContext.Consumer>
</FooContext.Provider>
```

React's new context system's Provider and Consumer are unique component instances where the component itself is tied to the context of the FooContext. And they also are commonly used in the `Foo.Bar` format.

The only modification I need to make to my suggested API to make it behave exactly like the context API is to make it so that you don't use Reparent directly as an element.

```js
const Reparent = React.createReparent();
// Keep a reference to the reparent, without changing its contents.
// Can be used as many times as you want.
<Reparent.Reference />
// Render something into the reparent's context
// and render the reparent into an element.
// Can only be rendered in one location at a time.
<Reparent.Scope>...</Reparent.Scope>
// Render a reparent into a detached state.
// This is distinct from just using `<Reparent.Reference />` to keep a reference
// to avoid the component unmounting and it detaching when it's no longer rendered.
// The distinction is that in this case you are explicitly updating the
// reparent's contents while it is detached.
<React.Detached>
  <Reparent.Scope>...</Reparent.Scope>
</React.Detached>
```

All three of these terms are up for better naming suggestions.

aapoalas:
> The only modification I need to make to my suggested API to make it behave exactly like the context API is to make it so that you don't use Reparent directly as an element.

Yes, this is good! Although I'd at least give thought to `<Reparent.Reference>` having to encapsulate the React tree within which `<Reparent.Scope>` will be used like how it works with Context API. This might end up with a distinctly weird, deep tree of `<Reparent.Reference>`s when an array of reparentable components are created, but would allow for control over where the Reparent might actually be returned from.

Overall, though, I think this might well be the True Form of the API 👍

Sidenote: I can't remember too well how `<Placeholder>` was meant to work and if it would break this idea, but I was imagining a reparent unmounting tied to "synchronized" render states; that is to say that if a reparent component was removed from the React tree and did not appear elsewhere, then the unmounting of said reparent would be delayed until the next complete, "unteared" render state is reached. This would cause delayed componentWillUnmount calls from reparents in Async React applications but would otherwise be semi-ok. However, this does not offer any way for non-Suspense components to handle Reparents "natively".

dantman:
> Although I'd at least give thought to <Reparent.Reference> having to encapsulate the React tree within which <Reparent.Scope> will be used like how it works with Context API.

Interesting idea. But no good, that would result in this.
```js
const ReparentA = React.createReparent();
const ReparentB = React.createReparent();
const ReparentC = React.createReparent();

<ReparentC.Reference>
	<ReparentB.Reference>
		<ReparentA.Reference>
			<Foo />
		</ReparentA.Reference>
	</ReparentB.Reference>
</ReparentC.Reference>
```

Doing this dynamically would be a mess. If your dynamic structure uses a list of dozens of children, then the DevTools tree would be dozens of elements deep for everything. And ironically, unless you make use of an additional `<HostReparent.Scope>` the entire tree outside the reparents will remount every single time you add/remove a reparent. 😱🤣

aapoalas:
> that would result in this

Yeah, I foresaw this. It would be so nice to look exactly like the Context API but it definitely is better without doing so.

philipp-spiess:
Sorry for throwing in the confusion about async React. I thought more about it and this is an implementation detail and not a point for deciding about the API.

👍 for the latest JSX API - I guess the `<React.Detached/>` component is only for demonstration purposes and not part of this RFC though, correct?

Here are some alternative name ideas for the `<Reparent.Reference />` component:

- `<Reparent.Keep />`
- `<Reparent.KeepAlive />`
- `<Reparent.Persist />`
- `<Reparent.KeepLatest />`
- `<Reparent.NoUnmount />`

And here are some ideas for `<Reparent.Scope/>`:

- `<Reparent.Target />`
- `<Reparent.Content />`
- `<Reparent.Inside />`
- `<Reparent.Reparentable />` 🙃

My favorites so far are `<Reparent.Persist />` and `<Reparent.Content />` but I urge you to just take this as additional input with no strings attached as I'm terrible at finding good names and always go for the "make it easy to grep and change later" solution 🙈

dantman:
> 👍 for the latest JSX API - I guess the <React.Detached/> component is only for demonstration purposes and not part of this RFC though, correct?

React.Detached would be included since I brought up the use case of rendering a reparent's children while it is detached. And this functionality is useful in non-dom scenarios as well so it makes sense for Detached to not be a ReactDOM/custom thing.

philipp-spiess:
@dantman What do you think about replacing this PR with a newer one that incorporates all the issues discussed? I think this will make the RFC much more approachable since the discussion here is already very long. I remember it took me quite some time to read through all that and we've since doubled the comments 😐 

Let me know if I can help you somehow!

falconmick:
Sorry for being that person, I have not read all of the comments. However I wanted to include a usecaes:

I have a React component that I need to transition from a list of posts that is animating, into a React portal which is outside of the <div id="app"> to avoid position: absolute and transform issues, then it will transform the location to a new spot in the page (the top full width) and re-attach itself to the react dom again.

Basically I'm trying to make the image transition effect you get in allot of IOS/Android apps.

benwiley4000:
> Does this RFC suggest that React would move around the actual DOM nodes? Is it technically possible? For example, if one of the child node was a playing `<video>` element, by removing it from DOM and reinserting into a different parent, would it still be playing?

IMO it should just move the element and leave the developer to do whatever they would do. For what it's worth, the behavior on moving a video element is clearly defined in the HTML spec (https://html.spec.whatwg.org/multipage/media.html#media-playback):

> When a media element is removed from a Document, the user agent must run the following steps:
> 
> 1. Await a stable state, allowing the task that removed the media element from the Document to continue. The synchronous section consists of all the remaining steps of this algorithm. (Steps in the synchronous section are marked with ⌛.)
> 
> 2. ⌛ If the media element is in a document, return.
> 
> 3. ⌛ Run the internal pause steps for the media element.

Note that the element is only paused if it was completely removed from the DOM (i.e. not moved). However in my tests in Firefox and Chrome I found that Firefox follows the spec (continues playing the video) while Chrome violates the spec (pauses regardless of the destination of the element). Nothing is mentioned about currentTime.

I'm filing a bug report for Chromium about this: https://bugs.chromium.org/p/chromium/issues/detail?id=910988

## So the summarize:

According to the spec, video should continue uninterrupted when moved around the DOM, but at worst, you might need to call `.play()` again. This is a compelling use case to support re-parenting in React.

dantman:
😆Coming back to this years later, I'm surprised that my proposed API still holds up. I thought I might need to revise it to work with hooks. But it looks like the proposal still works for functional components with hooks. Only perhaps with a user land helper for creating a singular reparent within a component.

It's hard to sift through the comments and figure out what revisions I accepted. But I am thinking of reversing some of the ones I might have accepted.

There was a lot of discussion about a JSX style API. It seems I accepted and got into this idea because of the similarity to the context render props API. However now that `Context.Consumer` has become something rarely used, with `useContext(Context)` being more common I'm rethinking it. The "declarative magic" of tracking renders where `.keep()` is used is no longer strange now that hooks do effectively that.

I'll need to sleep on some idea of what a reparenting API that fits in with hooks would look like.

ragnaroh:
This might be a stupid question (I'm not familiar with the inner workings of React), but couldn't reparenting be solved by introducing a `globalKey` attribute (akin to the `key` attribute), which must be unique across all components in the component tree? Then upon re-rendering, React would compare the global keys of the previous tree to the global keys of the new tree, and if any keys have "changed locations" in the tree, move the DOM nodes accordingly (as well as internal references to the components).

gaearon:
So, there is a _lot_ to unpack here. As requested in #182, I'm making a pass to review the things that I've been hesitant to review because a proper review would take a very long time to write. So I'll instead leave some cursory comments.

Reparenting is one of the Hard Problems(tm). It basically affects all features at the same time. So the bar for a proposal would have to be extremely high. I think over time we've come to realize that "true reparenting" in all senses of the word usually isn't really needed, and can be a distraction. Here's a few concrete scenarios, and what we tend to use instead:

- "Detachable video player". That seems like a canonical use case for reparenting. E.g. you start watching a video inside a tweet, then you scroll down but the video detaches and keeps playing. Intuitively this seems like a case for reparenting. However, it's not quite clear-cut if you think about what should happen to Context. I think you probably want the Context from the place where the video was detached. Not from the top level where it got reparented _to_. You might also want to keep receiving prop updates from the original place in the feed, e.g. if the like count changes. I know people like to say "UIs are graphs, not trees" but I really don't know that this is actually true. Anyway, for cases like this, it seems like the desired effect is **visual reparenting / DOM reparenting** while the React tree is intact, and the component stays in its original place. This is clumsy but already possible today. The way you do this is by combining refs and portals. The "tweet" (original parent) uses a ref to a DOM node, and places another DOM node into it. It portals the "video" (the child) into that second DOM node. Then, you can imperatively move that second DOM node anywhere in the DOM tree when needed, including reattaching it back. The effect is purely visual, and the actual data flow (and the React tree) stays the same, so there are no difficult questions to resolve.
- "Route transitions". There are actually two different cases.
  -  "Compatible transitions". The example in https://github.com/reactjs/rfcs/pull/34#issuecomment-678592702 mentions page layout. But that already works in React, in general. It only doesn't work if your framework forces you to declare route pages as `Home`, `About`, etc, which use `Layout` inside. But a more React-y way should rely on reconciliation directly. I think this says that routes shouldn't exactly be components, since we really want the Home page and the About page to be able to "line up" because they contain a `Layout`. This is not a purely visual problem: we want to maintain the state (e.g navigation bar hover, search state) as well. Bringing in reparenting here is an overkill when React can already handle reconciliation just fine. It's just that we need a routing abstraction that doesn't opt you out of proper reconciliation by default.
  - "Magic transitions". Sometimes, there really isn't a way to transition properly. Like actually different layouts. This is usually about animations specifically. However, in this case you often want to do it to completely different components anyway. There's no way to line them up. It's more of a layout animation or "magic move" use case where the source and the target might even be written by different people. Framer Motion does a version of this, and building some first-class support for this would be in scope of a React Animations project that we hope to get to in the future. However, that's not really related to the traditional React tree reparenting.

Now, this doesn't mean that reparenting is completely useless. There might be some use cases I've missed. But in many of these, a combination of the above four approaches together with more traditional approaches like lifting state up, is very likely sufficient. On the other hand, reparenting within the React tree raises so many deep questions that it's hard to justify the complexity.

depoulo:
> It's just that we need a routing abstraction that doesn't opt you out of proper reconciliation by default.

Now someone please tell me that's out there already and I don't have to write it myself.

lewisl9029:
> > It's just that we need a routing abstraction that doesn't opt you out of proper reconciliation by default.
> 
> Now someone please tell me that's out there already and I don't have to write it myself.

router5 fits the bill I believe: https://router5.js.org/

I've always been a fan of its routing as _state_ approach over the traditional routing as _entry points_ approach taken by most client side routers. router5's author eloquently explains in [his talk introducing router5](https://www.youtube.com/watch?v=hblXdstrAg0&feature=emb_title), that routing as entry points is a pattern that was blindly adapted from server side routing, which is mostly stateless and transactional (i.e. an API request will only ever result in 1 route branch being exercised for the lifetime of the request), while client side routing is inherently stateful and dynamic (i.e. an SPA will need to keep track of its "current" route and often transition between many different routes and nested routes in a single browsing session), which is where much of the impedance mismatch comes from.

In more concrete terms, an app built with router5 would just have plain react components subscribing to routing state changes and branching accordingly with plain if/else/switch statements and/or map look-ups (this is the actual "routing" part) at any arbitrary depth in the component tree, instead of relying on special snowflake components (i.e. `Route`s and `Router`s) at the highest level of hierarchy responsible for branching between routes and rendering everything downstream. 

It really is a shame that development on the project seems to have stalled. The concept itself is really simple though. I've been using a few custom routing utils in my own projects based on the same principles for a while. The entire routing logic fits in less than 100 lines and is super easy to reason about. Lately I've also been experimenting with adding some aspect of relative routing introduced by react-router v6 (IMHO that's one of the only redeeming ideas in there unfortunately, as the rest of it seems to be doubling down on routing as entry points), which so far seems to be a great fit but it will need a bit more battle testing.

itayperry:
Any progress on this issue? 

> The "tweet" (original parent) uses a ref to a DOM node, and places another DOM node into it. It portals the "video" (the child) into that second DOM node. Then, you can imperatively move that second DOM node anywhere in the DOM tree when needed, including reattaching it back.

saving refs so many levels deep sounds like the wrong way to go - I think it would be better to actually have an API for that.

PR #37, opened 2018-03-29
POJO Elements

iddan:
JSX will be transpiled to object literals, native types to factory functions.

### DIFF
--- a/text/0035-pojo-elements.md
+++ b/text/0035-pojo-elements.md
@@ -1,0 +1,104 @@
+- Start Date: 29 March 2018
+- RFC PR: 
+- React Issue: 
+
+# Summary
+
+JSX will be transpiled to object literals, native types to factory functions.
+
+# Basic Example
+
+```jsx
+<div>
+Hello, World!
+</div>
+```
+
+Will be transpiled to:
+
+```jsx
+{
+  type: ReactDOM.div,
+  props: {
+    children: "Hello, World!"
+  }
+}
+```
+
+Instead of:
+
+```jsx
+React.createElement('div', {}, 'Hello, World!')
+```
+
+Which resolves to:
+
+```jsx
+{
+  $$typeof: Symbol("react.element"),
+  type: "div",
+  props: {
+    children: "Hello, World!"
+  },
+  key: null,
+  ref: null,
+}
+```
+
+# Motivation
+
+ - No more JSX pragma: React-like libraries like Preact would be able to consume React components with **no synthetic syntax modifications**. Of course there can be many other differences but there is enough common ground that worths the changes in my opinion.
+
+ - Allows greater composabillity and easier modification of elements as they are just POJOs.
+ 
+ - Clearer line between ReactDOM components and renderer as types are now factory functions that can validate props themselves
+ 
+ - (Maybe) Reduced render time because React.createElement is not called N times every render
+ 
+# Detailed design
+
+### Security
+
+Unlike in previous discussion this proposal is secure because **type is always a function**.
+Unlike strings a function can't be injected by XSS or similar. If a JSON will be parsed type will remain string and will be considered an error:
+```json
+{ "type": "div", "props": { "children": "Hello, World!" } }
+```
+Any object not conforms to the exact shape of React element will be considered error as well:
+```typescript
+type Element = {
+  type: React.ComponentType<*>,
+  props: Object,
+  key?: any,
+  ref?: (null | Object) => void,
+}
+```
+
+# Possible Drawbacks
+
+ - `React.createElement()` is required to determine element's owner efficently
+ - `ReactDOM` would have to be imported for DOM elements creation which creates a new required pragma - can be discussed further.
+
+# Adoption strategy
+
+Babel plugin would be updated to the new syntax.
+The element APIs (`React.createElement()`, `React.Children`, `React.cloneElement()`...) will be modified to work with the new format.
+
+# How we teach this
+
+- Simplifies the public API of React to the point a component is just a function / class returning a POJO.
+  Example like this can make sense for a lot of people:
+
+```javascript
+const MyComponent = () => ({
+  type: ReactDOM.div,
+  props: {
+    children: 'Hello, World!'
+  }
+})
+```
+
+# Unresolved questions
+
+ - How will consumers use their own DOM elements? (`@jsx` pragma for elements)
+ - How does it affect rendering


text/0035-pojo-elements.md
```diff
@@ -0,0 +1,104 @@
...
+ 
+ - Clearer line between ReactDOM components and renderer as types are now factory functions that can validate props themselves
+ 
+ - (Maybe) Reduced render time because React.createElement is not called N times every render
+ 
+# Detailed design
+
+### Security
+
+Unlike in previous discussion this proposal is secure because **type is always a function**.
```
streamich: What about custom elements, how will their be transpiled?

```jsx
<my-custom-element />
```

text/0035-pojo-elements.md
```diff
@@ -0,0 +1,104 @@
...
+
+Will be transpiled to:
+
+```jsx
+{
+  type: ReactDOM.div,
+  props: {
+    children: "Hello, World!"
+  }
+}
```
streamich: Why not use JSON-ML:

```js
['div', {}, 'Hello, World!']
```

text/0035-pojo-elements.md
```diff
@@ -0,0 +1,104 @@
...
+    children: "Hello, World!"
+  },
+  key: null,
+  ref: null,
+}
+```
+
+# Motivation
+
+ - No more JSX pragma: React-like libraries like Preact would be able to consume React components with **no synthetic syntax modifications**. Of course there can be many other differences but there is enough common ground that worths the changes in my opinion.
```
streamich: To make other libraries consume JSX, like

```jsx
<div>Hello world</div>
```

one could make JSX transpiler emit closures with hyperscript function `h` dependecy injected:

```js
(h) => h('div', {}, 'Hello world');
```

text/0035-pojo-elements.md
iddan: Readability and consistency, but totally an option.

text/0035-pojo-elements.md
iddan: Should be added to questions. Any ideas?

text/0035-pojo-elements.md
jquense: it seems like you wouldn't avoid the pragma, since the compiled-to type `ReactDOM.div` is library specific. Wouldn't this have the same coupling as createElement?

text/0035-pojo-elements.md
```diff
@@ -0,0 +1,104 @@
...
+ - (Maybe) Reduced render time because React.createElement is not called N times every render
+ 
+# Detailed design
+
+### Security
+
+Unlike in previous discussion this proposal is secure because **type is always a function**.
+Unlike strings a function can't be injected by XSS or similar. If a JSON will be parsed type will remain string and will be considered an error:
+```json
+{ "type": "div", "props": { "children": "Hello, World!" } }
```
jquense: how will a consumer of the element be able to determine if the type is a host type or a composite type? At the moment one can do `isHost = typeof element.type === 'string'`

dantman:
Currently one of `createElement`'s jobs is to emit dev mode warnings when a ref or key is invalid. To my knowledge this is done in a function call instead of afterwards based on the POJO because afterwards we don't know where the source of the error was and it makes debugging harder. https://github.com/facebook/react/issues/3228

It also freezes things, adds _self, _source, adds warning getters for ref/key, etc... to help with debugging. And it handles defaultProps too.
https://github.com/facebook/react/blob/master/packages/react/src/ReactElement.js#L110-L257

dantman:
This will also hinder minification:

- `React.createElement('div', {}, 'Hello, World!')` can be reduced to [`a.createElement("div",{},"Hello, World!")`](https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540compilation_level%2520SIMPLE_OPTIMIZATIONS%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Amodule.exports%2520%253D%2520(function(React%252C%2520ReactDOM)%257B%250Areturn%2520%255B%250A%2520%2520React.createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520React.createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520React.createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520React.createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%250A%255D%253B%250A%257D)(React%252C%2520ReactDOM))
- If we transform `React.createElement` to `createElement` then `createElement('div', {}, 'Hello, World!')` can be simplified to [`a("div",{},"Hello, World!")`](https://closure-compiler.appspot.com/home#code%3Dmodule.exports%2520%253D%2520(function(React%252C%2520createElement%252C%2520ReactDOM)%257B%250Areturn%2520%255B%250A%2520%2520createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%252C%250A%2520%2520createElement('div'%252C%2520%257B%257D%252C%2520'Hello%252C%2520World!')%250A%255D%253B%250A%257D)(React%252C%2520React.createElement%252C%2520ReactDOM))
- Unless you enable really advanced (and scary to most people) key mangling, `{type: ReactDOM.div, props: { children: "Hello, World!" } }` can only ever be simplified to [`{type:a.div,props:{children:"Hello, World!"}}`](https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540compilation_level%2520SIMPLE_OPTIMIZATIONS%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Amodule.exports%2520%253D%2520(function(React%252C%2520ReactDOM)%257B%250Areturn%2520%255B%250A%2520%2520%257B%250A%2520%2520%2520%2520type%253A%2520ReactDOM.div%252C%250A%2520%2520%2520%2520props%253A%2520%257B%250A%2520%2520%2520%2520%2520%2520children%253A%2520%2522Hello%252C%2520World!%2522%250A%2520%2520%2520%2520%257D%250A%2520%2520%257D%252C%250A%2520%2520%257B%250A%2520%2520%2520%2520type%253A%2520ReactDOM.div%252C%250A%2520%2520%2520%2520props%253A%2520%257B%250A%2520%2520%2520%2520%2520%2520children%253A%2520%2522Hello%252C%2520World!%2522%250A%2520%2520%2520%2520%257D%250A%2520%2520%257D%252C%250A%2520%2520%257B%250A%2520%2520%2520%2520type%253A%2520ReactDOM.div%252C%250A%2520%2520%2520%2520props%253A%2520%257B%250A%2520%2520%2520%2520%2520%2520children%253A%2520%2522Hello%252C%2520World!%2522%250A%2520%2520%2520%2520%257D%250A%2520%2520%257D%252C%250A%2520%2520%257B%250A%2520%2520%2520%2520type%253A%2520ReactDOM.div%252C%250A%2520%2520%2520%2520props%253A%2520%257B%250A%2520%2520%2520%2520%2520%2520children%253A%2520%2522Hello%252C%2520World!%2522%250A%2520%2520%2520%2520%257D%250A%2520%2520%257D%250A%255D%253B%250A%257D)(React%252C%2520ReactDOM))

The `props`, `children`, and `types` keys can never be optimized away and make the code larger when minified. And JSX is used a **lot** in general making space wasting syntax in the JSX transform output grow exponentially.

text/0035-pojo-elements.md
thysultan: From a performance and compostability perspective, the current semantics allows react and react-like libs to have more control of the resulting data-structure created between patch/minor sem-versions, this would go against that, forcing them to conform to consume what ever data-structure is agreed upon within the reconcile phase.

When considering that these libs add meta data to the snapshot data-structures, this would effectively kill the monomorphic nature that one has control over with the current pragma composition pattern since there would still be a need to add lib specific meta data. i.e [Inferno](https://github.com/infernojs/inferno/blob/master/packages/inferno/src/core/implementation.ts#L53-L68) emits it's own meta data for performance reasons, [Vue](https://github.com/vuejs/vue/blob/dev/src/core/vdom/vnode.js) emits it's own meta data as does [DIO](https://github.com/thysultan/dio.js/blob/9.0.0/src/Core/Element.js#L1-L22) and i'm sure Mithril, Preact and other reactive libs that emit a primitive snapshot structure include specific meta data for their needs.

j-f1:
Comparison:

```js
// numbers are how many characters shorter
a.createElement("div",{},"Hello, World!")//34 (4 ch)
a("div",{},"Hello, World!")//3456789012345678 (18ch / 14ch)
{type:a.div,props:{children:"Hello, World!"}}
```

text/0035-pojo-elements.md
iddan: Well no, because anything not interacting directly with DOM elements or allows providing replacement components will benefit from this change. Anyway Preact uses class and React uses className and there are other small differences but many components will still be able to target both

text/0035-pojo-elements.md
iddan: isHost = element.type instanceof ReactDOM.Element

text/0035-pojo-elements.md
yordis: @streamich what is that JSON-ML? His example it is just JSON, what is the benefits of JSON-ML?

iddan:
If following @streamich proposal JSON ML syntax would actually reduce compiles code size.

text/0035-pojo-elements.md
iddan: 1. It’s a more compact in code and memory format
2. It’s a standard for describing XML like data structures in JSON

text/0035-pojo-elements.md
iddan: I see React as more than just a library but an ecosystem and reference for other libraries. This proposal suggests a trade off a custom element instanciation for standard data structure to represent the virtual DOM

streamich:
@iddan Here JSON-ML is discussed:

- https://github.com/facebook/react/issues/2932

trueadm:
There's a few things I want to point out here:

- As others have already mentioned, we use `React.createElement()` to do far more than just create "virtual nodes", they are used for dev-time related checks and validation.
- We also handle things like "key" and "ref" that are special cases and require runtime logic to make this work. Take for example `<MyComponent {...stuff} />`, if `stuff` contains `key` or `ref`, we need to extract those from the props – something we can't do statically if we don't know if either property exists.
- In terms of performance, I've found that using a lightweight monomorphic function to create the POJO was far faster than creating an inline object – because there's no guarantee that all the objects for ReactElement will share the same hidden class unless we ensure they always have the same properties. To do that, we'd need to ensure that both `ref` and `key` are added to every ReactElement POJO, even if they are not used (they would be set to `null`).
- There's far too much focus on JSX -> virtual DOM here. That might be what JSX de-sugars to, but in the future I expect we'll make compile time changes in React and move away from virtual DOM entirely.

I'm not too sure this RFC brings any benefits to React or its ecosystem because of what I mentioned above.

streamich:
> ...and move away from virtual DOM entirely.

Interesting.

text/0035-pojo-elements.md
bvaughn: Perhaps I'm misunderstanding this statement, but type isn't always a function. For example:
* `React.forwardRef`, `Context.Provider`, and `Context.Consumer` are objects
* `React.StrictMode` and `React.AsyncMode` are symbols
* Many React Native host types (e.g. `View`) are strings

This proposal seems very ReactDOM-centric.

text/0035-pojo-elements.md
bvaughn: What happens to `key` and `ref` attributes? Although not relevant to this specific example, these are important. Moving them into `props` would be a pretty big backwards breaking change (because of e.g. `{...props}`).

iddan: They are practically props aren't they today? What mechanisms differentiate them?

text/0035-pojo-elements.md
iddan: The problem is only with bare strings. So as long as type is an object, symbol or function there is no problem.

iddan:
@trueadm I don't think this proposal has anything to do with the virtual DOM actually. This is a proposal to represent JSX in a more generic and native JS way.

text/0035-pojo-elements.md
bvaughn: > Many React Native host types (e.g. View) are strings

text/0035-pojo-elements.md
bvaughn: They aren't props. They're stored separately on `ReactElement`. If they were in the `props` argument, then spreading props (e.g. `<Foo bar="abc" {...props} />`) would pass `ref` and `key` too. Particularly with refs, this could cause a problem, as it could result in the same `ref` callback being called for many types of element.

dantman: Current behaviour:
```js
const React = require('React');
const el = <div b='c' ref={() => {}} key='key' />;
el.props.b; // 'c'
el.props.ref; // undefined
el.props.key; // undefined
```

sebmarkbage:
I think any proposal like this would like want an intermediate upgrade path in the ecosystem using something like https://github.com/reactjs/rfcs/pull/107 regardless.

After that the question is basically if it should be a function call or inline. All our testing suggest that a function call is the way to go in this case because it allows the VMs to dedupe a lot of the code gen that happens for each of these callsites. So it'd probably end up being a function for that reason. It could in theory be a built-in function in the transpiler runtime rather than on React though. That's covered in #107 too though.

iddan:
@sebmarkbage Can you elaborate
> All our testing suggest that a function call is the way to go in this case because it allows the VMs to dedupe a lot of the code gen that happens for each of these callsites. So it'd probably end up being a function for that reason.

PR #38, opened 2018-03-31
Support named exports in Node.js native ES2015 modules implementation

dantman:
[View formatted RFC](https://github.com/dantman/rfcs/blob/native-esmodule-support/text/0000-native-esmodules.md)

I would appreciate it if we kept the merits of `.mjs` and how Node.js chose to implement ES2015 modules out of the discussion here. We should just accept the fact that ES2015 modules are getting a native implementation in Node.js, this is what we have to work with, and focus on fixing React so that it works with the implementation we are getting.

### DIFF
--- a/text/0000-native-esmodules.md
+++ b/text/0000-native-esmodules.md
@@ -128,9 +128,6 @@ export const createFactory = React.createFactory;
 export const isValidElement = React.isValidElement;
 export const version = React.version;
 
-export const Component = React.Component;
-export const PureComponent = React.PureComponent;
-
 export default React;
 ```
 
--- a/text/0000-native-esmodules.md
+++ b/text/0000-native-esmodules.md
@@ -1,0 +1,176 @@
+- Start Date: 2018-03-30
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Node.js is introducing [native support of ES2015 modules](https://nodejs.org/api/esm.html). This will be available in Node.js 10 which is marked as being released in April 2018 on the roadmap.
+
+For brevity **ESM** will be used to refer to the native implementation of ES2015 modules in Node.js with `.mjs` files.
+
+Due to significant differences in how JavaScript behaves in ESM vs. CommonJS code this is being implemented using a new `.mjs` file extension.
+
+ESM imports require named exports to be known ahead of time. However the names of CommonJS exports are not known until runtime and CommonJS also allows the entire `module.exports` to be replaced with a non-object. As a result when an ESM module imports CommonJS code it can only do so using the `default` export. You cannot use named imports in ESM code to import named exports from CommonJS code.
+
+React only exports a CommonJS module from the package. As a result named exports do not work.
+
+I recommend we add a small ESM export to React to allow named exports from React to work in ESM code.
+
+# Basic example
+
+The following examples are the current behaviour of React in various contexts.
+
+If you use React in CommonJS modules written in ES2015 you use it as such:
+
+```js
+const React = require('react');
+class MyComponent extends React.Component {}
+```
+
+```js
+const {Component} = require('react');
+class MyComponent extends Component {}
+```
+
+If you write ES2015 modules but transpile to CommonJS using Babel, use Rollup, or use WebPack the following will work:
+
+```js
+// Default imports
+import React from 'react';
+class MyComponent extends React.Component {}
+```
+
+```js
+// Named imports
+import React, {Component} from 'react';
+class MyComponent extends Component {}
+```
+
+If you write ES2015 modules and deliver them to the browser with `<script type="module">` you cannot use React without either making React a global script or using some extra step to convert the CommonJS React code into a ESM bundle. I have not personally investigated this option at the moment, but I believe this likely suffers from the same limitations as ESM.
+
+If you write ESM code: ES2015 module code in a `.mjs` file run by Node with `--experimental-modules` enabled (or without the flag in the upcoming Node 10). The following will work:
+
+```js
+// Default imports
+import React from 'react';
+class MyComponent extends React.Component {}
+```
+
+However in ESM code currently the following will break, as Component will be undefined:
+
+```js
+// Named imports
+import React, {Component} from 'react';
+class MyComponent extends Component {}
+```
+
+This proposal asks that we change this behaviour so that the former works in ESM code, just like it works in transpiled/bundled code.
+
+# Motivation
+
+Officially the React documentation only documents the use of named imports in examples. In fact the ES2015 source code of React just does an `export default React;` without exporting any named exports. Named imports only work due to the CommonJS bundling and exclusion of a `"module"` variant.
+
+However despite not being part of the official docs, named exports from React are used regularly in the React community.
+
+Here is a small sample list of popular libraries in the React ecosystem using named imports in ES2015 modules. Links go to a random .js file in the library's official repo showing one of their components using named imports.
+
+- [react-redux](https://github.com/reactjs/react-redux/blob/master/src/components/Provider.js)
+- [redux-form](https://github.com/erikras/redux-form/blob/master/src/Form.js)
+- [Downshift](https://github.com/paypal/downshift/blob/master/src/downshift.js)
+- [react-intl](https://github.com/yahoo/react-intl/blob/master/src/components/message.js)
+- [react-dnd](https://github.com/react-dnd/react-dnd/blob/master/packages/react-dnd/src/DragDropContext.js)
+- [Create React App's default template](https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/template/src/App.js) (and all users of *CRA* who just follow how the default *CRA* template imports React)
+
+All of these are users of the non-spec transpiled/bundled version of ES2015 modules and work due to it's non-spec handling of CommonJS exports. However if any of these parts of the community decide they wish to use or at least support the use of native ESM code, currently they will have to refactor their entire codebase to use `React.Component` off a default import instead of using named imports
+
+# Detailed design
+
+Here are the requirements that the ESM implementation in Node.js place on us and it's behaviours we can take advantage of:
+
+- ESM code **must** be inside of a `.mjs` file
+- ESM code **must** use ES2015 `import` and `export` and cannot use `require`
+- ESM code **must** declare all of its named exports in the `.mjs` file or they must come from another `.mjs` file exported using `export * from './somefile'`. (You likely could use `*` on a CommonJS module, but it would only export `default`)
+- If the file being imported by an ESM `import` is a CommonJS module the entire module's exports object will be available as the `default` export, named exports will all be undefined
+- A CommonJS module cannot `require` an ESM module, however it can asynchronously `await import('./somefile.mjs')` one
+- A CommonJS module's `require` with no extension will look for `.js` files but will not look for `.mjs` files
+- A ESM `import` without an extension will **first** look for an `.mjs` file and will only look for a `.js` if an ESM module is not found
+- As a result if you place a CommonJS `module.js` and an ESM `module.mjs` next to it `require('./module')` will require the `module.js` and `import './module';` will import the `module.mjs`
+- This behaviour is the same for the main `index.js`/`index.mjs` import of the package if you define `main` as `"main": "./index"` without the file extension
+- ESM code may not use conditional imports, but CommonJS modules may
+- The identifiers of all named exports from a ESM module must be defined ahead of time, **however** the value of those exports do **not** need to be defined in ESM code
+
+Thus it should be sufficient to do the following:
+- Change the `"main"` of react packages to be `"index"` instead of `"index.js"`
+- Add an `index.mjs` that does an `import React from './index.js';` and `export default React;`
+- In this file provide explicit exports of all the stuff on `React.*` that we wish to export
+
+For example, `react/packages/react/npm/index.mjs` may look like this.
+
+```js
+import React from './index.js';
+
+export const Children = React.Children;
+
+export const createRef = React.createRef;
+export const Component = React.Component;
+export const PureComponent = React.PureComponent;
+
+export const createContext = React.createContext;
+export const forwardRef = React.forwardRef;
+
+export const Fragment = React.Fragment;
+export const StrictMode = React.StrictMode;
+export const unstable_AsyncMode = React.unstable_AsyncMode;
+
+export const createElement = React.createElement;
+export const cloneElement = React.cloneElement;
+export const createFactory = React.createFactory;
+export const isValidElement = React.isValidElement;
+export const version = React.version;
+
+export const Component = React.Component;
+export const PureComponent = React.PureComponent;
+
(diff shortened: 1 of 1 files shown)


text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+
+All of these are users of the non-spec transpiled/bundled version of ES2015 modules and work due to it's non-spec handling of CommonJS exports. However if any of these parts of the community decide they wish to use or at least support the use of native ESM code, currently they will have to refactor their entire codebase to use `React.Component` off a default import instead of using named imports
+
+# Detailed design
+
+Here are the requirements that the ESM implementation in Node.js place on us and it's behaviours we can take advantage of:
+
+- ESM code **must** be inside of a `.mjs` file
+- ESM code **must** use ES2015 `import` and `export` and cannot use `require`
+- ESM code **must** declare all of its named exports in the `.mjs` file or they must come from another `.mjs` file exported using `export * from './somefile'`. (You likely could use `*` on a CommonJS module, but it would only export `default`)
```
Kovensky: `export * from` explicitly ignores reexporting a `default` export; so it'd export nothing.

To completely reexport all exports of a module you need both `export * from` and `export { default } from` for the same module.

text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+
+export const Fragment = React.Fragment;
+export const StrictMode = React.StrictMode;
+export const unstable_AsyncMode = React.unstable_AsyncMode;
+
+export const createElement = React.createElement;
+export const cloneElement = React.cloneElement;
+export const createFactory = React.createFactory;
+export const isValidElement = React.isValidElement;
+export const version = React.version;
```
Kovensky: These can even be drastically simplified to:

```js
export const {
  Children,
  createRef,
  Component,
  // ... etc
} = React
```

text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+- The source code of `packages/react/src/React.js` will need to be changed to both export named exports like `export {createRef};` and `export default` the default React object.
+  - Alternatively, React.js could exclusively export named exports and an alternative to `index.js` could both `export * from './React';` and `import * as React from 'react'; export default React;` to get the `React` object as a free side effect.
+- The build process will need to output `./esm/react.{production.min,development}.mjs` bundles in addition to the normal `./cjs/` bundles. I expect these will just be the Rollup bundle already being generated but without the module transpilation step.
+- A `index.mjs` will sit at the root beside `index.js`, it will act the same as `index.js` but will export the `./esm` bundles using ES2015 exports instead of the `./cjs` bundles.
+  - Unresolved: It is not actually possible to do the `if (process.env.NODE_ENV === 'production') {` conditional that `index.js` does in ESM, ES modules require that import is at the top of the file and does not permit conditional importing.
```
Kovensky: it is possible but only with dynamic `import()`, and that also requires that anything that must go through that code path be made async.

gaearon: I think the most natural solution is to do conditional re-exports.
Like

```
export const createElement = (process.env.NODE_ENV === 'production') ? createElementProd : createElementDev;
```

in the entry point. I don't think the ecosystem is ready for this yet though, given how fragile tree shaking is.

text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+
+However this may be acceptable for now as a transition period because browsers do not support the non-relative node style imports like importing `fbjs/lib/invariant` that React makes use of.
+
+# Alternatives
+
+## Drop support
+
+Alternatively we could decide that we are not going to support named imports of React from ESM.
+
+This will not be a breaking change. Use of .mjs and ESM is opt-in functionality. All packages using babel or a bundler to access node will keep working with the non-spec compliant handling of named exports from CommonJS modules.
```
gaearon: Unless their build system automatically chooses mjs when it's available. In this case it would be breaking.

text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+
+The only requirement will be that if someone decides to switch to `.mjs` and opt-in to using ESM instead of transpiling/bundling, they will be required to refactor their entire codebase to use `React.Component` instead of named exports.
```
gaearon: Also that 90% of tutorials / examples will be wrong. Seems like a pretty bad situation to be in. On the other hand TS already enforces using named imports only. I dunno.

text/0000-native-esmodules.md
TrySound: Here's how you can test treeshakability without assumptions.
https://github.com/TrySound/rollup-plugin-size-snapshot/blob/master/src/treeshakeWithWebpack.js#L15

text/0000-native-esmodules.md
TrySound: I already migrated in a bunch of react packages. I think the best thing we can do is a migration to `import * React from 'react'` in docs and blaming `import React, { Component } from 'react'` which doesn't fit to any way. I could start this migration if you will write clean post in this thread.

streamich: You can use default exports with TypeScript. But it is true that using default import for React from TypeScript does not work.

text/0000-native-esmodules.md
dantman: .mjs files aren't supposed to be imported from require in .js files (including all current babel ES6 import usage in .js, which really uses require behaviour under the hood).

In fact the esm page explicitly states that "`require('./foo.mjs')` is unsupported because "ES Modules have differing resolution and timing".

Any build-system automatically choosing .mjs when available would be violating ESM.

text/0000-native-esmodules.md
Kovensky: Migrating to `import * as React from 'react'` is the _wrong_ migration to do; it only "works" if you are using TypeScript's old commonjs emit. You should be using `--esModuleInterop` and using `import React from 'react'` instead.

TrySound: Can't typescript handle both cases with interop like everybody does? Rollup, webpack, babel handle this without problems.

TrySound: Can this be fixed in perspective so we could start migrating to `import * as React from 'react'`. Or do you think we stuck with default export forever?

Kovensky: That is what this RFC is meant to address.

chyzwar:
It is unknown when a node will support ESM and what will be the form. https://github.com/nodejs/modules Modules team in node did not publish timeline and previous PRs in the node were postponed. This is already the third attempt of ESM in the node. IMO focus should be on browser ESM implementation that works now.

gaearon:
We also need to consider what to do with `react-test-renderer/shallow` entry point which currently exports a function in CommonJS.

TrySound:
We can do the same or just change exports to
```js
import { TestRenderer, Shallow } from 'react-test-renderer';
```

sebmarkbage:
I think the general thinking here is that we want to do something like this but it'd be a breaking change and only worth doing along with other breaking changes...?

dantman:
> I think the general thinking here is that we want to do something like this but it'd be a breaking change and only worth doing along with other breaking changes...?

The Node.js ESM implementation has been pushed back to October and has changed a bit in implementation. So the RFC is kind of on hold.

Theoretically this change should not be breaking since it only affects people writing `.mjs` files. Additionally Node has even changed the behaviour in a way that adding an index.mjs won't break current users using `.mjs`. That said there's a chance WebPack could completely screw up that plan. So perhaps it should be bundled with breaking changes anyways.

It's tangentially related. But I wonder if we should also revisit TypeScript. It's really annoying having to use `import * as React from 'react';` instead of `import React, {useState} from 'react';`. Which could easily be solved with essentially a `const React = {...}; modules.export = {...React, default: React};`.

TrySound:
@dantman Default export is not the right way to import from react. `React` is a namespace. Logically it should not be default export. Default export also requires hacks like babel plugin to transform
```js
React.createElement()
// to
const { createElement } = React
createElement()
```

When bundler may do this for free and even better will not generate `const { createElement } = React` in every file.

Treeshaking will not be significant with react but it still can be considered as a feature.

Jessidhia:
React *should be* a namespace. It currently is not.

dantman:
> Default export is not the right way to import from react. `React` is a namespace.

You're going to have to rally for JSX to be changed then. Because the default is `React.createElement` and `React` must be defined in context.

> Treeshaking will not be significant with react but it still can be considered as a feature.

Treeshaking will be completely nonexistent. React is bundled into a single file and uses an object export.

TrySound:
> You're going to have to rally for JSX to be changed then. Because the default is React.createElement and React must be defined in context.

Not at all. `import * as React from 'react';` is universal solution. Default export can be provided as a temporary fix.

> Treeshaking will be completely nonexistent. React is bundled into a single file and uses an object export.

I'm talking about the time when react will have esm.

dantman:
> I'm talking about the time when react will have esm and named exports.

React's single file bundles were an intentional choice, even if React ever gets ESM it will likely still use single file bundles.

TrySound:
@dantman Wait, I didn't say anything about react bundles. I like them.

My point was that react should not have default export because it's increases user bundle a lot or requires babel plugins to solve the problem and still with trade offs.

gaearon:
How does this proposal hold up against the latest shipped Node implementation?

dantman:
> How does this proposal hold up against the latest shipped Node implementation?

There seem to be some changes to the ESM handling. The general idea of the RFC seems to still be valid, but about 10% needs a few tweaks for the ESM changes.

Scanning the new https://nodejs.org/api/esm.html docs here are the things that stand out:

- File extensions are now mandatory, so the wrapper would need to explicitly use `.mjs` and the suggestion of `./index` for the main field would be removed.
- `.js` files in a package can also be ESM if you include `"type": "module"` in the package.json but we're not changing the default (unless you want to switch to ESM code for React 17 with optionally a CommonJS .cjs fallback for compatibility with older node) so we'd do the opposite and explicitly declare `"type": "commonjs"`.
- Instead of placing an `index.js` and `index.mjs` next to each other, package.json has a new `exports` field. It supports conditional exports to direct `require` and `import` to different files, so this is how you'd setup that instead of an extension-less main.
  - It's not relevant here, but if `exports` gains wider adoption it could mean the disappearance of needing to include `lib`/`dist`/`es`/`module` folders in imports.
- The proposal I made of using an .mjs wrapper file to export named identifiers appears to be the recommended approach for packages vulnerable to the dual package hazard (where one dep doing `import` and another doing `require` would result in two instances of React being imported). So the general idea of the RFC seems to be validated by this.
- The ESM docs propose a `browser` conditional export variant (like the variants for import/require/node). If tooling adopts this (whatever tooling people would be using to generate browser native type=module packages) then this could possibly be a solution in the future for supporting browser native module packages without dropping the compatibility with CommonJS-only versions of node. (Basically export a full module version of React and only target it to browsers. This avoids the dual package hazard because Node still uses the wrapper not vulnerable and a browser that only supports ESM won't import the CJS version).

@gaearon Is there any desire to implement this anytime soon? If so I can revisit this now and update the RFC. Otherwise I can revisit it the next time there is new news on Node's ESM implementation (changes from experimental to stable, finishes the loader API, other toolkits adopt the ESM behaviours, or a number of React users indicate they wish to use ESM).

jaydenseric:
> How does this proposal hold up against the latest shipped Node implementation?

The import specifiers in ESM `.mjs` files *must* contain the full filenames, including extensions (except when using a bare specifier to import from main package exports).

E.g. when importing ESM into ESM:

```diff
- import { Foo } from './Foo'
+ import { Foo } from './Foo.mjs'
```

E.g. when importing CJS into ESM:

```diff
- import Foo from './Foo'
+ import Foo from './Foo.js'
```

gaearon:
>@gaearon Is there any desire to implement this anytime soon?

"Soon" is relative but it's always helpful to have an up-to-date plan in case it seems like the right moment to start the work. In particular, I'm worry about how to be compatible with Node, bundlers, Flow/TS, and everything else.

text/0000-native-esmodules.md
```diff
@@ -0,0 +1,173 @@
...
+- [react-dnd](https://github.com/react-dnd/react-dnd/blob/master/packages/react-dnd/src/DragDropContext.js)
+- [Create React App's default template](https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/template/src/App.js) (and all users of *CRA* who just follow how the default *CRA* template imports React)
+
+All of these are users of the non-spec transpiled/bundled version of ES2015 modules and work due to it's non-spec handling of CommonJS exports. However if any of these parts of the community decide they wish to use or at least support the use of native ESM code, currently they will have to refactor their entire codebase to use `React.Component` off a default import instead of using named imports
+
+# Detailed design
+
+Here are the requirements that the ESM implementation in Node.js place on us and it's behaviours we can take advantage of:
+
+- ESM code **must** be inside of a `.mjs` file
```
Andarist, line 90: This is not technically correct - you can use `.js` if you put within a directory containing package.json with `{"type": "module"}` and you can even use `.js` extension for both CJS and ESM entries, u "just" need to extra package.json files. For example, Rollup is using this "trick": https://unpkg.com/browse/rollup@2.23.0/dist/es/

dantman: > This could have not been true at the time of writing this RFC though.

Correct. At the time of writing the RFC this is how the Node.js ESM implementation worked. It was only available in .mjs files and the `type` in package.json did not exist. The implementation has changed significantly and `type` was noted in my list of differences.

For React's purpose though, `.mjs` would still be used for the ES module wrapper unless the project were to make a large breaking change switching entirely to a ES modules codebase.

stefee:
This RFC seems to be getting some attention at the moment. I’d like to direct folks to the Node.js documentation, and in particular the “Dual CommonJS/ES module packages” section which provides a more complete and up-to-date picture: https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_dual_commonjs_es_module_packages

The section titled [Writing dual packages while avoiding or minimizing hazards - Approach #1: Use an ES module wrapper](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_approach_1_use_an_es_module_wrapper) seems relevant to React in its present state. Here’s a quote:

> This approach is appropriate for any of the following use cases:
> 
> * The package is currently written in CommonJS and the author would prefer not to refactor it into ES module syntax, but wishes to provide named exports for ES module consumers.
> * The package has other packages that depend on it, and the end user might install both this package and those other packages. For example a utilities package is used directly in an application, and a utilities-plus package adds a few more functions to utilities. Because the wrapper exports underlying CommonJS files, it doesn’t matter if utilities-plus is written in CommonJS or ES module syntax; it will work either way.
> * The package stores internal state, and the package author would prefer not to refactor the package to isolate its state management. See the next section.

gaearon:
Our latest thinking was that we'd drop default exports altogether. Named only.

jonkoops:
> Our latest thinking was that we'd drop default exports altogether. Named only.

Perhaps this should be deprecated starting React 19?

PR #40, opened 2018-04-05
`prevProps` in `getDerivedStateFromProps()`

catamphetamine:
[Adding `prevProps` argument to `getDerivedStateFromProps()`](https://github.com/catamphetamine/rfcs/blob/master/text/text/0000-prevProps-in-getDerivedStateFromProps.md)

Don't hesitate to close this RFC if it doesn't fit: I just thought it would be interesting to submit my first RFC and share some ideas.
I'm not persisting in pushing this feature.

PR #44, opened 2018-04-21
Wrap the arguments of React class/instance methods with `{}`

wood1986:
See [this](https://github.com/wood1986/rfcs/blob/2b8e9df2481e8ad08b4b489a8427c141514ac527/text/0000-single-object-arg-for-all-methods.md)

### DIFF
--- a/text/0000-single-object-arg-for-all-methods.md
+++ b/text/0000-single-object-arg-for-all-methods.md
@@ -15,7 +15,7 @@ componentDidUpdate(prevProps, prevState, snapshot) -> componentDidUpdate({prevPr
 
 # Motivation
 
-With that change, The arguments' order of the class/instance method is no longer a barrier when contributors need to change/correct the existing class/method
+With that change, The arguments' order of the class/instance methods is no longer a barrier when contributors need to change/correct the existing class/instance method
 
 # Drawbacks
 
--- a/text/0000-single-object-arg-for-all-methods.md
+++ b/text/0000-single-object-arg-for-all-methods.md
@@ -3,13 +3,15 @@
 
 # Summary
 
-Wrap the arguments of React class/instance methods with curly brace
+Wrap the arguments of React class/instance methods with `{}`
 
 # Basic example
 
-`static getDerivedStateFromProps(nextProps, prevState) -> static getDerivedStateFromProps({nextProps, prevState})`
-`shouldComponentUpdate(nextProps, nextState) -> shouldComponentUpdate({nextProps, nextState})`
-`componentDidUpdate(prevProps, prevState, snapshot) -> componentDidUpdate({prevProps, prevState, snapshot})`
+```
+static getDerivedStateFromProps(nextProps, prevState) -> static getDerivedStateFromProps({nextProps, prevState})
+shouldComponentUpdate(nextProps, nextState) -> shouldComponentUpdate({nextProps, nextState})
+componentDidUpdate(prevProps, prevState, snapshot) -> componentDidUpdate({prevProps, prevState, snapshot})
+```
 
 # Motivation
 
@@ -17,7 +19,7 @@ With that change, The arguments' order of the class/instance method is no longer
 
 # Drawbacks
 
-No matter how this will be a breaking change. But there is no need a transition becuase the existing application just simply wrap the arguments with {}.
+No matter how this will be a breaking change. But there is no need a transition becuase the existing application just simply wrap the arguments with `{}`.
 The destructuring-simple in http://incaseofstairs.com/six-speed/ does not show any performance issue.
 
 # Adoption strategy
--- a/text/0000-single-object-arg-for-all-methods.md
+++ b/text/0000-single-object-arg-for-all-methods.md
@@ -15,12 +15,12 @@ componentDidUpdate(prevProps, prevState, snapshot) -> componentDidUpdate({prevPr
 
 # Motivation
 
-With that change, The arguments' order of the class/instance methods is no longer a barrier when contributors need to change/correct the existing class/instance method
+With that change, the arguments' order of the class/instance methods is no longer a barrier when contributors need to change/correct the existing class/instance method
 
 # Drawbacks
 
-No matter how this will be a breaking change. But there is no need a transition becuase the existing application just simply wrap the arguments with `{}`.
-The destructuring-simple in http://incaseofstairs.com/six-speed/ does not show any performance issue.
+- No matter how this will be a breaking change. But there is no need a transition becuase the existing application just simply wrap the arguments with `{}`.
+- It might have performance issue but the destructuring-simple in http://incaseofstairs.com/six-speed/ does not show it.
 
 # Adoption strategy
 
--- a/text/0000-single-object-arg-for-all-methods.md
+++ b/text/0000-single-object-arg-for-all-methods.md
@@ -1,0 +1,25 @@
+- Start Date: 2018-04-20
+- React Issue: (leave this empty)
+
+# Summary
+
+Wrap the arguments of React class/instance methods with curly brace
+
+# Basic example
+
+`static getDerivedStateFromProps(nextProps, prevState) -> static getDerivedStateFromProps({nextProps, prevState})`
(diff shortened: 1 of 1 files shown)


dantman:
This is a pretty extreme breaking change for a questionable benefit. Entire codebases must be changed all at once to upgrade. This alone would be reason to reject it because there are organizations like Facebook with such large codebases that updates must be done incrementally. But on top of this, this means that **every single React library using any of these methods must release a breaking change all at once**. If you use a bunch of libraries you have to coordinate updates so that every one of them is upgraded to the correct major version when you upgrade React. You cannot do this change before or after upgrading React, it must be done at the same time. And if you make use of a single tiny little library that lags behind in maintenance, you cannot upgrade React until it updates.

dantman:
> With that change, the arguments' order of the class/instance methods is no longer a barrier when contributors need to change/correct the existing class/instance method

Could you reference any past issue or rfc where argument order has been a barrier that blocked a change to the React api?

I am not personally aware of any. In fact the most recent rfc involving changing the arguments of a lifecycle hook (adding `prevProps` to `getDerivedStateFromProps`) is one that would not be helped by `getDerivedStateFromProps` accepting an object argument. In fact, it would make things even harder. When using normal arguments it's possible for the caller to check the arity of the function (how many arguments are defined in the function definition) and use that to implement a backward compat behaviour and emit a deprecation warning, to make breaking changes easier to migrate. However if we switch to using single object arguments this is no longer possible because the user defined functions now always have a single argument.

This also locks in argument names and mandates that users use them. `{next,prev}{Props,State}` is the React documentation convention. However we do not actually require this as the pattern, we only require `this.props` and `this.state` to use those names. I would bet that there are users who use the following patterns:

- `shouldComponentUpdate(newProps, newState)`
- `componentDidUpdate(oldProps, oldState)`
- `static getDerivedStateFromProps(props, state)`

Anyone who does not follow the `{next,prev}{Props,State}` naming pattern will not only have to wrap their arguments in `{}`, they will have to rename their arguments. This also means that the code in the function itself will need to be refactored to use the new argument name.

dantman:
It should also be noted that destructuring is an ES6/ES2015 feature and we technically still support users who are still writing ES5 code without transpiling it:
https://reactjs.org/docs/react-without-es6.html

In ES5 code destructuring will not be available and code will need to use a verbose reference to something like `opts.prevProps.*`.

```js
var Greeting = createReactClass({
  componentDidUpdate: function(opts) {
    if ( opts.prevProps.name !== this.props.name ) {
      console.log('name has changed');
    }
  },
  render: function() {
    return <h1>Hello, {this.props.name}</h1>;
  }
});
```

bvaughn:
Thanks for the proposal, @wood1986! I agree with the sentiment that named params can make certain types of changes easier.

I think @dantman makes several strong arguments against this change though. Given that, and the lack of discussion in the last couple of weeks- I'm going to go ahead and close this RFC for now.

wood1986:
Sorry guys, I completely forget to reply. I do not mind closing but I still want to leave my comment. It will be very long.

Here are my subjective reason
1. I personally do not feel confident about the function signature of getDerivedStateFromProps and getSnapshotBeforeUpdate or any new methods in the future (Again, the lifecycle is fine except the function signature) because they are new and controversial (#40). @bvaughn HAD some plans to modify them. If core team want to change, this will be a problem as lots of people has already been in old signature. And they may be difficult to write the polyfill for the smooth migration.
2. I have a "prediction" with the new lifecycle and static getDerivedStateFromProps, I do not think we need the props anymore in the future. I am talking about the stateful React class component not the stateless functional component because getDerivedStateFromProps has forced us to move every variable which will affect the rendering directly or indirectly to the state. Of course the communication from parent to child will still need props but after getDerivedStateFromProps this.props or prevProps has becomed useless. If react core team plans to remove this.props or prevProps, this will be very big change and difficult to write the polyfill for the smooth migration.

Here are my reply to @dantman
> This is a pretty extreme breaking change for a questionable benefit.

I am thinking from core team perspective when they need to change the existing signature in order to introduce some big features or correct the some existing signature. Of course, it minimizes the cost for writing a polyfill. From the consumer side, I can tell there is no benefit.

> Entire codebases must be changed all at once to upgrade. This alone would be reason to reject it because there are organizations like Facebook with such large codebases that updates must be done incrementally. But on top of this, this means that every single React library using any of these methods must release a breaking change all at once. If you use a bunch of libraries you have to coordinate updates so that every one of them is upgraded to the correct major version when you upgrade React. You cannot do this change before or after upgrading React, it must be done at the same time. And if you make use of a single tiny little library that lags behind in maintenance, you cannot upgrade React until it updates.

React itself just needs to introduce one major release which is asking the consumer to rename all variables. There will no logic change and no lifecycle change. Again I am not asking for logic change.

> In fact, it would make things even harder. When using normal arguments it's possible for the caller to check the arity of the function (how many arguments are defined in the function definition) and use that to implement a backward compat behaviour and emit a deprecation warning, to make breaking changes easier to migrate. However if we switch to using single object arguments this is no longer possible because the user defined functions now always have a single argument.

Both can achieve the same thing and you still can check the existence of the certain properties. The problem of n numbers of arguments is the order cannot change once you introduce. Or you need to do some checking if you need to change the existing signature. Using the number of arguments to identify the break change is still implicit, however, using the object's property name is definitely clear.

> This also locks in argument names and mandates that users use them. {next,prev}{Props,State} is the React documentation convention. However we do not actually require this as the pattern, we only require this.props and this.state to use those names

Think about which one will have least impact and less buggy? locking the name vs locking the order?

wood1986:
> I don’t see how this addresses @dantman’s concern.

IMO, if it is just a naming change without logic change, that breaking change is minor and it will be a find and replace task. If that big change cause the inconvenience to the third party libs, here is my idea.
Each component has a static var called `signatureMode` and it requires to be set explicitly 

```
// consumer.js
class F extends React.Component {
}

F.signatureMode = "1Arg"

// react.js
if (Component.signatureMode === "1Arg") {
  shouldComponentUpdate({prevProps, prevState})
} else if (Component.signatureMode === "nArgs") {
  shouldComponentUpdate(prevProps, prevState)
} else {
  console.warn("Component.signatureMode is not set correctly. 'nArgs' will deprecate in v20.0.0 and default is set to '1Arg'");
  shouldComponentUpdate({prevProps, prevState})
} 
``` 

React.

> Can you give a code example of how that would work with destructuring?

@dantman had mentioned 
```
componentDidUpdate(obj) {
   var prevProps = obj.prevProps;
       prevState = obj.prevState; 
}
```

j-f1:
> IMO, … it is just a naming change without logic change

Every single package that exports a React component that uses lifecycle methods will have to release a new major version since the same code can’t be used on both old and new versions of React. Additionally, people won’t be able to upgrade React until **all** such dependencies upgrade.

About the destructuring question, could you show how a function calling another function can find out if it takes a particular destructured argument? This is necessary to warn on usage of deprecated arguments and would be helpful in saving work for a polyfill.

j-f1:
My [JSPerf test](https://jsperf.com/react-rfc-44/1) says that the destructured form is about **50% slower**.

<sup>Run on Chrome 67.0.3396 / Chrome OS 10575.22.</sup>

wood1986:
Obviously, wrapping the arguments is a bad approach because of @j-f1 results. Anyways, thanks for the test. 

The main purpose of this rfc is trying to solve and eliminate the barrier if the existing method signature needs to be changed. Wrapping the arguments is one of my suggestion at the very beginning and this is failed. How about this?

```
app.js
class A extends React.Component {
  static signature = 16.2;  // we need to specify which version of the signature the stateful component is using
} 
```

```
react.js
if (Component.signature == 16.2) {
  console.warn("v20.0 will remove shouldComponentUpdate(nextProps, nextState) please do the migration");
  sCU(nextProps, nextState);
} else if (Component.signature == 20) {
  sCU(nextState);
} else {
  console.warn("Component.signature is not set. Default is set to '20'");
  sCU(nextState);
}
```

Let me know if it is worth to discuss. If it is not, I will stop the discussion. Otherwise, I will create another rfc.

Thanks for everyone's time for reading and discussing.

PR #46, opened 2018-04-27
shouldComponentHydrate

probablyup:
[View formatted RFC](https://github.com/probablyup/rfcs/blob/7c9d78cb910d0ce199af4059d94b1fc5f25340ae/text/0000-should-component-hydrate.md)

This RFC seeks to add the capability for a component subtree to opt out of rehydration diffing.

### DIFF
--- a/text/0000-should-component-hydrate.md
+++ b/text/0000-should-component-hydrate.md
@@ -1,0 +1,90 @@
+- Start Date: 2018-04-27
+- RFC PR:
+- React Issue:
+
+# Summary
+
+This RFC seeks to add the capability for a component subtree to opt out of rehydration diffing.
+
+# Basic example
+
+I propose adding a new lifecycle method: `shouldComponentHydrate`. Returning `false` from this method should cause React to opt out of reconciliation during rehydration and leave the subtree as-is.
+
+```jsx
+function getComponent() {
+    return (
+        typeof window === 'undefined'
+            ? require('./foo').default
+            : import('./foo').then(mdl => mdl.default)
+    );
+}
+
+class CodeSplittingComponent extends React.Component {
+    constructor() {
+        this.state = {
+            component: getComponent(),
+        };
+
+        if (this.state.component instanceof Promise) {
+            this.state.component.then(ReactClass => this.setState({ component: ReactClass }));
+        }
+    }
+
+    shouldComponentHydrate() {
+        return false;
+    }
+
+    /**
+     * Block re-rendering until the component is resolved and ready.
+     */
+    shouldComponentUpdate(props, state) {
+        return !(state.component instanceof Promise);
+    }
+
+    render() {
+        return (
+            <this.state.component />
+        );
+    }
+}
+```
+
+# Motivation
+
+The basic goal here is to allow something to be rendered by the server and remain untouched by React until the component enqueues a subsequent rerender via prop or state change.
+
+When working in a code splitting context, this becomes very important because what is immediately available on the server may not be on the client. A specific example of a typical SSR + code splitting flow:
+
+1. ReactDOMServer renders a complete page to string and sends it to the browser via some means.
+
+2. React in the browser rehydrates, but one or more component children in the tree are meant to be asynchronously-loaded in the browser context.
+
+3. When the code splitting component is mounted, this temporarily causes the pre-rendered subtree to be discarded since the asynchronously-loaded child has not yet been downloaded.
+
+4. When the child JS downloads, the code splitting component then enqueues a re-render and the child appears again.
+
+Note the UX problem in step 3: pre-rendered HTML ends up being thrown out because there is no current way to opt out of reconciliation during rehydration.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


text/0000-should-component-hydrate.md
```diff
@@ -0,0 +1,90 @@
...
+            component: getComponent(),
+        };
+
+        if (this.state.component instanceof Promise) {
+            this.state.component.then(ReactClass => this.setState({ component: ReactClass }));
+        }
+    }
+
+    shouldComponentHydrate() {
+        return false;
```
streamich: Can it be a static class property instead?

```js
class CodeSplittingComponent extends React.Component {
  static shouldComponentHydrate = false;
  // ...
}
```

alexmog2:
Further motivation for this is content that can't be hydrated at the client at all, e.g. when SSR involves significant work or additional data (such as data lookup, localization or large legacy libraries), or the content itselv is too large and sending a duplicate for hydration is prohibitively expensive.

alexmog2:
Note that the current behavior of `dangerouslySetInnerHTML` is to keep the current content when the content being set is different. In combination with `suppressHydrationWarning` it produces the desired behavior already.

```jsx
<div
    dangerouslySetInnerHTML={{
        __html: ''
    }}
    suppressHydrationWarning
/>
```

See a working example her: https://codesandbox.io/s/zx38ow3z8x

One obvious alternative is to document this behavior and commit to preserving it.

alexmog2:
The next step in the RFC process is

> Build consensus and integrate feedback. RFCs that have broad support are much more likely to make progress than those that don't receive any comments.

It appears that there is a consensus here (of all the people who paid attention so far) that this is a useful performance feature and its implementation is straightforward. What needs to happen for it to become a reality?

tlays:
+1

Can help page load performances & eliminate potential FOUC.
Implementation is straightforward and inline with React lifecycle methods.

All for it

probablyup:
Some guidance from core at this point would be useful as to how such a feature should be implemented. Not sure if it’s standard to tag people though

probablyup:
> Note that the current behavior of dangerouslySetInnerHTML is to keep the current content when the content being set is different. In combination with suppressHydrationWarning it produces the desired behavior already.

```jsx
<div
    dangerouslySetInnerHTML={{
        __html: ''
    }}
    suppressHydrationWarning
/>
```

^ Very interesting, wasn't aware of `suppressHydrationWarning` and the non-replacement behavior. I think that method could work for some specific use cases.

What I'm really looking for here is a first-class way to opt out of rehydration, while still allowing the content to potentially change in the future in future render passes. That's the blend of behavior that works best for an SSR->client side async component workflow without unload/load flashing, etc.

probablyup:
@jciolek actually yes I think that might work.

itssumitrai:
`suppressHydrationWarning` seems to only work for text. Real Ad libraries would most probably modify some DOM nodes, insert images, etc. React would still go ahead and clear out those nodes.

sebmarkbage:
I think the innerHTML thing is probably the way to go for now.

For anyone looking for the inverse problem where you only want to render something on the client, [Partial Hydration](https://github.com/facebook/react/pull/14717) will provide a mechanism for that. The server can render a Suspense boundary with a [permanent fallback in the HTML](https://github.com/facebook/react/pull/14884). In fact, I think that's probably what the synchronous server renderer will do by default if the server suspends a component. That will cause the client to throw it out and render the content client side.

You can kind of do the thing where you leave the HTML in place with partial hydration since it doesn't actually have to hydrate everything to be interactive. You can simply suspend by throwing a Promise that never resolves. The issue is that if any Context above that component changes, we'll throw out the content and trigger the fallback. Because that Context might have changed the HTML.

Ultimately, I think we'll want a way to pass rendered content as some kind of JSON object from the server. In a way that can be passed around and just rendered in the tree. Ideally with some embedded child components.

```js
<div>{serverStuff}</div>
```

If that was server rendered, maybe that data can simply refer to the HTML that was already sent instead of duplicating the HTML.

gaearon:
Thanks for the RFC, and sorry for a very late review (https://github.com/reactjs/rfcs/issues/182).

Quoting the motivation:

> When working in a code splitting context, this becomes very important because what is immediately available on the server may not be on the client. A specific example of a typical SSR + code splitting flow:
> 
> - ReactDOMServer renders a complete page to string and sends it to the browser via some means
> - React in the browser rehydrates, but one or more component children in the tree are meant to be asynchronously-loaded in the browser context.
> - When the code splitting component is mounted, this temporarily causes the pre-rendered subtree to be discarded since the asynchronously-loaded child has not yet been downloaded.
> - When the child JS downloads, the code splitting component then enqueues a re-render and the child appears again.

We have a first-class solution to this now, which you can read about here: https://github.com/reactwg/react-18/discussions/37. In particular, it allows code splitting to seamlessly work with SSR, and partially hydrate the page as the code loads. There is a demo there that you can play with.

There may be other related problems, so there may still be a space for additional related RFCs. But I'll close this one.

PR #47, opened 2018-04-27
New method to extract context value

alexeyraspopov:
[View formatted RFC](https://github.com/alexeyraspopov/rfcs/blob/ContextGetValue/text/0000-context-get-value.md)

### DIFF
--- a/text/0000-context-get-value.md
+++ b/text/0000-context-get-value.md
@@ -1,0 +1,131 @@
+- Start Date: 2018-04-27
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Introduce a method for unwrapping context value in rendering function.
+
+# Basic example
+
+For example, two contexts are defined in an application:
+
+```javascript
+let ThemeContext = React.createContext('light');
+let LanguageContext = React.createContext('en-US');
+```
+
+Each context represents a single value and a pair of `Provider` and `Consumer`
+components to define and get access to this value. Usage details are covered in
+[RFC 0002][new-context-rfc].
+
+Current implementation may lead to the code like this:
+
+```javascript
+function ThemeAndLanguageConsumer({ children }) {
+  return (
+    <LanguageConsumer>
+      {language => (
+        <ThemeConsumer>
+          {theme => children({ language, theme })}
+        </ThemeConsumer>
+      )}
+    </LanguageConsumer>
+  );
+}
+```
+
+Suggested API extension allows to shrink the code above to something like this:
+
+```javascript
+function ThemeAndLanguageConsumer({ children }) {
+  let language = LanguageContext.getValue();
+  let theme = ThemeContext.getValue();
+  return children({ language, theme });
+}
+```
+
+# Motivation
+
+While render prop pattern is extremely useful due to the way how it separates
+concerns, it affects code complexity by adding nesting levels (basically, each
+new context used in a component means +2 levels of indent for its children).
+This introduces following concerns:
+
+- Poor readability ([even for React Core Team members][acdlite-tweet])
+- Each render method call creates new children functions (since they are inline)
+
+This RFC does not address these concerns for all possible pattern applications.
+It reduces related complexity for the components provided by React itself.
+
+# Detailed design
+
+Introduces a new method of `Context` entity:
+
+```javascript
+type Context<T> = {
+  // existing properties
+  Provider: Provider<T>,
+  Consumer: Consumer<T>,
+  // the suggested method
+  getValue: () => T,
+};
+```
+
+The method can be used in the same places where `Context.Consumer` assumed to
+be used.
+
+The upcoming React Suspense feature introduces [effects handling][effects] and
+aims at solving async data rendering problems. Despite on being synchronous by
+its nature, from the implementation perspective, context unwrapping can be
+represented as an effect that is handled in the way that wraps target component
+with necessary context consumers so the values become available during
+rendering phase. In this case, effects are thrown only once during initial
+rendering phase (for mounted component). Subsequent changes of context value
+should be covered by the fact the target component wrapped by consumers that
+combine their values in a single cache.
+
+# Drawbacks
+
+Since the new method brings yet another way for accessing context value (along
+with `Context.Consumer`) there is always a problem to choose the right option.
+It usually leads to subjective decisions based on developers experience and
+preferences.
+
+Talking about implementation in user space, there is a package that solves
+nesting problem by providing composition helpers (using render prop pattern or
+higher-order component):
+https://www.npmjs.com/package/react-compose-context-consumers
+
+# Alternatives
+
+Alternative design once has been suggested by Andrew Clark in Twitter:
+https://twitter.com/acdlite/status/971598256454098944
+
+The idea was to use higher-order components in favor of omitting render props,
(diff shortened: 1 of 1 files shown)


TrySound:
Except the problem [you described here](https://github.com/reactjs/rfcs/pull/47/files#diff-2b61a64cf8b74c000cc5b06c2717dbcaR124) (which is what new context api was created for) you are missing another big problem with your proposal: context is not global variable, you should be able to inject its value with provider for different subtrees.

text/0000-context-get-value.md
```diff
@@ -0,0 +1,131 @@
...
+
+# How we teach this
+
+Since suggested method has similar background to how React Suspense handles
+promises (going from callback style to pull-based code), the same terminology
+and flow can be used for describing the concept.
+
+# Unresolved questions
+
+- How React should propagate context value changes to consumers that are using
```
streamich: This is the main question. From what I understand context is useful because:

1. It re-renders the subtree when its value changes.
2. It preserves parent-child relationship.

The proposed API does not provide both of those features. In this case it can very simply be implemented in user-space:

```jsx
let lang, theme;

const getLanguage() => lang;
const getTheme() => theme;

function ThemeAndLanguageConsumer({ children }) {
  let language = getLanguage();
  let theme = getTheme();
  return children({ language, theme });
}
```

gaearon:
We have some future API ideas that address this problem (in addition to others). We're not ready to share them yet but just leaving a note that we've looked into this.

streamich:
You can have a user-space solution like this:

```js
function ThemeAndLanguageConsumer({ children }) {
  return
    userspaceSolution(LanguageConsumer, ThemeConsumer, (language, theme) =>
      children({ language, theme }));
}
```

ackvf:
I also made one example: https://react-experiments.herokuapp.com/
[code](https://github.com/ackvf/react-experiments/blob/master/src/experiments/ContextAPI/superContext.js)
Used like this
```

<Provide
  theme={[themeContext, this.state.theme]}
  color={[colorContext, this.state.color]}
>
    <Consume
      theme={themeContext}
      color={colorContext}
    >
      {
        ({color, theme}) => <Box color={color} theme={theme}/>
      }
    </Consume>
</Provide>
```

Really simple
```
export const Provide = ({children, ...providers}) => {
  const pKeys = Object.keys(providers)
  return pKeys.reduceRight((nested, key) => {
    const [{Provider}, value] = providers[key]
    return <Provider value={value}>{nested}</Provider>
  },
  children)
}

export const Consume = ({children, ...consumers}) => {
  const pKeys = Object.keys(consumers)
  let lastKey, values = {} // collect values
  return (pKeys.reduceRight((nested, key, ix) => {
    const {Consumer} = consumers[key]
    return prop => {
      if (lastKey) values = {...values, [lastKey]: prop}
      lastKey = pKeys[ix]
      return <Consumer>{nested}</Consumer>
    }
  }, prop => children({...values, [lastKey]:prop}))
  )(values) // after everything is wrapped, call the outer-most function as a child with collected values
}
```

gaearon:
I think it’s fine to keep open if you find the discussion valuable. 

Our solution is similar in spirit but it’s significantly broader in scope so it will take more time to bake.

ackvf:
BTW: Is this a hidden feature or it should not be used like this?
```js
const MyComponent = () => JSON.stringify(themeContext._currentValue)
```

streamich:
@ackvf Well, if it does not have `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED__` prefix, it is not hidden.

ackvf:
@streamich That's what I thought. Maybe I should have asked right away - why it is not documented or even suggested here to solve @alexeyraspopov problem?

TrySound:
Guys with all your suggestions what would you like to see here rendered?
```js
function Language() {
  return LanguageContext._currentValue;
}

<>
  <LanguageContext.Provider value='en'>
    <Language />
  </LanguageContext.Provider>
  <LanguageContext.Provider value='fr'>
    <Language />
  </LanguageContext.Provider>
</>
```

ackvf:
@TrySound Ignore the implications. Of course it breaks about everything, but if a user knows what he's doing, why it is not documented? Like
```
class MyClass extends React.Component {

  componentDidMount() {
    setTimeout(this.forceUpdate, 100);
  }

  render() {
    const val1 = SomeContextA._currentValue
    const val2 = SomeContextB._currentValue
    const val3 = SomeContextC._currentValue
    const val4 = SomeContextD._currentValue

    return (...)
  }
}
```

TrySound:
@ackvf It's obviously not public api. And I'm not sure anybody will know what he is doing here. I'm asking seriously what is the result in my example? It's important case which shouldn't be broken.

ackvf:
The docs say
> When React renders a context Consumer, it will read the current context value from the closest matching Provider above it in the tree.

Since nothing regarding `._currentValue`, one could assume that it could behave in a similar way like if the docs said this:

> When the `_currentValue` of a context Consumer is accessed, it will read the current context value from the closest matching Provider above it in the tree.

Guess what, it really indeed returns `en` and `fr` as expected.

Moreover, given the old implementation of context api, I would actually not expect to access a global variable like you suggest, but instead indeed to access a value "somehow magically" available in a given tree.

probablyup:
What if this was an instance method on the generated Consumer component?

Like you could use it as a render prop like normal, or you could call `Context.Consumer.getValue()` imperatively as needed.

alexeyraspopov:
This RFC does not enforce to read from some global value. Also, no one is going to rely on `_currentValue` since it's not a documented piece of API.

alexeyraspopov:
I don't think it is a solution. The value may not be available at the moment the render function is called and it doesn't allow to make a subscription for subsequent value updates. Also, `_` in the name emphasizes that it is a not public implementation detail.

theKashey:
Just today I've tried to get value from context in `constructor`, in `willMount`, or anywhere before the render.
As result have to split one component in two, first passing context values as props to the second. Looking and working great, but every `hop` like this - wasted CPU time :(

ackvf:
@theKashey You could use the property `_currentValue` to avoid wrappers. [It does work...](https://github.com/reactjs/rfcs/pull/47#issuecomment-386047558) _(At your own risk.)_

TrySound:
@theKashey You can safely use this way until https://github.com/facebook/react/pull/13139 will be landed
```js
class Component extends React.Component {
  value = null
  componentDidMount() {
    this.value.method()
  }

  render() {
    return <Context.Consumer>
      {value => {
        this.value = value;
        return (
          <div></div>
        )
      }}
    <Context.Consumer>
  }
}
```

theKashey:
I am ok with component splitting. It's much easier to test them.

gaearon:
I don't recommend anyone to use `_currentValue`, stuff like this changes between patch releases. In fact we've already changed it once which broke a library that decided to rely on this.

We *are* experimenting with an API like the one suggested in this RFC. Note it won't work in commit phase lifecycles — only in the render phase. The distinction is confusing enough at the moment that we're going to keep the API unstable until we solve some other issues.

alexeyraspopov:
I think this PR can now be just closed, given the fact that hooks are covering the usage of context in the way that was described here. I'm so excited about the opportunity to try it.

Thanks React team for all their effort! ❤️

gaearon:
👍 Thanks for writing up the proposal!

PR #51, opened 2018-05-22
Profiler RFC

bvaughn:
### [View formatted RFC](https://github.com/bvaughn/rfcs/tree/profiler/text/0000-profiler.md)

---

Feedback is welcome!
* Would you use these timings?
* If not (or if so) is there something else we could add to make it more useful to you?
* Do you have ideas for abstractions (or visualizations) we could build on top of this to make it more useful and/or beginner friendly?

#### Related PRs
* facebook/react/pull/12745: Add Profiler component and tests
* facebook/react/pull/12852: Add start and stop time params
* facebook/react/pull/12886: Add production + profiling bundle for ReactDOM and ReactNative

### DIFF
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -69,14 +69,14 @@ Timing measurements should also be significantly lighter weight than the current
 The `onRender` callback is called each time a component within the `Profiler` renders. It receives the following parameters:
 ```js
 function onRenderCallback(
-  id,
-  phase,
-  actualTime,
-  baseTime,
-  startTime,
-  commitTime
-) {
-  // Aggregate or log render timings
+  id: string,
+  phaseL "mount" | "update",
+  actualTime: number,
+  baseTime: number,
+  startTime: number,
+  commitTime: number
+): void {
+  // Aggregate or log render timings...
 }
 
 ```
@@ -100,13 +100,6 @@ Start time isn't just the commit time less the "actual" time, because in async r
 
 #### `commitTime: number`
 Commit time could be roughly determined using e.g. `performance.now()` within the `onRender` callback, but multiple `Profiler` components would end up with slightly different times for a single commit. Instead, an explicit time is provided (shared between all `Profiler`s in the commit) enabling them to be grouped if desirable.
-
----
-
-Also see the following implementation PRs:
-* [facebook/react/pull/12745](https://github.com/facebook/react/pull/12745): Add `Profiler` component and tests
-* [facebook/react/pull/12852](https://github.com/facebook/react/pull/12852): Add start and stop time params
-* [facebook/react/pull/12886](https://github.com/facebook/react/pull/12886): Add production + profiling bundle for ReactDOM and ReactNative
 
 # Drawbacks
 
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -126,4 +126,6 @@ A reactjs.org blog post would be a good initial start.
 
 # Unresolved questions
 
-How will people use this/
+How will people use this? The current proposal is kind of low level and would probably benefit from some reusable abstractions being built on top of it that e.g. aggregate/batch render timings.
+
+How will this feature integrate with React DevTools? I have some ideas but nothing concrete yet to share.
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -71,8 +71,8 @@ The `onRender` callback is called each time a component within the `Profiler` re
 function onRenderCallback(
   id: string,
   phase: "mount" | "update",
-  actualTime: number,
-  baseTime: number,
+  actualDuration: number,
+  baseDuration: number,
   startTime: number,
   commitTime: number
 ): void {
@@ -87,12 +87,12 @@ The `id` value of the `Profiler` tag that was measured. This value can change be
 #### `phase: "mount" | "update"`
 Either the string "mount" or "update" (depending on whether this root was newly mounted or has just been updated).
 
-#### `actualTime: number`
+#### `actualDuration: number`
 Time spent rendering the `Profiler` and its descendants for the current (most recent recent) update. This time tells us how well the subtree makes use of `shouldComponentUpdate` for memoization.
 
 Ideally, this time should decrease significantly after the initial mount. Althoguh in async mode, under certain conditions, React might render the same component more than once as part of a single commit. (In this event, the "actual" time for an update might be larger than the initial time.)
 
-#### `baseTime: number`
+#### `baseDuration: number`
 Duration of the most recent `render` time for each individual component within the `Profiler` tree. This reflects a worst-case cost of rendering (e.g. the initial mount or no `shouldComponentUpdate` memoization).
 
 #### `startTime: number`
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -70,7 +70,7 @@ The `onRender` callback is called each time a component within the `Profiler` re
 ```js
 function onRenderCallback(
   id: string,
-  phaseL "mount" | "update",
+  phase: "mount" | "update",
   actualTime: number,
   baseTime: number,
   startTime: number,
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -8,7 +8,7 @@ New React profiling component that collects timing information in order to measu
 
 This component will also integrate with the [experimental `interaction-tracking` API](https://github.com/facebook/react/pull/13234) so that tracked interactions can be correlated with the render(s) they cause. This enables the calculation of "wall time" (elapsed real time) from when e.g. a user clicks a form button until when the DOM is updated in response. It also enables long-running renders to be more easily attributed and reproduced.
 
-Note that an experimental release of the profiler component is available in version 16.4 as `React.unstable_Profiler`. It does not yet support interactions as that package has not been released.
+Note that an experimental release of the `Profiler` component was first included with version 16.4 as `React.unstable_Profiler`. It did not yet support interaction tracing as that package had not been released.
 
 # Usage example
 
@@ -58,11 +58,11 @@ render(
 );
 ```
 
-Although `Profiler` is a light-weight component, it should be used only when necessary; each use adds CPU and memory overhead to an application.
+Although `Profiler` is a light-weight component, it should be used only when necessary; each use adds some CPU and memory overhead to an application.
 
 # Motivation
 
-It is important that render timing metrics work properly with React's experimental async rendering mode. When asynchronously rendering, React may yield periodically so that an app remains responsive even on low-powered devices. This yielded time (when React is not running) should not be included when considering the "cost" of a render. This distinction is not possible to implement in user space.
+It is important for render timing metrics to work properly with React's experimental concurrent rendering mode. In concurrent mode, React may yield to the browser periodically so that an app remains responsive even on low-powered devices. This yielded time (when React is not running) should not be included when considering the "cost" of a render. This distinction is not possible to implement in user space.
 
 Timing measurements should also be significantly lighter weight than the current User Timing API so that they can be gathered in production without negatively impacting user experience. (The User Timing API is currently disabled for production because it is slow.) In addition to a faster implementation, we can further limit the impact on existing apps by creating a new production + profiling bundle. This way, apps that don't make use of the `Profiler` component (or wish to disable it globally) will not incur any additional overhead. (The `Profiler` component will render its children in production mode but its `onRender` callback will not be called.)
 
@@ -77,7 +77,7 @@ function onRenderCallback(
   baseDuration: number,
   startTime: number,
   commitTime: number,
-  interactions: Array<{ name: string, timestamp: number }>,
+  interactions: Set<{ name: string, timestamp: number }>,
 ): void {
   // Aggregate or log render timings...
 }
@@ -91,14 +91,14 @@ The `id` value of the `Profiler` tag that was measured. This value can change be
 Identifies whether this component has just been mounted or re-rendered due to a change in `state` or `props`.
 
 #### `actualDuration: number`
-Time spent rendering the `Profiler` and its descendants for the current (most recent recent) update. This time tells us how well the subtree makes use of `shouldComponentUpdate` for memoization.
+Time spent rendering the `Profiler` and its descendants for the current (most recent recent) update. This time tells us how well the subtree makes use of memoization (e.g. `React.memo`, `useMemo`, `shouldComponentUpdate`).
 
 Ideally, this time should decrease significantly after the initial mount as many of the descendants will only need to re-render if their specific `props` change.
 
 Note that in async mode, under certain conditions, React might render the same component more than once as part of a single commit. (In this event, the "actual" time for an update might be larger than the initial time.)
 
 #### `baseDuration: number`
-Duration of the most recent `render` time for each individual component within the `Profiler` tree. In other words, this value will only change when a component is re-rendered. It reflects a worst-case cost of rendering (e.g. the initial mount or no `shouldComponentUpdate` memoization).
+Duration of the most recent `render` time for each individual component within the `Profiler` tree. In other words, this value will only change when a component is re-rendered. It reflects a worst-case cost of rendering (e.g. the initial mount or a tree with no memoization).
 
 #### `startTime: number`
 Start time identifies when a particular commit started rendering. Although insufficient to determine the cause of the render, it can at least be used to rule out certain interactions (e.g. mouse click, Flux action). This may be helpful if you are also collecting other types of interactions and trying to correlate them with renders.
@@ -109,9 +109,9 @@ Start time isn't just the commit time less the "actual" time, because in async r
 Commit time could be roughly determined using e.g. `performance.now()` within the `onRender` callback, but multiple `Profiler` components would end up with slightly different times for a single commit. Instead, an explicit time is provided (shared between all `Profiler`s in the commit) enabling them to be grouped if desirable.
 
 #### `interactions: Array<{ name: string, timestamp: number }>`
-An array of interactions that were being tracked (via the `interaction-tracking` package) when this commit was initially scheduled (e.g. when `render` or `setState` were called).
+Set of interactions that were being tracked (via the `interaction-tracking` package) when this commit was initially scheduled (e.g. when `render` or `setState` were called).
 
-In the event of a cascading render (e.g. an update scheduled from `componentDidMount` or `componentDidUpdate`) React will forward these interactions along to the subsequent `onRender` calls.
+In the event of a cascading render (e.g. an update scheduled from an effect or `componentDidMount`/`componentDidUpdate`) React will forward these interactions along to the subsequent `onRender` calls.
 
 # Drawbacks
 
@@ -135,3 +135,7 @@ Perhaps we could provide some sort of discoverability within React DevTools.
 
 * [facebook/react/pull/13253](https://github.com/facebook/react/pull/13253): Integration with the proposed `interaction-tracking` package
 * [facebook/react-devtools/pull/1069](https://github.com/facebook/react-devtools/pull/1069): Integration with React DevTools
+
+# Related documentation
+* [fb.me/react-profiling](http://fb.me/react-profiling): Instructions for profiling in production
+* [fb.me/react-interaction-tracing](http://fb.me/react-interaction-tracing): Instructions for using interaction tracing with React
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -60,7 +60,7 @@ Although `Profiler` is a light-weight component, it should be used only when nec
 
 # Motivation
 
-It is important that render timing metrics work properly with React's experimental async rendering mode. When asynchronously rendering, React may yield periodically so that an app remains responsive even on low-powered devices. This yielded time (when React is not running) should not be included when considering the "cost" of a render. This distinction is not possible to implement this in user space.
+It is important that render timing metrics work properly with React's experimental async rendering mode. When asynchronously rendering, React may yield periodically so that an app remains responsive even on low-powered devices. This yielded time (when React is not running) should not be included when considering the "cost" of a render. This distinction is not possible to implement in user space.
 
 Timing measurements should also be significantly lighter weight than the current User Timing API so that they can be gathered in production without negatively impacting user experience. (The User Timing API is currently disabled for production because it is slow.) In addition to a faster implementation, we can further limit the impact on existing apps by creating a new production + profiling bundle. This way, apps that don't make use of the `Profiler` component (or wish to disable it globally) will not incur any additional overhead. (The `Profiler` component will render its children in production mode but its `onRender` callback will not be called.)
 
--- a/text/0000-profiler.md
+++ b/text/0000-profiler.md
@@ -117,8 +117,10 @@ This is an entirely new component. Adoption can be organic and gradual.
 
 A reactjs.org blog post would be a good initial start.
 
+Perhaps we could provide some sort of discoverability within React DevTools.
+
 # Unresolved questions
 
-How will people use this? The current proposal is kind of low level and would probably benefit from some reusable abstractions being built on top of it that e.g. aggregate/batch render timings.
-
-How will this feature integrate with React DevTools? I have some ideas but nothing concrete yet to share.
+* How will people use this? The current proposal is kind of low level and would probably benefit from some reusable abstractions being built on top of it that e.g. aggregate/batch render timings.
+* Are the proposed timing metrics sufficiently useful?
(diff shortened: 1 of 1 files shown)


fbartho:
I certainly would use these timings / this component in my team's app. Identifying which "screen" suffers the most would be very valuable.

Is there a model in this for profiling `"time until interactive" - networkTime` -- This way we can see the fully loaded data, and optimize the React Render tree directly?

Also valuable: if it could automatically bubble up which sub-component is most at fault, or the number of re-renders of subcomponents. This way we can identify data misuse that leads to repeat renders.

tizmagik:
Would it be possible to detect/report on “useless” or unnecessary updates?

bvaughn:
> Is there a model in this for profiling "time until interactive" - networkTime -- This way we can see the fully loaded data, and optimize the React Render tree directly?

I'm sorry, but I'm not sure I understand what you're suggesting. Could you elaborate?

> Also valuable: if it could automatically bubble up which sub-component is most at fault, or the number of re-renders of subcomponents. This way we can identify data misuse that leads to repeat renders.

I think this might be better handled by DevTools. (I have an idea that I hope to prototype for how the new `Profiler` might integrate with DevTools to power some local visualizations/graphs of an application. More to come soon...)

Something like the number of re-renders seems easier to detect and handle locally, because it's easier to reproduce / more predictable (regardless of system specs). On the other hand, timing is often impacted by hardware capabilities and browser/versions, and may be difficult for devs to get an accurate idea about from their (generally high-end) machines.

Also, timing is probably the most important "cost" of rendering. If a component re-renders unnecessarily, but it's super fast- is it actually that important? On the other hand, if it's slow each time the app renders- regardless of whether the render was "useless" or not, it should probably be looked into.

That being said, I haven't thought a lot about non timing aspects of this. Maybe I'm overlooking something. Let's see what others think about it. 😄

> Would it be possible to detect/report on “useless” or unnecessary updates?

Something like [maicki/why-did-you-update](https://github.com/maicki/why-did-you-update)?

To be honest, I haven't considered this much. Initially, I think it might be tricky, because of e.g. inline functions for event handlers. (These might interact with the DOM even though it was unnecessary. Seems hard for React to detect without adding overhead of something like a per-renderer property whitelist, etc.)

Definitely seems useful. Although similar to the number of renders/re-renders, it seems more predictable / easier to reproduce locally than timing- so maybe this is something that belongs in DEV mode only, rather than a profiling build?

gaearon:
I think the notion of “useless” never made a lot of sense. It was a handy heuristic in some cases but it both failed to identify some “less useful” cases and incorrectly tagged “useful” updates as wasted.

Now, I know a lot of people want to see this heuristic come back in some form. I think what we really want there is a scale rather than a yes/no flag. If a big tree was reconciled and it took a lot of time, but the result was just a couple of DOM changes, it’s still worth looking into, even though it’s not completely “wasted”. Similarly, a fast re-render of a few components that didn’t lead to DOM updates isn’t necessarily useful to act upon if it’s fast.

So I think what we want is a scale of how much the update cost (in time) vs how much it affected the output (host effects, lifecycles). And we’d like to see the updates (as setState calls?) that have the worst ratio on that scale.

I don’t know if it needs to be a part of this RFC. But that’s how I see “wasted” heuristic in the future if we add it again. 

Does that make sense?

paularmstrong:
This looks really great!

> Would you use these timings?

Absolutely yes. We have a timing HOC at Twitter and it's not going to work once React hits v17.0.0 and deprecates some things. (it's rudimentary, but it gets us enough information to alert on critical regressions)

As for this particular RFC:

**`id` attribute:** I'm a bit unclear on this, as the RFC isn't specific...

Does this have to be unique? If I were to use the same `id` on many instances, e.g. in a timeline of Tweets, `<Profiler id="Tweet" onRender={this._handleProfilerRender}><Tweet /></Profiler>`, would `_handleProfilerRender` be called for each, or once per?

The way it looks with the definition for `baseTime` is that I will `_handleProfilerRender` will be called only once with the most recent time it was called.

fbartho:
@bvaughn in reply to your response: yes I totally agree that some of my requests might live better in the DevTools. -- I'm also considering this from the concept of a React-Native platform as my first target, I'm not sure that redirects your comments, but maybe that gives you more context?

In regards to your question re my comment about "time-till-interactive" I'm trying to optimize for performance when many of the components are embedded in GraphQL Query components from Apollo. It's want to have more insight into how long it takes for my UI to "settle" which is a time-diff between componentDidMount and the last re-render after data comes back. I admit I'm a bit confused as to how to best profile & track which parts of my component tree need love.

alvaropinot:
Thanks for sharing, I will definitely use it. 

I can think about the following ideas:

- Add a **generic implementation for the `onRender` method that might provide a mechanism for beginners to see perf results with 0 setup**. Maybe a separate package can do this. This could be the seed of a profile reporter that could be integrated within react dev tools.
- Consider **toggling the component behaviour on/off dependending on a prop**. This might allow disabling it in prod or disabling it when devtools are not open, or within any scenario where the performance impact of profiling might be negative.
- It can be specially interesting for **wrapping components in a performance/TDD approach** where min metrics should be fullfilled to obtain a passing perf test.

Again thanks for sharing, hope those 👆help 😁

miraage:
Opinion: it would be amazing to have an option to toggle `<Profiler>` via React DevTools or a browser extension.

I think it might be useful to avoid using "hacks" (e.g. global variables, global functions) in production.

E.g. you push potential bottlenecks wrapped into a disabled `<Profiler>` (so your site's users see no difference), while you can do your measurements by your own.

bvaughn:
@paularmstrong Thanks!

> `id attribute:` I'm a bit unclear on this, as the RFC isn't specific...
> Does this have to be unique? If I were to use the same id on many instances...would _handleProfilerRender be called for each, or once per render cycle for all of those with the same id?

Callbacks will always be called once per `Profiler` (assuming at least one of that `Profiler` has children that rendered).

The ID is for your benefit only. If you are using a single, centralized logging function- the ID can help identify which `Profiler` (which tree) it is referring to.

---

@fbartho

> yes I totally agree that some of my requests might live better in the DevTools. -- I'm also considering this from the concept of a React-Native platform as my first target, I'm not sure that redirects your comments, but maybe that gives you more context?

React Native support is a central focus for this new component, and will also be a central focus for the DevTools work I'll be doing soon. We realize that RN is currently a bit harder to profile than React DOM.

> In regards to your question re my comment about "time-till-interactive" I'm trying to optimize for performance when many of the components are embedded in GraphQL Query components from Apollo. It's want to have more insight into how long it takes for my UI to "settle" which is a time-diff between componentDidMount and the last re-render after data comes back.

Seems like you should be able to infer this based on when your callback is called? The first time it is called for a tree, the `phase` will be "mount" (so that's your starting point) and then subsequent calls (while it "settles"m as you put it) will have a `phase` of "update". You can use the `commitTime` of these calls to compute your own delta.

---

@alvaropinot

> Add a generic implementation for the onRender method that might provide a mechanism for beginners to see perf results with 0 setup.

This is where React DevTools fits in! Can't wait to show you what Sebastian and I have been brainstorming here. :smile:

> Consider toggling the component behaviour on/off dependending on a prop. This might allow disabling it in prod or disabling it when devtools are not open, or within any scenario where the performance impact of profiling might be negative.

You could do this via a HOC (if you wanted). The way we intend for this to work by default though will be: ON for development and profiling bundles, OFF for production. (It doesn't really matter if it's always on for dev mode, since the perf impact is minimal. You can switch which production bundle of React is used- regular or profiling- to turn it off for production mode.)

---

@miraage

> Opinion: it would be amazing to have an option to toggle <Profiler> via React DevTools or a browser extension.

Yes! This is planned! :grin: Will share something soon.

text/0000-profiler.md
```diff
@@ -0,0 +1,128 @@
...
+The `onRender` callback is called each time a component within the `Profiler` renders. It receives the following parameters:
+```js
+function onRenderCallback(
+  id: string,
+  phase: "mount" | "update",
+  actualTime: number,
+  baseTime: number,
+  startTime: number,
+  commitTime: number
+): void {
```
j-f1: Why not pass all of this as a single object so people can pick out the keys they need?

bvaughn:
@jamesreggio

> Just to confirm, an update to any descendent of a Profiler component (not just its immediate descendants) will trigger the `onRender` callback, correct?

Correct.

> It would be nice to be able to correlate `onRender` callbacks on separate Profiler components that fire during the same rendering transaction. For example, if a Redux dispatch causes an update in two disjoint subtrees, each with a separate Profiler, I'd like to be able to correlate the two onRender callbacks to the dispatch.

The `commitTime` parameter can be used for this.

> How does this behave if a subtree that contains a Profiler gets jettisoned during the dispatching of an exception?

In the scenario you describe, since the components underneath of the `Timeout` component would not be committed, the `Profiler` inside of it would not be called. You could measure the "cost" of this by wrapping a `Profiler` around the `Timeout` component though, I believe.

text/0000-profiler.md
bvaughn: Reasonable question!

I'd say the reasons are that I'm following precedent (we don't pass named parameters anywhere else that I can think of off the top of my head) and avoiding allocating a wrapper Object during commit.

I'd be interested to hear what others think about this aspect.

jamesreggio: I'd vote for an object, despite the fact that it breaks with existing React API precedent.

The order of these timing arguments is going to be tough to memorize, and I can imagine only being interested in a subset of them. Using an object also enables you to add additional timing data down the road.

I'd view it as somewhat analogous to an event object, which has a variety of keys, only some of which are of interest for any given listener.

gaearon: Why would you need to memorize it? I imagine you'd only use Profiler in a few places in the app, and each time could consult the docs.

The need to avoid allocations is pretty important because adding GC pressure can skew the profiling results.

bvaughn: Even if you use `Profiler` in more than one place, the callback you pass it is likely shared- (this is why the `id` parameter exists)- so you would only need to write these params (in the correct order) in a single place.

tsiq-swyx:
```function onRenderCallback(
  id: string,
  phase: "mount" | "update",
  actualTime: number,
  baseTime: number,
  startTime: number,
  commitTime: number
): void {

Comment (jamesreggio) [reply text/0000-profiler.md]: "Just consult the docs each time" could be used to justify any number of developer-unfriendly interfaces, and a function with the signature of `string, string, number, number, number, number` is pretty unfriendly.

I'm skeptical that allocating one object per profiler, per commit will make a difference, given that every commit implies a mount or update which, at minimum, will include plenty of small object allocations within `React.createElement`.

There are two ways I envision myself using this API. The first is for well-planned performance regression monitoring, where I thoughtfully chose a couple places to instrument and I wire them into a central system for reporting the data. In that scenario, an obtuse API is tolerable because the integration happens once.

The other way I expect I'll use this API is for spot-profiling, where I'm either trying to reduce the performance cost of a new feature, or I'm probing around the codebase to isolate problematic hierarchies. The easier it is to just drop a component in for spot-checking, the more I'll do it.

Comment (bvaughn) [reply text/0000-profiler.md]: Yup, this concern makes sense. And I agree that we wouldn't be allocating too many new object for this, because it would only be one per `Profiler` per commit. I was just sharing rationale for why it is currently the way it is.

File: text/0000-profiler.md [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+    </Profiler>
+  </App>
+);
+```
+
+Although `Profiler` is a light-weight component, it should be used only when necessary; each use adds CPU and memory overhead to an application.
+
+# Motivation
+
+It is important that render timing metrics work properly with React's experimental async rendering mode. When asynchronously rendering, React may yield periodically so that an app remains responsive even on low-powered devices. This yielded time (when React is not running) should not be included when considering the "cost" of a render. This distinction is not possible to implement this in user space.
```
fstoerkle: typo: "This distinction is not possible to implement ~this~ in user space."

text/0000-profiler.md
alvaropinot: Just my personal opinion here, but an object looks good from my point of view as long functions with more that 2/3 args are always hard to remember, you tend to start adding `null` for the values that you might not want to use, I'm looking at you `JSON.stringify(foo, null, 2)` 😅 , you also need to remember the order and it's harder to refactor as you impact anyone already using that order.

Plus with the actual syntax for destructuring the function signature looks pretty much the same but with curly braces 😁, the best of two worlds!

```js
onRenderCallback({ id, phase, actualTime, baseTime, startTime, commitTime })
```
vs

```js
onRenderCallback(id, phase, actualTime, baseTime, startTime, commitTime)
```

bvaughn:
Just an FYI, but the 16.4.1 release that went out yesterday includes a profiling+production build with support for the experimental profiler component– if you're curious in playing around with production timings: https://unpkg.com/react-dom@16.4.1/cjs/react-dom.profiling.min.js

text/0000-profiler.md
jamesreggio: Now that the Profiling API is out in the 16.4.1 release, I assume you decided to take no action on this?

bvaughn: The `unstable_Profiler` component was introduced in 16.4.0. The only thing that's new in 16.4.1 is a production+profiling build.

Unstable APIs can change. We haven't decided one way or another. This is kind of an open thread for discussion.

ljharb:
@bvaughn a new thing was added in a patch release?

bvaughn:
> @bvaughn a new thing was added in a patch release?

Not really. The profiling+production bundle is literally not connected to anything. The only way to load it would be import directly from "react-dom/cjs/react-dom.profiling.min.js" (or maybe use a Webpack alias to re-map imports).

bvaughn:
I'm not advocating that `Profiler` be used this way, but as an example– just to play around with it, I've plugged it into the documentation site for `react-window` in order to measure the amount of time each of the demo components spends rendering.

https://react-window.now.sh/

text/0000-profiler.md
bvaughn: Just circling back on this particular thread. Sebastian and I chatted about this yesterday, and we've decided to avoid named parameters because the overhead of the wrapper objects (however small each individual one is) will add up in larger applications.

jpnelson:
@bvaughn Nice work! I came up with a pretty similar proposal for a slightly different use case: https://github.com/jpnelson/react-component-timing/blob/master/README.md

The primary differences between what I would like and what is in this RFC are:

1) I would like to care a little bit about the definition of a "meaningful" state for the app, by examining props and determining the "loading" state (similar to https://github.com/reactjs/rfcs/pull/51#issuecomment-391225566)
2) I would like to have this data in runtime in production, to be a metric used for tools like new relic, google analytics and the like

So, my questions are

1) Do you imagine this as a low level component so that what I would want could be built on top of this? I would think this is the best approach, I don't think the Profiler should worry as much about the semantics of meaningfulness, only low level rendering times that user space libraries couldn't.

2) Would it be bad (or even possible?) to use this in a production environment, to report to third party tools? I am worried about the performance overhead. This could be fixed somewhat by sampling rates, but it's still not great. You mention that the User timing API has some overhead, I'm yet to test it but I'd be concerned if it slowed things down considerably for real users. If it does, I'd have to investigate cheaper ways of measuring timing of things.

I'm yet to try out `unstable_Profiler` so maybe some of my questions will be answered by doing that, but just a heads up about what I'm thinking.

bvaughn:
> 1. Do you imagine this as a low level component so that what I would want could be built on top of this? I would think this is the best approach, I don't think the Profiler should worry as much about the semantics of meaningfulness, only low level rendering times that user space libraries couldn't.

I'm not entirely sure what you have in mind, but yes you could build abstractions on top of the `Profiler` API.

Maybe you could create a proof of concept to see if the proposed API is adequate?

> 2. Would it be bad (or even possible?) to use this in a production environment, to report to third party tools? I am worried about the performance overhead. This could be fixed somewhat by sampling rates, but it's still not great. You mention that the User timing API has some overhead, I'm yet to test it but I'd be concerned if it slowed things down considerably for real users. If it does, I'd have to investigate cheaper ways of measuring timing of things.

The description mentions some related PRs, including facebook/react/pull/12886 which adds a production + profiling bundle for ReactDOM and ReactNative renderers. This build even shipped with [react-dom@16.4.1](https://unpkg.com/react-dom@16.4.1/cjs/) and [react-native@0.56.0](https://unpkg.com/react-native@0.56.0/Libraries/Renderer/oss/) if you'd like to play around with it. (You could use e.g. Webpack alias to remap "react-dom" imports to "react-dom/cjs/react-dom.profiling.min.js".)

The overhead added by calling `performance.now()` and accumulating times should be minimal, but it _is_ still added overhead.

I also think you're right that you would want to use some kind of sampling or aggregation when saving this data for real users. You wouldn't want to try saving all timings. 😄

bvaughn:
Hello everyone! 👋 It's been a few weeks since I've given an update on this RFC, but I just pushed a few changes that tie in a new proposed [`interaction-tracking` package](https://github.com/facebook/react/pull/13234) and how I envision it [integrating with the `Profiler` API](https://github.com/facebook/react/pull/13253).

Related PRs for anyone interested in digging in:
* `interaction-tracking` proposal: https://github.com/facebook/react/pull/13234
* `Profiler` / `interaction-tracking` integration: https://github.com/facebook/react/pull/13253

I'd like to also share that I've made progress on the DevTools profiler plug-in which you can see an [early version of here](https://www.youtube.com/watch?v=vfm4018YBEY). In case anyone would like to dig in deeper with that, here's the PR:
* DevTools Profiler plug-in: https://github.com/facebook/react-devtools/pull/1069

jrnail23:
I just wanted to chime in and say that one of my key points for React performance tools is that some of our key performance optimization work targets somewhat older browser versions, as some of our most important customers have IT-controlled workstations that tend to lag behind quite a bit.

In particular, we have performance issues specific to Firefox v52 (yep, mandated by one of those big companies), but since FF52 doesn't support `PerformanceObserver`, profiling has proven to be a real pain.

So I guess I'm generally advocating for fallback support or customizable measurement mechanisms instead of relying exclusively on newer (non-universal) browser APIs.  This concern may apply more directly to some of your other performance/profiling features, but this looked like as good a place as any to raise it. 😄

bvaughn:
> So I guess I'm generally advocating for fallback support or customizable measurement mechanisms instead of relying exclusively on newer (non-universal) browser APIs.

I don't understand what you mean with regard to this proposal. The profiler API does not use the `PerformanceObserver` API. It uses `performance.now()` _if available_ and falls back to `Date.now()` otherwise– just as React's scheduler does.

jrnail23:
@bvaughn, if that's the case, then great!  I may be getting my tools mixed up here -- sorry for any misunderstanding.

bvaughn:
No worries 😄

lixiong-intelex:
Hi @bvaughn , a dump question, will I be able to use the new Profile feature after I upgrade React to 16.5.0 and React Dev Tools to 3.3.2?

bvaughn:
Yes, the profiler plugin will work with 16.5– assuming you're either running in dev mode or using the profiling build (e.g. `import ReactDOM from 'react-dom/profiling'`)

jrnail23:
@lixiong-intelex, I also had to update react-dom to v16.5.0 to get the profiling tab to show.

lixiong-intelex:
@jrnail23 that's what I found out as well. Thanks!

bvaughn:
Ah, I see the confusion. `react` and `react-dom` should always use the same versions. When you asked about upgrading "react" I assumed you meant both. :)

sw-yx:
is this rfc settled?

sw-yx:
was just browsing rfc's and saw this was still open thats all. thought it was stale. congrats on shipping!

lippyDesign:
Can someone explain to me how wrap works with async calls? I'm trying to measure how long it takes to make api request and render:
`
import React, { unstable_Profiler as Profiler } from 'react';
import { unstable_trace as trace, unstable_wrap as wrap } from "scheduler/tracing";
import axios from 'axios';

import List from './List';

  state = {
    count: 0,
    data: null
  };

  componentDidMount() {
    this.fetchData();
  }

  logProfile = (id, phase, actualTime, baseTime, startTime, commitTime, interactions) => {
    console.log('--------- logProfile fired -----------')
    console.log(`${id}'s ${phase.toUpperCase()} phase:`);
    console.log(`Actual time: ${actualTime} ms`);
    console.log(`Base time: ${baseTime} ms`);
    console.log(`Start time (since component mounted): ${startTime} ms`);
    console.log(`Commit time (since component mounted): ${commitTime} ms`);
    console.log(interactions);
  };

  go = direction => () => this.setState(({ count }) => ({
    count: direction === "up" ? count + 1 : count - 1
  }));

  fetchData = () => {
    trace("Fetch todos", performance.now(), () => {
      this.setState({ data: null });
      wrap(async () => {
        try {
          const { data } = await axios.get('https://jsonplaceholder.typicode.com/todos/');
          this.setState({ data });
        } catch (e) {
          console.log(e);
        }
      })()
    });
  }

  render() {
    return (
      <Profiler id="listPage" onRender={this.logProfile}>
        <button onClick={this.go("up")}>up</button>
        <div>The count is {this.state.count}</div>
        <button onClick={this.go("down")}>down</button>
        <hr />
        <button onClick={() => this.fetchData()}>refetch</button>
        <hr />
        <List data={this.state.data} />
      </Profiler>
    );
  }
}

export default App;
`

text/0000-profiler.md
```diff
@@ -0,0 +1,137 @@
...
+  commitTime: number,
+  interactions: Array<{ name: string, timestamp: number }>,
+): void {
+  // Aggregate or log render timings...
+}
+
+```
+
+#### `id: string`
+The `id` value of the `Profiler` tag that was measured. This value can change between renders if e.g. it is derived from `state` or `props`.
```
eps1lon: ~I'm a bit confused about that part. Why would the `id` change between renders? It sounds like the second part was meant for an additional parameter e.g. something like `reason`?~
Was a bit confused about this part. It sounded like the `id` might include some metadata why `onRender` was called but what this refers to is the value of the `id` prop e.g. `<Profiler id={'tweets-' + props.userId} onRender={handleRenderCallBackWithChangingIds} />`

eps1lon:
@bvaughn What are the plans for this component regarding `react-test-renderer`? As it stands right now `onRender` is called neither in the base renderer nor with the shallow renderer. Created a little [playground for it](https://codesandbox.io/s/jxjoj2p0v) (can ignore the `enzyme` related stuff).

I wanted to experiment with the Profiler a bit and see if I can use this to create some performance regression test that work with a performance budget. While I'm not sure if this makes sense with `shallow` the base renderer should probably work. Note that the `shallow` renderer explicitly calls `render` but `onRender` is never called and `shallow` even throws when the `Profiler` is the root component.

The timings probably don't make much sense with the base or `shallow` renderer. At least for counting the render cycles it might add some value.

text/0000-profiler.md
```diff
@@ -0,0 +1,137 @@
...
+The `onRender` callback is called each time a component within the `Profiler` renders. It receives the following parameters:
+```js
+function onRenderCallback(
+  id: string,
+  phase: "mount" | "update",
+  actualDuration: number,
+  baseDuration: number,
+  startTime: number,
+  commitTime: number,
+  interactions: Array<{ name: string, timestamp: number }>,
```
eps1lon: Shape changed slightly and matches in 16.8.4 (guess this originates in the `scheduler`?)
```suggestion
  interactions: Set<{ id: number; name: string, timestamp: number }>,
```

Will the implementation change or is the proposal outdated?

bvaughn: Ah, yeah. This RFC is a bit outdated.

bvaughn:
> @bvaughn What are the plans for this component regarding `react-test-renderer`? As it stands right now `onRender` is called neither in the base renderer nor with the shallow renderer.

The test renderer is a reconciler-based renderer, so it could support profiling and the `onRender` callback like any other renderer. (In fact, [the `Profiler` tests use test renderer](https://github.com/facebook/react/blob/master/packages/react/src/__tests__/ReactProfiler-test.internal.js).) Whether it does is just a question of whether [the `enableProfilerTimer` feature flag](https://github.com/facebook/react/blob/7ad7386308abd55f12ad9f1cddac6a97860e92eb/packages/shared/ReactFeatureFlags.js#L29) is enabled, and [it isn't currently for the test renderer](https://github.com/facebook/react/blob/7ad7386308abd55f12ad9f1cddac6a97860e92eb/packages/shared/forks/ReactFeatureFlags.test-renderer.js#L20).

When `Profiler` is released as a stable API (probably soon, I assume) we'll remove the feature flag, and it will work in the test renderer as we..

bvaughn:
FYI this RFC has entered the final comment period. I plan to merge it this week.

God-Universe (commented):
Gh

PR #53, opened 2018-06-13
Allow null as a value on controlled components

RyanNerd:
See: https://github.com/facebook/react/issues/11417
See: https://github.com/facebook/react/issues/5013#issuecomment-340898727


text/0000-allow-null-on-controlled-components.md
```diff
@@ -0,0 +1,111 @@
...
+When implemented this feature change would be documented on the [React Blog](https://reactjs.org/blog).
+
+# Unresolved questions
+
+If the design choice is made where `value={null}` is always treated as string:
+* What about the situation where the controlled component is expecting a number or date?
+  Always treating `value={null}` as a string could be problematic.
+
+If the design choice is `value` for controlled and `defaultValue` for uncontrolled: 
+* What should be done in the case `value=undefined`? (throw a warning or an error, do nothing)
```
j-f1: I’d say that `value={undefined}` means the same thing as not specifying `value`, and the behavior should be identical.

text/0000-allow-null-on-controlled-components.md
```diff
@@ -0,0 +1,111 @@
...
+> (Aside: it would also be nice to be able to bind Date objects to the value of
+> `<input type="date" />`, and I only bring it up because that would make this case even stronger.)
+
+@ryannerd :
+
+In SQL `null` is a fundamental concept. You never store an empty string in something like an
+`Address2` column; this would negatively impact storage space, and performance. Another example
+of `null` being a valid "value" is that averages, sums and other aggretate calculations exclude
+nulls to perform valid calcuations (`0 !== null`). This is important for React since persistent
+storage for many React apps is SQL based and thus would benefit from `null` as an allowed value.
```
AjayPoshak: strong argument 👍

alacret:
This is a very common problem on large systems, especially using types and Data Models o Models classes where you set to default or initial values.

So far 2 very ugly solutions:

1) To initialize string fields with empty strings, one by one because null is also used for object fields, or

2) End up with every single Input with: `value={value || '' }` 

and then create more functions or code to replace empty strings for null values.

Imagine maintaining this with 80 forms and around 700 fields.

text/0000-allow-null-on-controlled-components.md
```diff
@@ -0,0 +1,111 @@
...
+# Unresolved questions
+
+If the design choice is made where `value={null}` is always treated as string:
+* What about the situation where the controlled component is expecting a number or date?
+  Always treating `value={null}` as a string could be problematic.
+
+If the design choice is `value` for controlled and `defaultValue` for uncontrolled: 
+* What should be done in the case `value=undefined`? (throw a warning or an error, do nothing)
+
+For the alternitive design choice of `forceControlled` attribute:
```
styfle:
```suggestion
For the alternative design choice of `forceControlled` attribute:
```

asfernandes:
Adding more arguments about the SQL `NULL` being common (and also valid in JSON) and coming to the UI, in Oracle an empty string becomes `NULL` when read. The current behavior of using `value || ''` is very annoying.

Bessonov:
@gaearon @sophiebits is there any progress on this?

I think the convention of controlled/uncontrolled component is a bad idea. It's make things contra intuitive and leads probably to undefined or unwanted behavior (`A component is changing a controlled input of type number to be uncontrolled.`). Furthermore depending on `null` (and not for example only on `undefined`) make it worse.

While I agree that it's not easy to specify how `null` should work, the current state isn't satisfying at all. It's a bunch of awkward stuff without a clear and reliable workaround.

> The majority of the design is internal to React itself, and was under discussion by the core
> developers. The two design choices are:
> 
> 1. Always treat `value={null}` as empty string for controlled components.
> 2. @IndifferentDisdain :
>    > ... use `value` for controlled inputs and `defaultValue` for uncontrolled

Well, I think both ways are OK. I would be very happy to react on this breaking change. The second one makes the distinction more explicit. But I would like to propose two changes not depending on each other:
1. Introduce a new flag to indicate uncontrolled components. Something like `uncontrolled={booleanValue || false}` (I don't care about the name). I think most of the community use controlled components, because lack of knowledge that uncontrolled components exists at all (most of blogs and tutorials are focused controlled components) and in most cases it's used as a way to improve the performance of UI. Therefore I assume that this breaking change would affect only a small group of developers. But it's a win for quality and expectations.
2. If the react team is concerned about the large breaking change in regards of treating an empty string as `null`, then may be a flag like `nullValue=""` (or `isNullValue={component => component.value === ''}`) could help here.

kokushkin:
H! Come on, how this can break something, if there is a big red arrow message every time when you try to use a value={null} ? This is an Error! So what could be done better to make it possible to be implemented in the future?
image
Now, so many people want it to be fixed, including myself who run into this with Formik. And there are hundreds of more issues in the different libraries based on React. Why don't we fix the source of the problem instead of making  workarounds here and there? And if people don't like it then change it back.

gaearon:
I appreciate the effort, but this doesn't read as an RFC. It's mostly a copy paste of the existing discussion, and doesn't take any principled stance. It's hard to tell which of the options people in the comments are actually arguing for.

I will close this one but I would appreciate if someone could file an RFC for a concrete way forward that they feel confident in, together with the planned adoption strategy.

PR #54, opened 2018-06-19
Render children at most once through the use of Singletons

dkrutsko:
[View formatted RFC](https://github.com/dkrutsko/rfcs/blob/master/text/0000-create-singleton.md)

Referenced Issue: facebook/react#12744


gaearon:
Hi, thanks for the RFC.

The use case appears to be rather niche, and so far nobody has expressed support for it. My impression is that this is primarily useful for SVG filters, which are not that common in the wild. This doesn't mean the feature is unnecessary, but it's probably best implemented in userland.

The fact that multiple components could "propose" conflicting content for the same singleton seems like a red flag. It's relying on sibling order across the whole app, which makes reordering any subtree potentially a breaking change. We've been trying very hard to avoid such APIs in React because they are inherently fragile.

Instead, I would suggest to manage this explicitly — have each component export a pair: the actual component and the filter. The user needs to mount the filter somewhere. You can even enforce that it's mounted by getting it by ID in a layout effect.

PR #55, opened 2018-07-09
RFC: safeInContext hooks

mikesamuel:
Per ["Options for Hardening React &| JSX"][design] this proposes a hook API
to allow trusted framework code to prevent XSS by vetting values from
*AttributeExpression*s inside JSX before they reach browser internals.

[Proof of concept code][POC]

[design]: https://gist.github.com/mikesamuel/094d3fe4b1b2f702f543fa0c263ca8ff
[POC]: https://github.com/mikesamuel/react/commit/a63858f73a379aa1c12af6642ac84bcc8179f69c

### DIFF
--- a/text/0000-safe-in-context.md
+++ b/text/0000-safe-in-context.md
@@ -1,0 +1,303 @@
+- Start Date: 2018-07-09
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Per ["Options for Hardening React &| JSX"][design] this adds a hook API
+to allow trusted framework code to prevent XSS by vetting values from
+*AttributeExpression*s inside JSX before they reach browser internals.
+
+[Proof of concept code][POC]
+
+
+# Motivation
+
+```jsx
+// Value controlled by attacker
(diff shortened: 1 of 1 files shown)


mikesamuel:
@sophiebits Here's a minimal hook proposal and proof of concept code that would allow checking values.

This puts the details of defining value safety outside React proper.

I'm going to work on a standalone hook implementation.  Once I have that, we'll be able to evaluate all the moving pieces at once and decide what ideally goes where and how which parts a React.native analogue would benefit from.

sebmarkbage:
I think the current thinking is that we'll just add support for the Trusted Types spec in React. This will also provide additional coverage to other holes that still exist in life-cycles and imperative code. There is significant cost to migrating any sufficiently large app to a model like either of these solutions, so we want to make sure that it has broad support and has broad coverage.

(We might also just forbid href="javascript:..." specifically as a mitigation to the most common issue.)

gaearon:
We've added support for TT behind a flag, and also deprecated `javascript:` support. So I think this can be closed. Thank you.

PR #58, opened 2018-08-19
React.HeadlessComponent

sw-yx:
a proposal formally introducing Headless Components to React to enable better tooling! 

**Thanks for feedback so far, but please do not focus on the API at this point.** The API can change and of course can be implemented in userland - that is not the point of this RFC. **The point is to get this pattern recognized as a fundamental React component type so that we can build tooling on it.** No point debating the API until we decide whether or not this new component type belongs in React in the first place

# Please see [direct link to current text for easy reading](https://github.com/sw-yx/rfcs/blob/master/text/0000-React.HeadlessComponent.md)

# Basic example

adapted from @iammerrick 's [Headless User Interface Components](https://www.merrickchristensen.com/articles/headless-user-interface-components/):

```js
const flip = () => ({
  flipResults: Math.random()
})

// creation of headlesscomponent
class CoinFlip extends React.HeadlessComponent {
  state = flip()
  handleClick = () => this.setState(flip)
  children() { // note this is new
    return { // return an object
      isHeads: this.state.flipResults < 0.5
    }
  }
}


// usage: note CoinFlip's instance is the second param.
<CoinFlip>
  {({ isHeads }, {handleClick, state, setState}) => ( 
	// ...
```

Most importantly: You can toggle whether or not `<CoinFlip>` shows up in devtools. **Imagine what this looks like for apps with extensive usage of `react-redux` or `styled-components`.**


diegohaz:
I think that's a good idea. But what's the reason to pass properties and methods automatically? I'd like to choose what I want to pass to children explicitly in the children method.

sw-yx:
thanks for the comment @diegohaz! Its going to be the top use case for using a HeadlessComponent, so I'd like to make it easy for folks to use. that's all. if people really object then the auto passing can be done in userland, but that'd be annoying to add yet another library for an obviously very common usecase. 🤷‍♂️  API design always has tradeoffs.

text/0000-React.HeadlessComponent.md
```diff
@@ -0,0 +1,160 @@
...
+
+# Detailed design
+
+> This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with React to understand, and for somebody familiar with the
+implementation to implement. This should get into specifics and corner-cases,
+and include examples of how the feature is used. Any new terminology should be
+defined here.
+
+`React.HeadlessComponent` would be similar to `React.Component` in every way except it throws an error if `render` is implemented. In its place, it can (optionally) implement a `children` method that expects an object to be returned. This object will be spread together with its state when calling its child function.
```
j-f1: How do you feel about calling this `childState` and requiring the spread to explicitly happen in the method? It could default to `() => this.state`, though.

text/0000-React.HeadlessComponent.md
```diff
@@ -0,0 +1,160 @@
...
+> How should this feature be taught to existing React developers?
+
+"If you are making reusable components that only control behavior, use this for better tooling support and less boilerplate!"
+
+# Unresolved questions
+
+> Optional, but suggested for first drafts. What parts of the design are still
+TBD?
+
+- Whether or not HeadlessComponent needs to compile down to plain Component for production bundle size saving (i suspect it is immaterial)
```
j-f1: If it can be compiled down to a regular `Component`, then by definition it can be implemented in userland.

text/0000-React.HeadlessComponent.md
probablyup: I don’t see the point in having this function at all. If the component automatically passes what is in state, why add a second way to provide data and complicate things? Using lifecycle methods to populate state is sufficient.

text/0000-React.HeadlessComponent.md
probablyup: The component itself is implementable in UL, but the OP mentions several times that part of the appeal is tooling. 

That’s why this is being requested as a first party API, such that it can become a known thing and possibly optimized by React itself in the future.

text/0000-React.HeadlessComponent.md
sw-yx: ah i'm sorry, i actually wrote that before i realized that i could simply pass the HC's state as a second argument instead of spreading it together with the `children()` result. so that guarantees no name conflicts and hides the internal state while being accessible if you really need it!

text/0000-React.HeadlessComponent.md
sw-yx: ha, you got it Evan

alexkrolick: > it can be implemented in userland

Indeed: https://github.com/alexkrolick/react-renderless

text/0000-React.HeadlessComponent.md
```diff
@@ -0,0 +1,180 @@
...
+
+- if the child node of the HC is anything other than a function
+- if the HC implements a `render()` method
+- returning JSX inside `children()`
+
+# Drawbacks
+
+> Why should we *not* do this? Please consider:
+
+- **implementation cost, both in term of code size and complexity**: Not much, pretty much the same cost as PureComponent
```
bvaughn: I don't think I agree with this, unless I misunderstood something. 😅

What you're describing sounds like a new component API— No render method, but a new "children" method with unique behavior— so there would have to be a new code path to handle this component type in the reconciler.

Would this new type need similar will-update/did-update lifecycles to react to changes in state/props? (Seems a bit odd to have a "did mount" method for a headless component.)

Is the primary purpose of this RFC to flatten trees in DevTools or is that just a side effect? The PR description suggests this ("most importantly...") but reading the RFC it seems more of a secondary consideration. I ask because I think there are probably more lightweight options for achieving the DevTools goal.

sw-yx: > so there would have to be a new code path to handle this component type in the reconciler.

sure, but i figured the `children` method would just feed into the normal component `render` method under the hood? didn't think that would take more than a couple lines. obv its easy for me to be overoptimistic haha

> Would this new type need similar will-update/did-update lifecycles to react to changes in state/props? (Seems a bit odd to have a "did mount" method for a headless component.)

great qtn. i don't know; but to the extent that you can implement cDM for a current function-as-a-child higher order component i'd pretty much just expect it to work the same. can always put in a warning to guide people but i dont foresee any real harm leaving it in

> Is the primary purpose of this RFC to flatten trees in DevTools or is that just a side effect? 

Dev tools is the primary concern, yes. But the side effect of blessing this as a way to write headless components is probably also interesting for the rest of the ecosystem since it goes to the heart of how a lot of us write reusable components these days. If there are definitely more lightweight options you see that I don't I'd be fine closing this RFC. I actually figured there weren't any, since you need a hard guarantee of no render.

cellog:
I don't think react-redux would use this, because of the need to access context, which is done through a component now. Can you provide an example that can't be implemented through user land code showing why its necessary? This would help a lot with comprehension of why it is needed. Preferably one that uses the life cycle methods since that is what differentiates from functional component

text/0000-React.HeadlessComponent.md
satya164: > sure, but i figured the `children` method would just feed into the normal component `render` method under the hood? didn't think that would take more than a couple lines. obv its easy for me to be overoptimistic haha

do you mean it'll be implemented something like this?

```js
class HeadlessComponent extends Component {
  render() {
    return this.props.children(this.children());
  }
}
```

satya164:
Regarding the devtools stuff, personally I'd like to see a more generic API to collapse an element in DevTools, since this proposal implements a very specific version of render callback, i.e. a children function which accepts an object, which is the component state by default, and doesn't help with some other use cases:
  - Some components might want multiple render callbacks, for example this is one of my libraries which uses multiple render callbacks: https://github.com/react-native-community/react-native-tab-view
  - Some people prefer a `render` prop instead of a `children` function
  - HOCs are still nicer in many cases (theming for example), this will not help with HOCs

Though do people always want to hide these components in devtools? It receives some props and has some state, so if these are filtered out, how would we inspect their state and props? Also how would we inspect which arguments were passed to the `children` function?

Regarding IDE integration, maybe add a little bit more regarding how it'll help with autocomplete and typechecking and how syntax highlighting will be better? Given this only applies to writing these headless components and not using one (since usage is going to stay the same), these cases currently work great with static typecheckers and autocompletion. I also can't think of how syntax highlighting will change since there's no new syntax here. Do you mean IDEs can collapse things at usage site? Can't they already do that?

text/0000-React.HeadlessComponent.md
jamesplease: > Dev tools is the primary concern, yes.

Thought I'd pop in here to say that I'm not _too_ concerned with dev tools benefits here. I think there's value in improving the API of the "render prop pattern" for encapsulating state.

I'm not sure what the best approach to improving it is. Maybe it's some observable-like API (optionally with decorators, like MobX), or maybe it's this Unstated-like API, or maybe something else entirely. But I like seeing different options!

sw-yx:
@cellog - HeadlessComponent should have access to context; why wouldn't it? the key point is that react-redux doesn't really use `render`. any example i do can be done in userland; this is about declaring the fundamental type of the component to indicate it doesnt have any render impact beyond passing data to its children. aka "headless".

@satya164 sure, lets discuss API when we get there. we're not there yet.

regarding hiding headless components in devtools - its a toggle! :) these can all be solved once we provide devtools with the type info.

IDE: this is intentionally a bit more vague because i dont have experience making IDE extensions. i will say that its actually more for using rather than writing. the hope is that users have less cognitive load when it comes to distinguishing between a regular component and a headless one. i will move this down to Unresolved since i havent fully worked it out yet.

yea i just called it headless to pay homage to Merrick's post. naming things is hard 🤷‍♂️

PR #59, opened 2018-08-20
Ref convenience methods

EdwardDrapkin:
[View RFC](https://github.com/EdwardDrapkin/rfcs/blob/master/text/0000-ref-convenience-methods.md)

### DIFF
--- a/text/0000-ref-convenience-methods.md
+++ b/text/0000-ref-convenience-methods.md
@@ -71,7 +71,7 @@ Proposed new ref API:
 		
 		// reference interaction callback
 		// will be invoked whenever the ref changes or is set for the first time
-		onReference: (React$Ref<E>) => void,
+		onReference(React$Ref<E>): void,
 	}
 
 ```
--- a/text/0000-ref-convenience-methods.md
+++ b/text/0000-ref-convenience-methods.md
@@ -1,0 +1,96 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Introduce a small API to the ref object returned by createRef() that will make handling references more convenient.
+
+
+# Basic example
+
+```jsx
+import React, {Component} from 'react';
+import {animateLinkPressed, animateLinkAppearance} from './animations'; // hypothetically
+
+class RefExample extends Component {
+	buttonOne = React.createRef();
+	buttonTwo = React.createRef();
+	
+	constructor(props) {
+		super(props);
+		
+		this.buttonOne.onReference(ref => {
+			animateLinkAppearance(ref, 'green');
+		});
+		
+		this.buttonTwo.onReference(ref => {
+			animateLinkAppearance(ref, 'blue');
+		});
+	}
+	
+	onClick = (ref) => {
+		animateLinkPressed(ref);
+	}
+	
+	render() {
+		return <div>
+			<a ref={this.buttonOne} onClick={() => this.buttonOne.ifCurrent(onClick)}>Button One</a>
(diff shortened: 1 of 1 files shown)


text/0000-ref-convenience-methods.md
```diff
@@ -0,0 +1,96 @@
...
+		current: ?React$Ref, 
+		
+		// new additions 
+		
+		// optional handling
+		ifCurrent: (React$Ref) => void,
+		ifNull: () => void
+		
+		// reference interaction callback
+		onReference: (React$Ref) => void,
```
j-f1: Can you describe in more detail what each of these methods does?

text/0000-ref-convenience-methods.md
```diff
@@ -0,0 +1,96 @@
...
+Proposed new ref API:
+
+```
+	{
+		current: ?React$Ref, 
+		
+		// new additions 
+		
+		// optional handling
+		ifCurrent: (React$Ref) => void,
```
j-f1: How about typing these as `<T>(React$Ref) => T`?

mAAdhaTTah: FWIW, these typings aren't exactly correct. Probably need more like this:

```js
ifCurrent: <T>(callback: (React$Ref) => T) => ?T
```

I think the idea if having `ifCurrent` return the result of the callback, or `null` if the ref is `null`, would be nice (hence `?T`).

EdwardDrapkin: Can't see any good reason not to change the signatures, so I will update the RFC

text/0000-ref-convenience-methods.md
EdwardDrapkin: I added additional comments to that section, is more clarity necessary?

text/0000-ref-convenience-methods.md
EdwardDrapkin: I went ahead and typed them as `<T>(React$Ref) => T` because T can capture a type that's either nullable or not, and if it's not, we probably don't want to introduce a potential null in someone's code path that would otherwise have not had one.  It shouldn't get invoked if the `current` ref is null, otherwise we've just reshaped the issue with handling potentially null refs, rather than solved it.

I'm envisioning usage something like:

```jsx

render() {
  return <div>
    {this.state.video && <Video ref={this.video} />}
    <a onClick={() => {
      this.video.ifCurrent((ref) => ref.play());
      this.video.ifNull(() => alert('Video is not loaded'));  
    }}>Play</a>
  </div>;

}```

File: text/0000-ref-convenience-methods.md [hunk 3]
```diff
@@ -58,16 +58,20 @@ Proposed new ref API:
...
 		// new additions 
 		
 		// optional handling
-		ifCurrent: (React$Ref) => void,
-		ifNull: () => void
+		
+		// would be immediately invoked with the ref if the ref is valid
+		ifCurrent(callback: <T>(React$Ref<R>) => T): T,
+		// would be immediately invoked if the ref is invalid
+		ifNull(callback: <T>() => T): T,
```
j-f1: I think these `<T>` parameters need to be right after the function name (`ifCurrent`/`ifNull`) so the type can be used of the function.

EdwardDrapkin: good catch

gaearon:
Thanks for the RFC!

We have generally avoided convenience methods like this. There is a fixed cost for every extra field or lookup so it's not ideal from performance point of view if it can be avoided. It hides the meaning behind some indirection. If you know about the existence of this indirection, you might as well check for `.current` yourself. In my experience, people who don't check `.current ` also won't know to call `.ifCurrent()`. So I don't think this RFC would solve the issues outlined in the motivation. Combined with other concerns, I think it's not worth the benefit, but I appreciate you writing it up.

PR #61, opened 2018-09-20
React.Children.toFlatArray

arty-name:
[View formatted RFC](https://github.com/arty-name/rfcs/blob/Children-toFlatArray/text/0000-children-toflatarray.md)

Per [a suggestion from Dan Abramov](https://github.com/facebook/react/issues/11859#issuecomment-355656427): this RFC proposes adding a new API `React.Children.toFlatArray` to get the children as an array while treating the `Fragments` as transparent containers.

### DIFF
--- a/text/0000-children-toflatarray.md
+++ b/text/0000-children-toflatarray.md
@@ -1,0 +1,155 @@
+- Start Date: 2018-09-20
+- RFC PR:
+- React Issue:
+
+# Summary
+
+A new API `React.Children.toFlatArray` will traverse not only the direct children
+but will also go into `Fragments` and recursively traverse their children as well
+so that the resulting array does not contain any `Fragments`.
+
+# Basic example
+
+```jsx harmony
+import React, { Fragment, Children } from 'react'
+
+function TabsContainer({ children }) {
+  console.log(Children.toFlatArray(children)) // [<Tab/>, <Tab/>, <Tab/>] instead of [<Tab/>, <Fragment />]
+}
+
+function App() {
+  return (
+    <TabsContainer>
+       <Tab/>
+       <Fragment>
+         <Tab/>
+         <Tab/>
+       </Fragment>
+    </TabsContainer>
+  )
+}
+```
+
+# Motivation
+
+There has been a need to operate on sets of sibling elements without a wrapper DOM element. 
+This need has been addressed by introducing `<React.Fragment>` which allows passing around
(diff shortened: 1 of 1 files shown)


TryingToImprove:
I like the proposal, but I think that their should be a `dept` like with `Array.flat`, but seems like such a thing would allow developers to take the entire three in their Root element. 

I would prefer a option to `React.Children.toArray(children, { threatFragmentsChildrenAsChildren: true })`.

text/0000-children-toflatarray.md
```diff
@@ -0,0 +1,155 @@
...
+       )}
+    </TabsContainer>
+  )
+}
+```
+
+# Drawbacks
+
+- larger API surface due to the new API
+- the rest of the APIs in the `React.Children` namespace keep their potentially surprising behaviour
```
Andarist: I would also be surprised by this, not only to sole fact that Fragments are not flattened - as this can be more or less easily memorised, but surprising to me is that regular methods flatten array (& iterators), but not Fragments. This is imho inconsistent slightly and will be even more surprising to see some children being already flattened by no-flattening (taking into account its name) method when you introduce API with the name flat in it.

arty-name: Unfortunately I don’t know the motivation behind the current state of affairs but I assume that there are reasons for this intentional inconsistency. Hopefully the developers will explain them in the comments here.

arty-name:
I think I have failed to explain my idea well enough. The current behaviour of `React.Children` with `<Fragment>` is already problematic and I would like to have an unproblematic way to handle them. Either by this new API or by changing existing behaviour of `React.Children`. So the result my proposal will not be changed by extracting a `<Fragment>`.

It would be nice to have a more convenient stop-gap solution until you guys figure out the right approach.

birkir:
https://github.com/birkir/react-children-addons#toflatarray

dreampulse:
Is there a solution on the horizon?

grrowl:
I just open-sourced some code from my own projects which handles this (in a specific way which ensures stable keys like `Children.toArray` does), would love feedback if anyone uses it! https://github.com/grrowl/react-keyed-flatten-children

gaearon:
Since https://github.com/grrowl/react-keyed-flatten-children seems to implement this and `React.Children` is in maintenance mode (and will not be recommended after we figure out something like call/return), I'm going to close this.

PR #62, opened 2018-10-18
shouldComponentRemount

alexkrolick:
**[👉  Formatted RFC](https://github.com/alexkrolick/rfcs/blob/patch-2/text/0000-shouldComponentRemount.md)**

This is a proposal to add a new lifecycle method or higher-order component to handle resetting component state when props change. (Think of this as a controlled `key`).

Returning `true` from this method would be equivalent to changing the `key` prop -
it unmounts and remounts the component instead of preserving the instance.

---

This method would help component authors define the relationship between components and the data they represent by providing a way to say "the entity this component represents has changed, remount".

Per @vcarl:

> adding a key is a very elegant solution for when you need to re-run mount behaviors, it has the significant drawback of being completely outside the component's control.

This method would also remove a lot of housekeeping boilerplate that is written to make components able to receive props that require a full state reset (typically a combination of componentDidMount, componentWillReceiveProps/getDerivedStateFromProps, and componentDidUpdate).

---

Userspace implementation of a proposal using a higher-order component: https://github.com/alexkrolick/react-remount-component

### DIFF
--- a/text/0000-shouldComponentRemount.md
+++ b/text/0000-shouldComponentRemount.md
@@ -61,7 +61,7 @@ class ImgZoomer extends Component<{ imgUrl: string }> {
 The goal is to provide a way for components to manage the relationship between entitites provided by props and
 the UI state associated with them.
 
-Example: The [You Might Not Need Derived State][ymnnds] blog post recommends a pattern
+Example: The [You Might Not Need Derived State][ymnnds-key] blog post recommends a pattern
 using an uncontrolled, stateful component with a `key` prop to reset state.
 
 One pitfall with the key approach is that the consumer of the component is responsible for picking a
--- a/text/0000-shouldComponentRemount.md
+++ b/text/0000-shouldComponentRemount.md
@@ -2,18 +2,69 @@
 - RFC PR: (leave this empty)
 - React Issue: (leave this empty)
 
-_**Draft**_
+| Date | Change |
+| ---- | ------ |
+| 2018-10-17 | Initial |
+| 2019-02-03 | Add HOC option |
 
 # Summary
 
-This is a proposal to add a new lifecycle method, `shouldComponentRemount`.
-Returning `true` from this method would be equivalent to changing the `key` prop -
+This is a proposal to add a new lifecycle method or higher-order component to handle resetting component state when props change. (Think of this as a controlled key).
+
+Returning true from this method would be equivalent to changing the key prop -
 it unmounts and remounts the component instead of preserving the instance.
 
-The method would receive `(nextProps)` as an argument. Alternatively it could be a static method
-that receives`(nextProps, prevProps)`.
+This method would help component authors define the relationship between components and the data they represent by providing a way to say "the entity this component represents has changed, remount".
+
+Per @vcarl:
+
+> adding a key is a very elegant solution for when you need to re-run mount behaviors, it has the significant drawback of being completely outside the component's control.
+
+This method would also remove a lot of housekeeping boilerplate that is written to make components able to receive props that require a full state reset (typically a combination of componentDidMount, componentWillReceiveProps/getDerivedStateFromProps, and componentDidUpdate).
+
+Userspace implementation of a proposal using a higher-order component: alexkrolick/react-remount-component
 
 # Basic example
+
+## HOC for Function Components/Hooks
+
+(Also works with class components but a lifecycle method may be more natural, see below).
+
+```jsx
+function UserCard(props) {
+  const { user } = props;
+  const [isExpanded, setExpanded] = React.useState(false);
+  return (
+    <div className="user-card">
+      <div>Name: {user.name}</div>
+      <div>
+        <button onClick={() => setExpanded({ expanded: true })}>
+          See Details
+        </button>
+      </div>
+      {isExpanded && (
+        <React.Fragment>
+          <div>Birthday: {user.birthDate}</div>
+          <div>Favorite Color: {user.favoriteColor}</div>
+        </React.Fragment>
+      )}
+    </div>
+  );
+}
+
+export default withRemount(UserCard, (props, prevProps) => {
+  const prevUserId = prevProps.user && prevProps.user.id;
+  const nextUserId = props.user && props.user.id;
+
+  if (nextUserId !== prevUserId) {
+    return true; // remount!
+  }
+
+  return false; // don't remount otherwise
+});
+```
+
+## Lifecycle Method For Class Components
 
 ```diff
 class ImgZoomer extends Component<{ imgUrl: string }> {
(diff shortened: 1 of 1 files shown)


vcarl:
I like this! I was thinking about this the other day: adding a key is a very elegant solution for when you need to re-run mount behaviors, it has the significant drawback of being completely outside the component's control. 

Another possibility I see would be something like an `identity` function that returns something analogous to a `key` prop. If it changes, the component is destroyed and recreated.

alexkrolick:
> Another possibility I see would be something like an identity function that returns something analogous to a key prop. If it changes, the component is destroyed and recreated. getDerivedKeyFromProps?

Yeah, I put something like that in the Alternatives section. I think I prefer returning true/false because you don't have to serialize the result to a string which is a known difficulty with keys currently. Internally React could basically assign a random key whenever the function returns false.

text/0000-shouldComponentRemount.md
```diff
@@ -0,0 +1,125 @@
...
++    if (newProps.imgUrl !== this.props.imgUrl) return true;
++  }
+
+  zoomIn = () => this.setState(state => ({ zoom: state.zoom + 1 }));
+  
+  zoomOut = () => this.setState(state => ({ zoom: state.zoom - 1 }));
+
+  render() {
+    if (this.props.error) return 'Error';
+    if (this.props.loading) return 'Loading...';
```
gaearon: It seems like the idiomatic way to do this example would be with Suspense, which wouldn't suffer from this problem in the first place.

alexkrolick: Also, in this scenario it might actually be easier not to use Suspense. Externalizing the data-loading responsibility into a cache means that the component can't directly access loading events and use them to reset UI state, unless you use something more formal like a state machine or Redux store that would pull state out of the component anyways.

alexkrolick:
> How does this work for function components?

Function components would fall more under the use case of differentiating items in a list without having to reduce/serialize the identity down to a string.

I think you could approach this in a few ways. In both cases the function signature would need to be `(nextProps, prevProps)`. I am not sure if keeping prevProps around for comparison is viable.

1. Make `shouldComponentRemount(nextProps, prevProps)` static on both class and function components:

```js
function MyComp (props) { return <h1>{props.text}</h1> }
MyComp.shouldComponentRemount = function(nextProps, prevProps) {return true}
```

2. Make `key` accept a function, or make `shouldComponentRemount` a prop that operates the same way. This would end up looking similar to https://ui.reach.tech/component-component

sag1v:
@alexkrolick 
>Returning `false` from this method would be equivalent to changing the key prop -

Maybe i misunderstood but if the method is named `shouldComponentRemount` shouldn't it return `true` to mimic the `key` prop change?

So i would expect to see this:  
 *Returning `true` from this method would be equivalent to changing the key prop -*

sag1v:
>How does this work for function components?

@gaearon If this is categorized as a life-cycle method why should it be any different from any other life-cycle methods? none of them are used in function components (at least not in user-land).

sag1v:
I like this suggestion and its sure can help, but just for the protocol we can achieve similar behavior today with a simple HOC.  

(Naive implementation) :  
```
const withRemount = (Component, key) => (props) => {
  const asKey = props[key];
  return <Component key={asKey} {...props} />
}
```

This is how you can use it:  
```
// this is our component (maybe our library)
class User extends React.Component {
  componentWillUnmount() {
    // just to test it
    console.log('componentWillUnmount')
  }
  render() {
    const { name, userId } = this.props;
    return (
      <div>{`Hello ${name}, your id is ${userId}`}</div>
    )
  }
}

// We can export this version to our consumers
// export default withRemount(User, 'userId');
const RemountingUser = withRemount(User, 'userId');

  state = { name: 'john', userId: 123 }

  onNameChange = ({ target }) => this.setState({ name: target.value });

  onIdChange = ({ target }) => this.setState({ userId: target.value });

  render() {
    return (
      <div className="App">
        <input type="text" onChange={this.onNameChange} placeholder="name" />
        <input type="text" onChange={this.onIdChange} placeholder="userId" />
        {/*The consumer doesn't know but the userId they provide is actually our key */}
        <RemountingUser {...this.state} />
        {/*
          The great thing about it is thta they can provide a different key 
          and we are not breaking the default behaviour
          <RemountingUser {...this.state} key={this.state.name}/>
        */}
      </div>
    );
  }
}
```

theydonthaveit:
The responsibility to decide "whether or not" or "at which moment to check if" a component should remount or not needs to reside with the engineer. We write the logic. Either way, the component remounts when it needs to. This additional lifecycle method seems frivolous.

popcon21:
shouldComponentRemount

alexkrolick:
OK so now that stateful function components are a thing thanks to hooks, I think a static method on both class and function components would make the most sense.

```js
  static shouldComponentRemount(oldProps, newProps) {
    if (oldProps.id !== newProps.id) return true
    return false
  }

  render() {
    return (/* something */)
  }
}
```

```js
function Bar(props) {
  /* hooks n stuff */
  return (/* something */)
}

Bar.shouldComponentRemount = function (oldProps, newProps) {
  if (oldProps.id !== newProps.id) return true
  return false
}
```

Doesn't look great on functions but there's precedent with defaultProps, propTypes, contextTypes, etc.

alexkrolick:
Userspace implementation:

```js
const withShouldComponentRemount = WrappedComponent =>
  class WrapperComponent extends React.Component {
    static getDerivedStateFromProps(props, {prevProps, key}) {
      return {
        prevProps: props,
        key: WrappedComponent.shouldComponentRemount(props, prevProps)
          ? `${Math.random()}`
          : key,
      };
    }

    static displayName = `withRemount(${
      WrappedComponent.displayName || WrappedComponent.name
    })`;

    state = {
      key: `${Math.random()}`,
      prevProps: {},
    };

    render() {
      return <WrappedComponent key={this.state.key} {...this.props}  />;
    }
  };
```

sag1v:
@alexkrolick Have you tested your implementation? not sure you have access to the `key` prop when it passed like that.

BTMPL:
If you only care about props, maybe a better solution for functional components would be a wrapper, like memo?

```
const Bar = React.withRemount(Foo, (oldProps, newProps) => {
  return oldProps.id !== newProps.id
});
```

sag1v:
@BTMPL As i mentioned in my [comment](https://github.com/reactjs/rfcs/pull/62#issuecomment-431298837) this can be achieved in user-land.

In fact, I was playing with the idea and created a small library ([react-remount]( https://github.com/sag1v/react-remount)) that is doing just that.
If you want to play with it.  :nerd_face:

alexkrolick:
I like how @BTMPL's HOC allows comparing prop changes in a function and feels similar to new APIs like React.lazy, and React.memo. 

@sag1v, in your version the key is a copy an existing prop, and doesn't compare prop changes. Works for some use cases but it's less flexible.

sag1v:
@alexkrolick not sure i follow, what use case it doesn't work for?
It will just inject a `key` prop with the value of the `prop` you set. the comparing will be automatically handled by react.

BTMPL:
Having a function is just more flexible, e.g. you can recreate on specific condition, not only on change. Can still be implemented in userland though?

sag1v:
@alexkrolick 
>I don't want to have to come up with a serialization of the key

You don't need to. you provide a prop / props that you think should re-mount when their value is changed.
>Having the key be randomly generated when a function returns true allows more complex logic and decouples the remount logic from the particular implementation of using a key at all.

That's exactly the point, you are not using a key as the user of the component, you just tell it what changes should re-mount it. under the hood it will pass the _value_ of the props you defined as a key for you.

alexkrolick:
Again, that depends on the value of the props in question being serialized. That shouldn't matter. If you want to implement that as the solution in your library, that's fine, but the ask in this RFC is more general.

sag1v:
@alexkrolick Sorry if i wasn't messaging clearly, i'm really not trying to be defensive here.
I never used this library and i doubt i'll use it anytime soon. I wrote it only because i thought this was an interesting challenge, and actually i still think it's interesting. This is why i asked for a specific use case where you think it can't cover. I just wanted to keep playing with it and see if it is doable without another React API.

All in all i think that what you are asking in your RFC is something that is doable already with the tools `React` provides today.
The mechanism for re-mounting a component is available with `key` props.
The only downside with it, is that the consumer of A component is responsible to provide that `key`, which is kind of implementation details and totally a none-friendly API.  
and this is what i was trying to solve with the library i wrote, some way to provide a key to a component that would be transparent to the consumer of that component.

alexkrolick:
Implementation of a user-space solution: https://github.com/alexkrolick/react-remount-component

Based on @BTMPL's idea of an HOC that doesn't require static methods.

gaearon:
My impression is that a `withRemount` API doesn't buy us anything meaningful over

```js
function Inner() {
  // ...
}

function Outer({ id, ...rest }) {
  return <Inner {...rest} key={id} />
}
```

So I think we can close this.

gaearon:
Addressing this specifically:

>adding a key is a very elegant solution for when you need to re-run mount behaviors, it has the significant drawback of being completely outside the component's control.

A component can always control this by exporting a wrapper. This should be unobservable.

PR #63, opened 2018-10-19
React.memo()

gaearon:
### [View formatted RFC](https://github.com/reactjs/rfcs/blob/gaearon-patch-1/text/0000-memo.md)

**Note: I just wrote up the RFC. The semantics are designed by @sebmarkbage and @acdlite.**


sag1v:
This is a great idea, I do think the name `equal` may confuse some people.  
Maybe something closer to the life-cycle methods we have?  
`shouldUpdate` 
`shouldRender`

gaearon:
The argument name doesn't matter, it's not technically a part of the API. We can call it `arePropsEqual` in the docs.

maciejsikora:
I very have the same doubt about naming of the second function argument `equals` is not familiar name in React domain. I think `shouldUpdate` as @sag1v proposed is more accurate, as this function does the same what `shouldComponentUpdate` in orginal PureComponent

gaearon:
>What if component always stays the same, it would simply have a .shouldComponentUpdate prop, which—if not implemented—results in current behaviour.

This introduces an extra runtime property check for every single function component, making the whole app slower and potentially causing deoptimizations. This design was considered (see "Alternatives") and rejected.

>You defined the second parameter to only be a function, I would personally choose an object as it is more future proof and allows for more customization later on, eg. implementing life cycle methods that change props here.

There are no plans for making this function more advanced as it would defeat the optimizations it's supposed to provide.

foiseworth:
For clarification, is this RFC about the addition of the second optional argument? I thought this PR meant that in a future release we would have a pure function: https://github.com/facebook/react/pull/13748/files

baddox:
The shouldComponentUpdate property also mutates the functional component, which means couldn’t easily export the original component and a pure version.

gaearon:
> I thought this PR meant that in a future release we would have a pure function:

This *is* an RFC covering that PR. We've merged it to test it at FB but now we're getting closer to an open source release and want to get community feedback.

falconmick:
I would vote for the breaking change. I would even almost like this to be the case for Component (i.e. get rid of PureComponent and just make shallow compare the default) and instead of making code which works well when mutable, make it so that if you mutate props or state that you get bugs right away instead of in 3 months time when the code has solidified.

This would really suck for new devs, but could help drive home the idea of immutability and pure funcs. oh and good luck with migrating the millions of facebook Components, I don't think a codemod would fix immutability bugs.

tomhicks:
I think the second argument should be the ”same” as `shouldComponentUpdate` in that it returns true if it should update.i know the arguments are different for this than a class (nextProps, nextState vs prevProps, nextProps) but returning true when it should update is kind of a React idiom.

To do the opposite (presumably because you just want to put `shallowEqual` in there) is weird, IMO.

Unless I'm missing something...

gaearon:
>I would vote for the breaking change. I would even almost like this to be the case for Component (i.e. get rid of PureComponent and just make shallow compare the default) and instead of making code which works well when mutable, make it so that if you mutate props or state that you get bugs right away instead of in 3 months time when the code has solidified.


It wouldn't be just bad because it's a breaking change. It would also be worse because:

1. Many components will never have shallow bailouts, meaning they would overall make the app *slower*
2. It would be impossible to use *any* regular function component with mutable data coming from above, effectively causing lots of people to convert them to classes (and also make their apps slower)

Nimelrian:
> I would even almost like this to be the case for Component (i.e. get rid of PureComponent and just make shallow compare the default)

This would introduce a hefty performance fine in many cases

> This would really suck for new devs, but could help drive home the idea of immutability and pure funcs. 

I'm a fan of "learn by pain", but the downsides are way too big to introduce this breaking change.

Your comment regarding the internal Facebook components actually makes me think you're a troll.

streamich:
Would be nice if the second argument was called `shouldUpdate` instead of `arePropsEqual`.

- More consistent with prior art.
- Maybe your decision to update (or not) the component is not purely based on props being equal.

tomhicks:
@Nimelrian almost certainly he's not trolling. He's saying, if you did go with the breaking change, good luck updating. It's a bit of happy sarcasm that I think you've misinterpreted.

inferusvv:
Why `arePropsEqual`? In Component we use `shouldComponentUpdate` which will re-render component if function returns `true`.  And `arePropsEqual` should return `false` to re-render component. I think it is a bit confusing and better keep shouldComponentUpdate behavior everywhere.

falconmick:
@Nimelrian no not a troll. I used facebook as an example of how this wouldn't work because I am well aware that given their amount of components that they would struggle to support this as their usual strategy for breaking changes are codemods.

It's unrealistic to expect facebook or even more so, the world to be able to handle such a drastic change like that without way too much pain.

tbh my comment was pointless given it's just not feasible to do something as drastic as that, so I really shouldn't have left any more feedback other than I would prefer the default behaviour to be shallow compare. However it's just not something that i see actually having enough net positive.

tomhicks:
Ok I'm glad I'm not the only one thinking it should be shouldUpdate not are same.

I'll go further and say what I wanted to say originally now I know I'm not crazy - unless there's some particular technical objection to it, it's an absolute no-brainer that it should have `shouldUpdate` behaviour

sag1v:
What i love the most about react's naming stuff in API is that its telling us what is the *Purpose* of the function / parameter,  rather than what it **Does** (implementation details?). 
Therefor `shouldComponentUpdate` or `shouldUpdate` are a great fit.

Same goes for `pure()` IMO, this is a pure function component. 
Rather then `memoized()`, this function is memoized (do i need to know this detail?)

Nimelrian:
@falconmick Than I misread your comment, sorry!

aweary:
This API seems like it enables making third-party components pure. What would be the behavior if you passed an already pure component to `pure`?

```
// I don't know if Button is already using React.pure
import {Button} from 'third-party'
import {pure} from 'react'

// Noop? Error?
export default pure(Button);
```

Following that thread, what happens if you pass a class component? Can users safely apply `pure` to third party components without knowing how they're implemented?

jamby77:
How about `pureComponent(Component, shouldUpdate)`

BurntCaramel:
This looks great! What would be current advice of when to use this and when not to? Is this generally a good default or often preoptimization?

giuseppeg:
> How about pureComponent(Component, shouldUpdate)

Another idea `React.createPure(Component, shouldUpdate)`. Although I equally like `React.pure` 👌 
Would `displayName` still be the function component name?

aweary:
I think `arePropsEqual` is better, mainly because of this point:

> you can more easily pass a different off-the-shelf implementation (which tend to be written in the positive form).

That means you can do things like:

```jsx
import {isEqual} from 'lodash'
import {pure} from 'React'

const Button = props => {
  // ...
}

export default pure(Button, isEqual)
```

Not that you _should_ use lodash's `isEqual` all the time, but most custom equality modules are implemented this way, making it easy to reuse common equality logic.

streamich:
@aweary Then it should be `shouldNotUpdate`.

grabbou:
This proposal looks great! 

I have one comment regarding the public-facing API and the naming. In my opinion, `pure` as a name, together with a second, optional, argument might be a bit confusing for developers.

Alternatively, we could consider exposing two functions. First one, `shouldUpdate` would take a functional component to optimise and a `shouldComponentUpdate` function, which is required.

The second function, `pure`, would be implemented on top of it by supplying the default `shouldComponentUpdate` function. This step is optional and could as well live in the user land (e.g. inside recompose). This would enable community implementations to provide similar performance.

```js
const pure = (component) => shouldUpdate(component, defaultShouldUpdate)
```

troch:
> Function components could be `pure` by default (breaking change)

It would be great. I wonder how many people would be affected by this breaking change. It would definitely result in hard to track issues. Dev warnings could be added, but they might be expensive?

villesau:
What about baking the info straight to the SFC, like:

```
const Button = () => <button />
Button.pure = true;
export default Button;
```

This way you wouldn't need to import anything, and you tell react how the component should be treated.

Even better would be if you would explicitly need to set `Button.pure = false;` in case you don't want to treat it as pure.

danburzo:
(_Hoo boy, this thread has filled up from the time I wanted to write until I finally managed to._ 😄)

I wanted to echo the opinion that `arePropsEqual`, in combination with `shouldComponentUpdate` may baffle beginners, and IMHO `shouldUpdate` would make more sense to keep in line with the parallels between `pure` and `PureComponent`. I don't see many cases where you'd need a custom comparison function, and in any case it's not too much of a hassle to write:

```js
pure(MyComponent, (prevProps, nextProps) => !_.isEqual(prevProps, nextProps));
```

(And, even if theoretically the comparison *should* be pure, there's no reason why someone would not be able to take into account other aspects than `prevProps` and `nextProps`; `shouldUpdate` would convey this in a clearer fashion)

Also, I wanted to put out an idea that I haven't really thought through, but: would it be possible to leverage `pure` to optimize _render props_, or would that defeat the optimizations planned for it?

streamich:
@danburzo 

> ... would it be possible to leverage pure to optimize render props, or would that defeat the optimizations planned for it?

Do you mean to ignore `render` and `child` props if they are functions?

tomhicks:
@danburzo had the best argument for "shouldUpdate" over "arePropsEqual".

Say you wanted to do a performance comparison of memorized Vs not, you would have to state "props are not equal” which is not true, when you actually mean “this should update” which is true.

tomhicks:
@rtorr not so sure about memoize as a name, just because it’s not “normal” memoisation. It’s memoisation based on shallow equality of an object, which is not like “normal” function memoisation that you would find in any fp lib.

Konstantin6487:
What about decorators to RFC?

danburzo:
@streamich, sorry — I meant to visualize how / if we might apply `pure` to other areas that use [render props](https://reactjs.org/docs/render-props.html), like the Context API. For example, I wonder if the Consumer will be updated to work with `pure`; and in the case of other instances of the "function as a child" pattern, would you need to adhere to sending the values from the parent component to the child as a single argument (faux `props`)? (As I said, my idea is not fully fleshed out :) )

```js
const fn = pure(value => <Component value={value}>);
...
<Consumer>{ fn }</Consumer>
```

3rd-Eden:
> This introduces an extra runtime property check for every single function component, making the whole app slower and **potentially** causing deoptimizations. This design was considered (see "Alternatives") and rejected.

Has there been any testing on this or are these just assumptions. Because if I look at the `pure` functionality right now it returns an object:

```js
{
    $$typeof: REACT_PURE_TYPE,
    render,
    compare: compare === undefined ? null : compare,
}
```

With an optional `compare` property. One would assume that it the same amount of overhead of checking a functions property vs an object property.

My main concern and HOC pattern in general is that they can potentially obfuscate the stacktrace by returning a different class/function from a different part of the software stack.

rtorr:
@tomhicks very true. I guess I some what imagined a long term thing where a lower level memoize utility got pulled out of react and could be used (with user supplied rules) on any data comparison. The built in one for react happens to supply a shallow equality on the props object. I guess that could be named pure for react context. 

Anyway, other than bikeshedding the name, I’m a 👍 on this rfc either way.

CharlesStover:
My 2¢:

I don't care if it's a HOC or not. I'd use whichever has better performance.

I think a good choice that solves a lot of the drawbacks would be to call the `pure` function `shouldComponentUpdate` and deprecate `shouldComponentUpdate` inside classes (and deprecate `PureComponent`).

This removes duplication with the `PureComponent` API. This removes confusion over whether the second argument is `arePropsEqual` or `arePropsDifferent`. This removes the confusion around the name `pure` as it no longer overlaps with the concept of purity. This does not limit bailouts to classes. This is not a breaking change for function components. This gives it a longer and different name. This allows a custom equality function. This can work with class components too.

I'm not seeing any downsides to `shouldComponentUpdate(functionalOrClassComponent, customEqualityFunction)`.

Something along the lines of this, but with better implementation:

    const shouldComponentUpdate = (Component, componentShouldUpdate) => {
      let memoizedInstance = null;
      let oldProps;
      return newProps => {
        if (componentShouldUpdate(oldProps, newProps)) {
          memoizedInstance = <Component {...newProps} />;
        }
        oldProps = newProps;
        return memoizedInstance;
      };
    };

    const Button = ({ children }) => <button children={children} />;
    return shouldComponentUpdate(Button, (oldProps, newProps) =>
      oldProps.children !== newProps.children
    );

      render() {
        return <button {...this.props} />;
      }
    }
    export default shouldComponentUpdate(Button, React.shallowEqual);

If it's a big deal that the class implementation of `shouldComponentUpdate` responds to state, the callback for `setState` can be `shouldComponentUpdate` such that it returning true or false controls updating, maybe defaulting to `React.shallowEqual`.

trueadm:
@gaearon I really dislike `React.memo`. The word "memo" means an entirely different thing than memoization (especially in British, where the word memo is frequently used). I know it's longer, but  `React.memoize` just makes more sense to me.

kingdaro:
Maybe even `memoized`, since `memoized(Button)` reads nicely

RomainLanz:
I always prefer to have a full name than some abbreviation. No need to win 4 characters by naming it `memo`.

Let's go for `memoized` 👍

ConAntonakos:
Thank you for being transparent with these changes. I agree with @trueadm. I think the naming should be a verb as well to maintain consistency that it is composing/enhancing/augmenting your input/argument. I think `memoize` or even `purify` are good candidates.

gaearon:
>means an entirely different thing than memoization

As @acdlite pointed out, from https://en.wikipedia.org/wiki/Memoization#Etymology:

>The term "memoization" was coined by Donald Michie in 1968[3] and **is derived from the Latin word "memorandum" ("to be remembered"), usually truncated as "memo" in American English,** and thus carries the meaning of "turning [the results of] a function into something to be remembered." 

So in that sense it is related exactly to what the function is doing. (Remembers last rendered result.)

Additionally, `memoize` is misspelled and misunderstood too often (people think it's "memorize" or "memorise"). It's also unfortunately one of those verbs that are written differently in British and American English. So I think shortening it to `memo` makes practical sense.

gaearon:
> I think memoize or even purify are good candidates.

Argument against `memoize`: it **does not** do what the `memoize` function does. If you import `memoize` from Lodash in the same file it won't work. That's pretty confusing.

Argument against `purify`: it's even further from truth than `pure`. At least with `pure` we were "tagging" a component as being "pure" in React sense (which was already confusing because it doesn't match the computer science definition of purity). With `purify` it's as if you can take something impure, and then this function makes it pure. That's not at all what it does.

gaearon:
>With an optional compare property. One would assume that it the same amount of overhead of checking a functions property vs an object property.

We are already checking `$$typeof` on every type. But if it's a function component we don't need to do more checks. Whereas if we used a flag, we would have to check every single function component.

aweary:
Using `memo` instead of `pure` also makes the `shouldComponentUpdate` argument moot because the API is no longer in parallel to `PureComponent`. It makes much more sense to pass an `isEqual` function to a memoization utility.

ConAntonakos:
> Argument against memoize: it does not do what the memoize function does. If you import memoize from Lodash in the same file it won't work. That's pretty confusing.

Ah, yes, naming conflicts. That makes sense. 👍 

>Argument against purify: it's even further from truth than pure. At least with pure we were "tagging" a component as being "pure" in React sense (which was already confusing because it doesn't match the computer science definition of purity). With purify it's as if you can take something impure, and then this function makes it pure. That's not at all what it does.

That makes sense as well. 👍 Thank you for explaining!

This is quite the challenge. 😄

text/0000-memo.md
```diff
@@ -0,0 +1,84 @@
...
+By having `memo` as a first-class API in React itself, we can remove the need to make a choice between these optimizations. It makes it easy to memoize the output of function components without introducing other extra costs.
+
+This also helps address a common argument in teams that can't decide whether to use one or the other optimization, by letting you use them together. And unlike a userland `memo()` higher-order component implementation, the one built into React can be more efficient by avoiding an extra component layer.
+
+
+# Detailed design
+
+`React.memo` returns a component type that tells React "render the inner type, but bail out on updates if props are shallowly equal". In other words, the rendering result is memoized. The prop comparison function can be specified as a second argument to `React.memo`.
+
+`React.memo` only accepts function components as the first argument. It doesn't accept classes since those can extend `PureComponent` directly, and mixing the two approaches is confusing. `React.memo` is intended to be applied at the definition site.
```
aweary: How does this work with `React.forwardRef`?

theKashey:
Would this `memo` work on the server side between different renders, as real `memoization` could? (like [this](https://twitter.com/tannerlinsley/status/1052279127229706241))
Sounds like - no, it wouldn't.

aweary:
@theKashey memoizing between requests could run the risk of caching and leaking stale user data.

tomhicks:
@aweary not if it's pure. That's basically the entire point of memorisation...

aweary:
It's still possible if a component in the cached subtree read data from an external source. `memo` doesn't guarantee that every descendant component is actually pure. Context can also complicated this.

einarq:
How does it compare performance-wise to wrapping SFC in a PureComponent? Is it using PureComponent under the hood or is it more optimized?

tomhicks:
@aweary in that case it's not pure.

I realise that this is probably academic and your real-wotld example is probably more helpful, but just sayin'

text/0000-memo.md
aweary: [This PR answers my question](https://github.com/facebook/react/pull/13897_

NE-SmallTown:
> Argument against memoize: it does not do what the well-known memoize function does.

If so, the `memo` name is wrong because it still indicates it is the same with 'memoize', just like you want to distinguish between 'Dan' and 'Dan Abramov'

> If you import memoize from Lodash in the same file it won't work. That's pretty confusing.

Because any 3rd party library could have the name `memoize`, so in this situation, we should `import * as lodash from 'lodash'` rather than don't use the `momoize` as name.

fardarter:
My request is that code example include how to wrap around a redux-connected component or how to compose with other HOC libraries.

I'd say being able to pass an equality function down is great but it would be great if the code examples included an implementation example.

Also, would be great if the examples showed an example of where shallow equality will fail.

`arePropsEqual` could be `renderOnDifference`.

gaearon:
>If so, the memo name is wrong because it still indicates it is the same with 'memoize', just like you want to distinguish between 'Dan' and 'Dan Abramov'

This is fair but you can also think it's short for "memoizing component" or something like this. In practice I don't think it's a problem.

Take "props". It also has a completely different meaning in English. :-) React uses it as short for "properties". Or take "render". React doesn't actually render during the "render" function!

So I understand why it might seem a bit odd at first. But another important consideration is to make it feel "light". This is an optimization API, and it shouldn't feel like a hassle to type. It's weird but that's how people think about APIs. Shorter words "feel" lighter which is what we want to express here.

So we think "memo" will work fine.

>

maciejsikora:
I feel changing the name to memo is very good approach, but what problem I see is that from now we will have two the same purpose concepts - PureComponent and memo, where their naming give a feeling that we deal with totally different things. This can be very misleading. Following this, my question is - does React team have a plan to change PureComponent name ( MemoComponent ... oh dont thing so ) or totally deprecate it, as I don't see a place for extending if we can compose by memo HOC.

maciejsikora:
I feel changing the name to memo is very good approach, but what problem I see is that from now we will have two the same purpose concepts - PureComponent and memo, where their naming give a feeling that we deal with totally different things. This can be very misleading. Following this, my question is - does React team have a plan to change PureComponent name ( MemoComponent ... oh dont thing so ) or totally deprecate it, as I don't see a place for extending if we can compose by memo HOC.

maciejsikora:
I feel changing the name to memo is very good approach, but what problem I see is that from now we will have two the same purpose concepts - PureComponent and memo, where their naming give a feeling that we deal with totally different things. This can be very misleading. Following this, my question is - does React team have a plan to change PureComponent name ( MemoComponent ... oh dont thing so ) or totally deprecate it, as I don't see a place for extending if we can compose by memo HOC.

gaearon:
>Following this, my question is - does React team have a plan to change PureComponent name ( MemoComponent ... oh dont thing so ) or totally deprecate it, as I don't see a place for extending if we can compose by memo HOC.

Not immediately but it's plausible that in the longer term `memo` will be the primary way to do this.

gaearon:
We decided to allow to pass any component type to `memo`, and not just functions. Without this, the composition of `memo`, `lazy`, and/or `forwardRef` seemed very confusing.

PR #64, opened 2018-10-19
React.lazy()

gaearon:
### [View formatted RFC](https://github.com/reactjs/rfcs/blob/gaearon-patch-2/text/0000-lazy.md)

**Note: I just wrote up the RFC. The semantics are designed by @sebmarkbage and @acdlite.**

### DIFF
--- a/text/0000-lazy.md
+++ b/text/0000-lazy.md
@@ -7,6 +7,10 @@ Start Date: 2018-10-19
 
 `React.lazy` adds first-class support for code splitting components to React. It takes a module object and returns a special component type.
 
+>Note
+>
+>This RFC is intentionally scoped to supporting default imports. We may in the future submit another RFC concerning named imports. In the meantime you can also implement support for named imports in userland with a lower-level Suspense API (to be discussed in another RFC).
+
 # Basic example
 
 ```js 
@@ -68,7 +72,7 @@ The second part of the question is why don't we still _allow_ passing something 
 const Button = lazy(async () => {
   const Components = await import('./components');
   // Resolve to named export:
-  return components.Button;
+  return Components.Button;
 });
 ```
 
--- a/text/0000-lazy.md
+++ b/text/0000-lazy.md
@@ -54,9 +54,9 @@ and not like this:
 
 ```js
 // Annoying and confusing:
-const Button = lazy(() => import('./Button').then(Button => Button.default);
+const Button = lazy(() => import('./Button').then(Button => Button.default));
 // Named imports don't make this better:
-const Button = lazy(() => import('./Button').then(Button => Button.Button);
+const Button = lazy(() => import('./Button').then(Button => Button.Button));
 ```
 
 (Note this doesn't mean you're forced to use default exports for *all* your components. Even if you primarily use named exports, consider default exports to be "async entry points" into just the components you want to code split.)
@@ -73,6 +73,17 @@ const Button = lazy(async () => {
 ```
 
 We **intentionally** don't support this in the scope of this proposal. This gives React an opportunity to integrate more tightly with the module system on the server in the future. There are no proposed standards for this yet, but the new Suspense-capable React server renderer we'll soon be working on will benefit from having a chance to introspect the `import()` result directly (rather than just a component). For example, it could change a priority of a pending request for the `Button.js` module once it knows that a component of the `Button` type is going to be lazily rendered. We can't do this if we break the link between the module and the component. While this depends on future experimentation and standardization work, this design leaves more space for it. We can always remove this restriction later if it doesn't end up being beneficial. (That's also why the proposal doesn't support arbitrary Promises as component types.)
+
+### Named Exports
+
+While named exports aren't currently supported by this proposal, it doesn't exclude them in the future. For example:
+
+```js
+// Not a part of this RFC but plausible in the future
+const Button = lazy(() => import('./components'), components => components.Button);
+```
+
+This is out of scope of the current initial proposal but could be added later.
 
 # Drawbacks
 
--- a/text/0000-lazy.md
+++ b/text/0000-lazy.md
@@ -1,0 +1,98 @@
+
+Start Date: 2018-10-19
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+`React.lazy` adds first-class support for code splitting components to React. It takes a module object and returns a special component type.
+
+# Basic example
+
+```js 
+import Input from './Input'; // Regular import
+const Button = React.lazy(() => import('./Button')); // Dynamic import
+
+function Dialog() {
+  // You can mix normal and lazy-loaded components in one tree
+  return (
+    <form>
+      <Input />
+      <Button />
+    </form>
+  );
+}
+```
+
+The first time `<Button />` is rendered, it would trigger the dynamic `import` which would start loading the code. When the `Button` code has loaded, React would resume rendering.
+
+Note that this **wouldn't** immediately render a spinner or an empty space instead of the `<Button />` like userland solutions would do today. Instead, the render would get *suspended*. You would have full control over how far above in the tree to put the loading indicator, as well as an opportunity to skip the indicator altogether on fast connections. This is a part of the "Suspense" feature umbrella. The RFC specifying how Suspense itself works exactly is not published yet, but you can find a demonstration in the [second half of this talk](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html). Specifying the exact mechanics of Suspense is orthogonal to this RFC and doesn't need to block it.
+
+# Motivation
+
+Code splitting is one of the most effective ways to reduce the size of client-side code. It is achievable in React today but it requires either managing state manually or using a library that does it for you. However, even with either of these solutions, the typical user experience isn't ideal.
+
+Existing code splitting solutions tend to put the loading indicator (or a "hole") directly in place of the loaded component. As a result, you might end up with a cascade of spinners and "holes" as different leaf components are being loaded. Both spinners and "holes" make the loading sequence feel more janky, and cause unnecessary layout work for the browser, making the app load slower. They are hard to orchestrate and handle in a visually consistent way so often people don't use code splitting at all, or only reserve it for the top-level components. However, leaf components (e.g. a complex text input) benefit from code splitting too.
+
+The goal of this API is to make it easy to code split any particular component regardless of whether it's closer to the root or to the leaves. Adding code splitting to a component shouldn't require you to restructure your components or their data flow. The Suspense API (to be discussed in a separate later RFC) would let you declaratively specify where to display the loading indicator, and use it both for code splitting and data fetching. React would make sure that only intentional loading states are displayed, and that async component boundaries aren't creating visual noise in the app loading sequence.
+
+# Detailed design
+
+`React.lazy` accepts a Promise factory, and returns a new component type. When React renders that type for the first time, it triggers the Promise factory (thus, in case of dynamic `import`, starting the request). If the Promise is fulfilled, React reads the `.default` value from it (assuming the resolved value is a module object), and uses it as a component type for rendering. If the Promise is rejected, the rejection is handled in the same way as React normally handles errors (by letting the nearest error boundary handle it). After the code has loaded, React caches the Promise result. Next renders of the components with this type become synchronous and have no extra cost.
+
+## Why the `.default` Field?
+
+You might be wondering why it reads the `.default` field from the Promise result rather than, for example, assume that the Promise result _itself_ is a component type. There are two reasons.
+
+We propose to read `.default` automatically so that we can write code like this:
+
+```js
+const Button = lazy(() => import('./Button'));
+```
+
+and not like this:
+
+```js
+// Annoying and confusing:
+const Button = lazy(() => import('./Button').then(Button => Button.default);
+// Named imports don't make this better:
+const Button = lazy(() => import('./Button').then(Button => Button.Button);
+```
+
+(Note this doesn't mean you're forced to use default exports for *all* your components. Even if you primarily use named exports, consider default exports to be "async entry points" into just the components you want to code split.)
+
+The second part of the question is why don't we still _allow_ passing something without a `.default` property?
+
+```js
+// Why don't we want to support this though?
+const Button = lazy(async () => {
+  const Components = await import('./components');
+  // Resolve to named export:
+  return components.Button;
+});
+```
+
+We **intentionally** don't support this in the scope of this proposal. This gives React an opportunity to integrate more tightly with the module system on the server in the future. There are no proposed standards for this yet, but the new Suspense-capable React server renderer we'll soon be working on will benefit from having a chance to introspect the `import()` result directly (rather than just a component). For example, it could change a priority of a pending request for the `Button.js` module once it knows that a component of the `Button` type is going to be lazily rendered. We can't do this if we break the link between the module and the component. While this depends on future experimentation and standardization work, this design leaves more space for it. We can always remove this restriction later if it doesn't end up being beneficial. (That's also why the proposal doesn't support arbitrary Promises as component types.)
+
+# Drawbacks
+
+* You can't render a `React.lazy` component with the current server renderer implementation because it doesn't support suspending. (That's the case for all Suspense features so it's not unique to this proposal.)
+* Reading `.default` only can be annoying to teams that have settled on only ever using named exports.
+
+# Alternatives
+
+* Keep implementing this manually (status quo).
+* Don't implement this at all until the Suspense-capable server renderer is done.
+* Call it something long like `createLazyComponent`.
+* Support Promises as component types directly.
+* `lazy` renders a "hole" or lets you specify an indicator as part of its API.
+* `lazy` doesn't try to read the `.default` export.
+* `lazy` reads the `.default` export but also works with Promise resolving to a component directly.
+
+# Adoption strategy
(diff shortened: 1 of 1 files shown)


jquense:
Is the idea that the return value of `lazy()` MUST return a promise that resolves to `{default: Component}`? The downsides of that strike me as out weighing the convenience of having write `lazy(async () =>  (await import('./Button')).default)` or the same in plain promises. React almost always prefers explicit longer code to convenience so it seems a bit of an odd choice out here to require `.default` for what is probably(?) code that will be written not super often.

gaearon:
>Is the idea that the return value of lazy() MUST return a promise that resolves to {default: Component}? 

In the scope of this proposal, yes.

>The downsides of that strike me as out weighing the convenience of having write lazy(async () => (await import('./Button')).default) or the same in plain promises. 

Which downsides are you referring to?

>React almost always prefers explicit longer code to convenience so it seems a bit of an odd choice out here to require .default for what is probably(?) code that will be written not super often.

I think the RFC goes into detail about why we specifically want access to the *module* object. This will be important for future work on the new server renderer.

An "explicit" proposal that doesn't lose this future potential would be something like

```js
lazy(() => import('./Button'), Button => Button.default);
```

At this point it's going to be so common though that it might as well be the default (pun intended) behavior.

jquense:
> Which downsides are you referring to?

the inability to use named exports, or potentially do any post-import transform/instrumentation

> I think the RFC goes into detail about why we specifically want access to the module object.

Sorry yeah, I can see how that would be useful, but it also feels premature? As you noted what the module object returned here even looks like is nebulous and undefined, even in Node for the SSR case. This limitation doesn't seem to really put react in a better position for the future since there is no reason why users can't do `import('./Buttons').then(_ => ({ default: _.SaveButton })`. Potentially breaking any future deeper integrations.

I would also add that it feels uncomfortable to me that React would want integrate so tightly into the module space, before there is a clear sense of what the standard there is. Folks do so. many. weird things in webpack, et al, to meet their needs that this is going to be a moving target for a looong time.

jquense:
(do I think this a really cool addition to the API!)

gaearon:
>the inability to use named exports, or potentially do any post-import transform/instrumentation

The proposal mentions we *could* add named exports to this.

>Sorry yeah, I can see how that would be useful, but it also feels premature?

We're starting work on this now (both standard proposals and the new server renderer implementation) so I wouldn't say it's premature. This is going to become a very active area of work for us.

>This limitation doesn't seem to really put react in a better position for the future since there is no reason why users can't do import('./Buttons').then(_ => ({ default: _.SaveButton }).

We'll document that you're supposed to pass the module object, and nothing else. For sure, some people will ignore it, but then adopting the new server renderer isn't forced upon them. People who follow the recommendation would have an easier path.

Again, if this doesn't work out, we can surely relax it in the future. In fact that's what the original proposal was (it took any Promise) but then we decided to make it stricter.

mellogarrett:
```
// Why don't we want to support this though?
const Button = lazy(async () => {
  const Components = await import('./components');
  // Resolve to named export:
  return components.Button;
});
```
I think the return statement should have a capital 'C'?

text/0000-lazy.md
```diff
@@ -0,0 +1,109 @@
...
+(Note this doesn't mean you're forced to use default exports for *all* your components. Even if you primarily use named exports, consider default exports to be "async entry points" into just the components you want to code split.)
+
+The second part of the question is why don't we still _allow_ passing something without a `.default` property?
+
+```js
+// Why don't we want to support this though?
+const Button = lazy(async () => {
+  const Components = await import('./components');
+  // Resolve to named export:
+  return components.Button;
```
mAAdhaTTah: Small typo -> `Components.Button` (needs capital C, or lowercase on line 69).

sebmarkbage:
I think that once some of the other pieces of this puzzle fall into place. I suspect that the syntactical overhead will strongly push people towards the default exports. Especially if we can even get rid of the call to “lazy” in the default export case.

pomber:
Nice (I'm already using it https://github.com/pomber/code-surfer-editor/blob/master/src/LazyCodeEditor.js#L4 😁). 
Do you have in mind any special API for preloading? Or is the idea to preload it like any other suspenseful component (rendering it, maybe hidden)?

matt-casey:
In the alternatives, it's listed as a possibility to:
> Call it something long like createLazyComponent.

Has there been any more public discussion on the naming? Outside of the context of this discussion,`React.lazy` wouldn't necessarily be clear or self-documenting in the same way the rest of the React API is. 

For example, the functions dealing with refs say so explicitly (createRef, forwardRef).
The functions dealing with components say so explicitly (createElement, cloneElement, isValidElement).

Since this probably isn't used in too many places throughout a code base, I think the tradeoff of adding extra characters for the sake of explicitness might be worth considering.

styfle:
On the topic of **default exports**...
Here are a list of reasons to avoid default exports (via @basarat)

https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

I think most of the React community uses default exports however many, following the practices above, do not use default exports and prefer named exports.

text/0000-lazy.md
```diff
@@ -0,0 +1,109 @@
...
+const Button = lazy(() => import('./Button'));
+```
+
+and not like this:
+
+```js
+// Annoying and confusing:
+const Button = lazy(() => import('./Button').then(Button => Button.default));
+// Named imports don't make this better:
+const Button = lazy(() => import('./Button').then(Button => Button.Button));
```
TrySound: This is better
```js
const Button = lazy(() => import('./Button').then(module => module.Button));
```

sebmarkbage:
The longer term vision includes making literally every component in the whole code base use this. So you can imagine that syntax overhead will be a big deal and familiarity counter balance any explicitness concerns.

text/0000-lazy.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+(Note this doesn't mean you're forced to use default exports for *all* your components. Even if you primarily use named exports, consider default exports to be "async entry points" into just the components you want to code split.)
```
TrySound: This adds inconsistency in modules style. The point of component code splitting to be invisible. You take any module, wrap it with `lazy` and it becomes loadable. Converting to `default` export is additional and wrong action. We don't need to define that module as asynchronous. Module should be only consumed as asynchronous in place.

jquense:
> The longer term vision includes making literally every component in the whole code base use this

I have a hard time imagining how this will place nicely with bundlers and efficient code splitting...at the moment that would create a new bundle per file in webpack no?

text/0000-lazy.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Named Exports
+
+While named exports aren't currently supported by this proposal, it doesn't exclude them in the future. For example:
+
+```js
+// Not a part of this RFC but plausible in the future
+const Button = lazy(() => import('./components'), components => components.Button);
```
TrySound: This is nice, but we need something right now. Would be good to add at least an ugly example.

```js
const Button = lazy(() =>
  import('./components').then(module => ({ default: module.Button }))
);
```

sebmarkbage:
*at the moment* yes

matt-casey:
> The longer term vision includes making literally every component in the whole code base use this.

I might be able to see the utility in that if you have a massive team working on components in a distributed fashion, where what code is being rendered for any given user might be impossible to reason about.

For an average sized site, though, the extra latency of loading each component bundle individually would surely outweigh whatever gain you have from a faster initial page load.

Maybe in an HTTP/2 world?

jquense:
practical named exports example btw Material UI exports a lot of components as named exports, if it makes sense maybe it's worth including the extension now? I also seethe value in limiting new API's initially

text/0000-lazy.md
sebmarkbage: I think maybe we’ll try to warn against that pattern since it will break future work

salzhrani:
When rehydration is possible, there needs to be away to tell the client to eagerly load components that were loaded lazily on the server and/or prime the lazy components cache before the initial render.
maybe a provider like api that can be queried to provide such info, and given the injected data to prime the cache

streamich:
Trying to find a component in `.default` seems totally arbitrary. It should just use the resolved value as a component. Or maybe check `.default` afterwards if the resolved value is not a component.

GarethSmall:
If I place outside of Suspense I get:

`An update was suspended, but no placeholder UI was provided.`

I haven't seen any examples that note this, is this expected?

https://codesandbox.io/s/n731ny9kjl

text/0000-lazy.md
```diff
@@ -0,0 +1,109 @@
...
+
+```js
+const Button = lazy(() => import('./Button'));
+```
+
+and not like this:
+
+```js
+// Annoying and confusing:
+const Button = lazy(() => import('./Button').then(Button => Button.default));
```
streamich: Does not have to be written like that. Could be:

```js
const Button = lazy(async () => (await import('./Button')).default);
```

or even:

```js
const Button = lazyDefault(import('./Button'));
```

j-f1:
Can you explain how Suspense will be able to make use of the actual module record?

GarethSmall:
Here's an example: https://codesandbox.io/s/yx95qp06v

text/0000-lazy.md
milesj: It most definitely needs to be a function that returns a promise, so the last 3 examples are invalid.

streamich: @milesj thx, corrected

gabemeola: it'll need the intermediate function so the dynamic import isn't immediately resolved.

milesj: Yup, without the function it's a side-effect.

j-f1:
@GarethSmall Were you replying to me in your first comment? If so, how would the behavior be different if Suspense couldn’t access the module record, just the component?

GarethSmall:
@j-f1 Suspense  doesn't access the module record, Suspense is expecting a component or at least the promise of a component.

jquense:
@GarethSmall that behavior is changing, that some of the point of the RFC here

GarethSmall:
@j-f1 Do you mean the lazy function or the Suspense component?

j-f1:
@GarethSmall I mean the `React.lazy` function, but the RFC says the module record will be useful to Suspense.

GarethSmall:
@j-f1 Ahh okay, I apologize I mis-spoke from a lack of understanding, so what I was doing was a bad practice, here's an expert from the RFC:

Is this what you were talking about? 

The second part of the question is why don't we still allow passing something without a .default property?

```
// Why don't we want to support this though?
const Button = lazy(async () => {
  const Components = await import('./components');
  // Resolve to named export:
  return components.Button;
});
```

We intentionally don't support this in the scope of this proposal. This gives React an opportunity to integrate more tightly with the module system on the server in the future. There are no proposed standards for this yet, but the new Suspense-capable React server renderer we'll soon be working on will benefit from having a chance to introspect the import() result directly (rather than just a component). For example, it could change a priority of a pending request for the Button.js module once it knows that a component of the Button type is going to be lazily rendered. We can't do this if we break the link between the module and the component. While this depends on future experimentation and standardization work, this design leaves more space for it. We can always remove this restriction later if it doesn't end up being beneficial. (That's also why the proposal doesn't support arbitrary Promises as component types.)

j-f1:
Thanks for explaining; that makes a lot more sense.

aweary:
How does this interact with `React.memo`? Would this work?

```jsx
const Button = React.memo(React.lazy(
    () => import('third-party-button')
);
```

gaearon:
No. The argument to `memo` has to be a function component, which `lazy` isn't (it is a special type). We should error on this.

It works the other way around though. Which is consistent with what we say in `memo` RFC: it should be applied at the definition site. You can wrap in your own component at the call site if you *really* need this.

text/0000-lazy.md
gaearon: Why do you feel so strongly against writing a single line of code that does the export?

aweary:
> The argument to memo has to be a function component, which lazy isn't (it is a special type). We should error on this.

Is this also true for `React.forwardRef` then?

text/0000-lazy.md
gaearon: The motivation for this is explained just below.

text/0000-lazy.md
gaearon: The motivation for why it's not this way is explained just below. Did you get a chance to read it?

gaearon:
>For an average sized site, though, the extra latency of loading each component bundle individually would surely outweigh whatever gain you have from a faster initial page load.

@matt-casey 

I think you might be confusing the *syntax* (import is asynchronous) with a specific *behavior* (create a bundle per component). There's absolutely no reason why tools like webpack need to create a separate bundle per dynamic import. If they had access to usage statistics, they could instead be smart about bundling modules together in the most efficient way based on which are likely to be rendered together. 

The "longer term" vision Sebastian alluded to is not about the current webpack behavior. But about allowing a better bundling solution.

gaearon:
**Regarding named exports.**

I hear where you’re all coming from. If you disagree with our recommendation to use default exports for split points, you can use a workaround described in https://github.com/reactjs/rfcs/pull/64#discussion_r226683684. You could even your own `lazyNamed` helper that does this. Yes, it’s hacky and it might make it more difficult for you to benefit from the future work on the streaming rendering. But it seems like this isn’t a concern shared by many people on the thread. This makes sense since it applies to the future and not to today. And because details are very vague at this point (we’re just beginning to investigate this). We have to think about the future though because otherwise we’ll have to make breaking changes later.

This RFC has a limited surface. It's targeted at the main use case we want to start handling (default imports). There is a door open for supporting named imports in future RFCs, but it is out of scope of this one. This one is intentionally very limited. We expect to understand this problem space more within the next several months and then we can revisit that discussion with a better understanding.

I understand you want us to support _your_ use case today. But we don’t have all the information yet, and if the solution is too broad, it’s very likely we’ll have to deprecate or make breaking changes to it in the next release. Nobody likes breaking changes. Therefore, for now we’d like to focus it on the parts we feel more confident in.

I hope this makes sense, even if you disagree.

matt-casey:
> The "longer term" vision Sebastian alluded to is not about the current webpack behavior. But about allowing a better bundling solution.

@gaearon 
I think that makes sense, I was thinking very much in the context of today's webpack.

I can definitely imagine having everything lazy-loaded if the compiler were smart enough. It would be great to move that decision of where to code split out of the code-base itself and put it into the compiler's hands.

I realize now that in order to give the compiler the flexibility to do that, you'd have to do exactly what @sebmarkbage said and wrap `literally every component` in `React.lazy`.

text/0000-lazy.md
theKashey: Right now `default` exports are _discouraged_. There are a lot of reasons behind:
- like `allowSyntheticDefaultImports:false` in typescript
- like autoimports in IDE
- like auto "naming" things

Why not to support basic "babel" interop, ie use `.default` if `_esModules` set, mimicking how "real"(babel/webpack) imports work?

I also prefer to have a default export on code splitting point, as some sort of an "interface", but it should not be the law.

gaearon: Babel interop has made the ESM ecosystem very messy and complex (even though it was instrumental in getting people to adopt ESM — perhaps too early). We don’t wait to add more tooling dependency on this behavior because it’s super hard to fix properly.

>I also prefer to have a default export on code splitting point, as some sort of an "interface", but it should not be the law.

 See https://github.com/reactjs/rfcs/pull/64#issuecomment-431507924.

gaearon:
>When rehydration is possible, there needs to be away to tell the client to eagerly load components that were loaded lazily on the server and/or prime the lazy components cache before the initial render.

Yes. This is something we want to be able to do automatically, which is why we need the module information (rather than just the Promise to a component). Then the server could send this information to the client embedded in the small runtime.

theKashey:
It's not clear how server-side rendering works. To be more concrete - _when_ component got imported.
It is not a question on a client side - all "userspace" loaders execute `import` function on Component render, and `lazy` will do the same. This is right and clear - import something only when you used it.

But on server side, it's very opinionated:
- loadable-components _waits_ for all components to be loaded, to do this library traverse react-tree by it's own
- universal-component uses babel plugin to convert async `imports` to sync `requires` but still require some stuff only on component render
- imported-components imports everything on HOC definition, so by the time server started - everything is already imported, and there are no actions to be made later. This also enabled some features, like per-connection mocking([here is](https://medium.com/@antonkorzunov/ssr-dependency-mocking-is-the-answer-d8d8c371aa94) an explanation why one need it), which could be very useful to make rendering more server-side friendly.

So - _when_ imports will work on a server?

gaearon:
@theKashey 

Supporting _today's_ server renderer is out of scope of this proposal (as mentioned in the RFC).

In the future Suspense-capable server side renderer (which is fully asynchronous and can stream chunks while waiting for data) my mental model is that dynamic imports would be somehow rewritten to be synchronous, but the renderer would remember which modules got rendered, so that it can increase their priority as they're being sent to the client interleaved with data and HTML.

shawnmitchell:
This is almost certainly outside the scope of this RFC but I wanted to see if there was any interest in lazy imports as a codebase security mechanism. What I'm thinking of is a paradigm where the code is stored in the cloud (Firebase Storage or the like) and streamed to and loaded by the client contingent on some separate backend user authentication.

gaearon:
Going with this limited version (`default` imports only) for now. It's possible we'll add named imports later.

text/0000-lazy.md
Avi98: i think its line 71

PR #65, opened 2018-10-19
contextType: convenience API for reading context in a class

gaearon:
### [View formatted RFC](https://github.com/reactjs/rfcs/blob/gaearon-patch-3/text/0000-contexttype.md)

**Note: I just wrote up the RFC. The semantics were designed by @sebmarkbage.**

### DIFF
--- a/text/0000-contexttype.md
+++ b/text/0000-contexttype.md
@@ -1,0 +1,94 @@
+Start Date: 2018-10-19
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+`contextType` is a convenience API for accessing the React [context](https://reactjs.org/docs/context.html) from class components. It works similar to `Context.Consumer` but lets you handle the common case where you only need a single context more ergonomically.
+
+It **does not replace** `Context.Consumer`. It is a shortcut for the most common case that helps reduce component tree nesting.
+
+# Basic example
+
+```js 
+// Works with new context API
+const ThemeContext = React.createContext('dark');
+
+class Button extends React.Component {
+  static contextType = ThemeContext; // Alternative to using Consumer
+
+  componentDidMount() {
+    console.log(this.context); // You can read it in lifecycles
+  }
+
+  render() {
+    const theme = this.context; // And in render
+    return <div className={theme} />;
+  }
+}
+```
+
+This component subscribes to context similar to `ThemeContext.Consumer`, but lets you avoid an extra nesting layer when you want to access context from a lifecycle method.
+
+# Motivation
+
+The new context API introduced in 16.3 is a low-level primitive. It was designed to work the same way both with class components and with function components, and it solves name clashes and other issues with the legacy context API.
+
+However, the legacy context API is still more ergonomic for the case when you need to access context in a lifecycle method. With the render prop context API, you can still do it but it is awkward:
+
+```js
+class ThemedButton extends React.Component {
+  componentDidMount() {
+    console.log(this.props.theme);
+  }
+
+  render() {
+    return <div className={this.props.theme} />;
+  }
+}
+
+const Button = React.forwardRef((props, ref) => (
+  <ThemeContext.Consumer>
+    {theme => <ThemedButton {...props} theme={theme} ref={ref} />
+  </ThemeContext.Consumer>
+);
+```
+
+This particular case (when we need to access a single context from lifecycle methods) is sufficiently common compared to other cases that it's worth building a convenience API for. It is also valuable for some [interop](https://github.com/facebook/react/issues/13336) scenarios.
+
+Importantly, this API is designed to speed up the adoption of the new context. Supporting legacy context makes React larger and slower, and we see the need for more pragmatic solutions to help people move away for it.
+
+>Note
+>
+>This isn't the first "convenience API" in React. It is similar to how React has a convenience `createRef` API even though a more powerful callback ref API exists. The goal is to help simplify the 80% case with a more direct API, at the cost of some redundancy.
+
+# Detailed design
+
+When a class component has `contextType` defined, it is assumed to be a Context object (return value of `React.createContext`). React flags this component as a context consumer, and provides the current value of this context as `this.context`. It is available both in the render phase and in the lifecycle methods.
+
+## Why Didn't We Do This Way Back?
+
+Arguably this could have been a part of the initial `React.createContext` proposal. We'd still, however, need a low-level API for function components or combining multiple contexts. So this couldn't be the only API. At the time `React.createContext` was proposed, we were hoping to be able to make more targeted optimizations to context propagation. However, as we worked on Suspense, we realized that we need the capability to read context without an explicit `Consumer` render prop anyway — otherwise accessing a Suspense cache is too cumbersome. Since this makes the original performance optimization justification weaker, and we might as well add a convenience API if we can't take advantage of the low-level API constraints.
+
+This API proposal is intentionally limited and **doesn't** support multiple contexts directly. It is only a shortcut for the common case. If you need to read several contexts (or read it from function components), you can use the existing API.
+
+# Drawbacks
+
+* Adds another way to do the same thing. (But without the nesting.)
+* Might be confused with the legacy API due to the similarity.
+* Doesn't let you use multiple contexts without a render prop.
+* Code using this API is more hassle to move into a function component.
+
+# Alternatives
+
+* Don't do anything (status quo). This significantly slows down the adoption of new context.
+* Call the field something different, like `contextConsumerType`.
+* Support using multiple contexts from one class component.
+
+# Adoption strategy
+
+You don't need to use it, but it's a nicer migration path from the legacy context. 
+
+# How we teach this
+
+https://github.com/reactjs/reactjs.org/pull/1265


pshrmn:
1.  `context.Consumer` is essentially a proxy for `Context`, correct? i.e. `static contextType` can be a consumer. I ask because it would be nice if packages can export their context consumer, but keep the provider private.
```js
import { Consumer } from "some-package";
  static contextType = Consumer;
}
```

2. `prevContext` was removed from `componentDidUpdate` in v16, with the recommendation to store the previous context value on the class instance. Will this be the same recommendation for `contextType` or will `prevContext` be brought back?
```js
  static contextType = context;
  componentDidMount() {
    this.prevContext = this.context;
  }
  componentDidUpdate() {
    if (this.context !== this.prevContext) {
      // ...
    }
  }
}
```

jquense:
ya this is awesome! supporting multiple contexts would be really nice as well, but i see how that would be harder. I would say that if it was feasible it may be worth doing, our apps are more and more looking like `withFoo(withBar(Component))` for relay, react-router, react-intl etc.

AlexGalays:
I'm in favor of not doing anything.

- Aren't classes slowly going to get phased out?

- subscribing to only one context is actually easy with the <Consumer/> component. Subscribing to multiple contexts is what's annoying in vanilla react but this spec doesn't change that.

- We can easily write a `connectToContexts([Context1, Context2], (value1, value2) => ({ value1, value2 })`  kind of helper today.

- This seems difficult to type without React specific hacks in flow/TS.

gaearon:
>context.Consumer is essentially a proxy for Context, correct? i.e. static contextType can be a consumer. I ask because it would be nice if packages can export their context consumer, but keep the provider private.

No, we discourage using Consumer for anything other than a render prop. It accidentally is the same object as context today but we’re going to start warning about this. I’ll write up a better explanation next week for why this makes sense. But for now please export Context from libraries directly. 

>prevContext was removed from componentDidUpdate in v16, with the recommendation to store the previous context value on the class instance. Will this be the same recommendation for contextType or will prevContext be brought back?

This `componentDidUpdate` argument is already taken by the snapshot so I don’t think we’ll bring it back. Storing it manually still works though.

>our apps are more and more looking like withFoo(withBar(Component)) for relay, react-router, react-intl etc.

We have something in the works that helps with this. Will likely post an RFC next week.

>Aren't classes slowly going to get phased out?

Even if classes become less relevant we still plan to support them for the observable future. Speeding up the adoption of new context is a high priority which makes this worth the redundancy compromise to us.

>This seems difficult to type without React specific hacks in flow/TS.

I don’t see why, can you explain more?

jfo84:
If I'm understanding correctly, you would no longer need to initialize Context outside the class with this change. Is that correct or is that considered an anti-pattern?

text/0000-contexttype.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Why Didn't We Do This Way Back?
+
+Arguably this could have been a part of the initial `React.createContext` proposal. We'd still, however, need a low-level API for function components or combining multiple contexts. So this couldn't be the only API. At the time `React.createContext` was proposed, we were hoping to be able to make more targeted optimizations to context propagation. However, as we worked on Suspense, we realized that we need the capability to read context without an explicit `Consumer` render prop anyway — otherwise accessing a Suspense cache is too cumbersome. Since this makes the original performance optimization justification weaker, and we might as well add a convenience API if we can't take advantage of the low-level API constraints.
+
+This API proposal is intentionally limited and **doesn't** support multiple contexts directly. It is only a shortcut for the common case. If you need to read several contexts (or read it from function components), you can use the existing API.
```
blainekasten: So excited for this proposal! Can you explain why this limit is here?

AlexGalays:
> If I'm understanding correctly, you would no longer need to initialize Context outside the class with this change. Is that correct or is that considered an anti-pattern?

This change is just a shortcut for <Consumer />; you still need a matching <Producer/> somewhere above.

strayiker:
It's great, but!

```javascript
  static contextType = ThemeContext; // Put this here

  render() {
    const theme = this.context; // Then you can use this here
    ...
  }
}
```
magic

Binding the context through static field always perceived as something alien in React. I hope there is less strange way to do this. E.g.:
```javascript
class Button extends ThemeContext.Component { // <---
  render() {
    const theme = this.context;
    ...
  }
}
```

jfo84:
@AlexGalays Okay got it. Still a big fan of the change

mAAdhaTTah:
@strayiker The API for multiple contexts would be much worse if you did it w/ inheritance.

strayiker:
@mAAdhaTTah Totally agree. But I assume that this RFC for simplest case with single context.

cangoektas:
Sounds really cool! I'd also love to see support for multiple contexts if we can agree on an intuitive API.

What about something like this?

```jsx
  static contextTypes = {
    theme: ThemeContext,
    route: RouteContext
  }

  render() {
    const theme = this.context.theme;
    const route = this.context.route;
  }
}
```

TrySound:
@cangoektas `contextTypes` is reserved by the old context api

pshrmn:
> No, we discourage using Consumer for anything other than a render prop. It accidentally is the same object as context today but we’re going to start warning about this. I’ll write up a better explanation next week for why this makes sense after I get back from React Conf. But for now please export Context from libraries directly.

I'm probably missing some context :smirk: that you'll cover in your explanation. My understanding of the `context`/`context.Consumer` parity issue was (based on https://github.com/facebook/react/pull/13829) that the context could be used as an element and that properties could be chained off of the consumer a la `context.Consumer.Provider`.

As far as actual design goes, `this.context` would be read only (like the consumer), right? Again, I'm probably missing some context, but I personally would prefer for `static contextType` to be a consumer, not the full context.

text/0000-contexttype.md
gaearon: Static typing and performance reasons.

gaearon:
>I'd also love to see support for multiple contexts if we can agree on an intuitive API.

It would cause us to allocate intermediate objects on every render. This is too expensive and brings back some of the problems with the legacy context API. Such as making it difficult to type statically.

text/0000-contexttype.md
```diff
@@ -0,0 +1,94 @@
...
+
+* Adds another way to do the same thing. (But without the nesting.)
+* Might be confused with the legacy API due to the similarity.
+* Doesn't let you use multiple contexts without a render prop.
+* Code using this API is more hassle to move into a function component.
+
+# Alternatives
+
+* Don't do anything (status quo). This significantly slows down the adoption of new context.
+* Call the field something different, like `contextConsumerType`.
```
aweary: I wonder how many people will try and use `PropTypes` to define `contextType` thanks to `contextTypes`. Maybe it should use a name that is even further from `contextTypes` so that there's no risk of confusing them?

text/0000-contexttype.md
```diff
@@ -0,0 +1,94 @@
...
+# Drawbacks
+
+* Adds another way to do the same thing. (But without the nesting.)
+* Might be confused with the legacy API due to the similarity.
+* Doesn't let you use multiple contexts without a render prop.
+* Code using this API is more hassle to move into a function component.
+
+# Alternatives
+
+* Don't do anything (status quo). This significantly slows down the adoption of new context.
```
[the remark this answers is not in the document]
gaearon: Accessing context in lifecycles and the need to forward refs to do that. It's awkward.

text/0000-contexttype.md
gaearon: ¯\\_(ツ)_/¯

gaearon: We warn immediately if you pass something else.

theKashey:
`observedBits` is no longer a thing?

gaearon:
It still is but also still unstable. Not an official API and we don't know yet if it will be official.

This particular RFC doesn't include support for it, you'd need to still  use `Consumer`.

sebmarkbage:
That’s a good point

sebmarkbage:
@pshrmn Thats good feedback. One consideration is that we don’t want to necessarily keep too many concepts and object allocations around by default. It’s also a common useful pattern to allow shadowing contexts to be able to work around issues.

As a library author I’m sympathetic to constraining it in certain use cases too. So if we start with the full context we still have the option to add the ability to pass a Consumer too. That way you can just think of the whole context as a thing but libraries can expose write only versions of those things.

So I think we can actually get what you want without blocking this RFC as a first step.

dschinkel:
I want nothing to do with the context API.  Calling other stuff "legacy" is BS.  Not being able to pass int he store directly from my tests is complete BS. WTF @gaearon what are you and the React team doing over there?

dschinkel:
Ah so all the people who don't use shallow thumbs down my last post.  Got it.

So, are you going to try to phase out the ability to send store in directly to components in the future like Redux just did?  If you're thinking about it stop.  Many of us test with shallow specifically by passing props directly which is how we actually TDD.  And it's not just a matter of "find another way", it's going to break all our tests as well as we have many benefits we gain by passing in store from tests.

dschinkel:
BS: https://github.com/facebook/react/pull/13728 give me a break.

theKashey:
Could you please clarify - what exactly is _broken_ here? What does not work anymore or would be broken in a future?
Could you provide some code you are(were) happy about, and not happy after _this change_?

I also a [shallow-rendering follower](https://hackernoon.com/why-i-always-use-shallow-rendering-a3a50da60942), but I could not see any obstacles due to this RFC.

dschinkel:
I'm very concerned with the direction Facebook has been going in regards to ignoring isolated testing.  We're already seeing Redux make changes such as outlined here.  The changes that are going to cause friction for many devs are devs who are testing React components shallowly and it does not make sense to have to use something like provider in our tests.  But the way React is going with requiring stuff like Context API in the future, and possibly doing the same (I hope not) which is to prevent being able to pass a store as a prop from shallow tests really limits devs who want to still continue to test in isolation, small focused tests which are easy to maintain and where we do want to be able to pass in simple stores (dummies) for direct testing not through mount and the Provider API.

We have ways we test where we do not need that heaviness in our tests.  Vue supports shallow tests.  I just don't want to see React go so far that we can't shallow anymore.  It's already getting hard to shallow with promoting so many wrapping and nested components, as well as now we're starting to get totally coupled to having to use provider in our tests and mount.  We don't all want that.

theKashey:
Personally I am trying to solve this puzzle as a puzzle.
- _limited depth components_, you may confidently test with `mount`
- and _extend_ providing values for slots in the _integration component_, you may test with `shallow`

Today React broke your way to code and test, better to say - your taste or smell. But it's not about React, but about your own code, your way to write it (more about DI/mocking stuff than the code itself). 

The real problem here is only one, and it's not related to this issue, it just _how it is_ - enzyme is not a part of React, while behavior of react-test-renderer is not acceptable by majority. And does not contain shallow renderer at all (which is not a problem for _limited-depth component_).

PR #66, opened 2018-10-20
Update 0000-memo.md

fardarter:


gaearon:
Thanks for PR.

It doesn't matter what this argument is called in the scope of the RFC. It is *not named* so we can change it at any point in the future. We'll think more about names when documenting this feature, but it doesn't block consensus.

PR #74, opened 2018-10-29
Controlled type comparison

theKashey:
[View rendered RFC](https://github.com/theKashey/rfcs/blob/master/text/0000-compare-and-render.md)

React-Hot-Loader is not yet 16.6 compatible, and no more 16.7 compatible. It could not handle hooks, and would probably break on another stuff in a future.

This is a simple proposal to fix it - just let control type comparison, to stop faking types in a user space, as React-Hot-Loader usually does.

Already implemented in Inferno - https://github.com/infernojs/inferno/issues/1336 (logic is inverted)
Tested with Preact - https://github.com/developit/preact/pull/1120

### DIFF
--- a/text/0000-compare-and-render.md
+++ b/text/0000-compare-and-render.md
@@ -1,0 +1,99 @@
+- Start Date: 2018-10-28
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+During the reconciliation ReactDOM compares "old" rendered element, and a new one.
+If type or key does not match - the old element will be removed, and new one will took it's place.
+
+This makes stuff, like hot-reloading, a bit complex, and very flaky.
+
+Everything could be solved by a dev-time ability to control tree reconciliation, via a "compare-control" hook inside React-DOM internals.
+"compare-control" hook will be able to control the moment _type comparison_, telling React that
+ types are still the same, even if not equal by a reference, and let React-Hot-Loader rely on
+React internals, not to hack them.  
+
+# Basic example
+```js
+// default behavior - pick A if A and B the same
+React.compareControl( (typeA, typeB) => typeA === typeB ? typeA : null)
+
+// threat any components with the same name as the same
+React.compareControl( (typeA, typeB) => typeA.displayName === typeB.displayName)
+
+```
+# Motivation
+
+This change would be able to solve some React-Hot-Loader issues:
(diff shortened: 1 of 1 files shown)


text/0000-compare-and-render.md
```diff
@@ -0,0 +1,99 @@
...
+ types are still the same, even if not equal by a reference, and let React-Hot-Loader rely on
+React internals, not to hack them.  
+
+# Basic example
+```js
+// default behavior - pick A if A and B the same
+React.compareControl( (typeA, typeB) => typeA === typeB ? typeA : null)
+
+// threat any components with the same name as the same
+React.compareControl( (typeA, typeB) => typeA.displayName === typeB.displayName)
```
brigand: In one you return `?React.ComponentType` and in the other a boolean. Could you clarify?

theKashey: That's something I could not decide yet. Which type should be used - a new one, or an old one?
Right now - an old one, always. This is why it could work.
With controlled comparison - it could use a new one. That would mitigate some RHL side-effects.
But sometimes you may render an absolutely another type(proxied by RHL), moving "side-effect" a bit _deeper_ to realization, which also would fix https://github.com/gaearon/react-hot-loader/issues/304

gaearon:
We generally avoid global configuration like this.

I wonder if it's sufficient to have something like `React.lazy` but that can take an Observable of module values instead of a Promise. It would always use the latest module type for rendering, but switching the wrapped type to the next one would not reset the reconciliation.

theKashey:
That would work for lazy, memo or forwardRef, as long as it's possible to maintain ref equality, and change only `type` it refers to. But not for SFC or Classes - they just simple "direct" refs.

- It's possible to create new `container` type, to _hold_ ref to a component, but it would break `type comparison`.
- it's possible to use some sort of `wrapper`(as we always did), but it would break `go-to-code` feature of Dev Tools, __and__ `type comparison`. Plus updating the wrapper is a tricky task for HOCs and friends.
- moving `element.type comparison` into React does not break anything, plus makes "wrapper" updates an easy task - no need of rendering react-tree by hot-loader. 

This RFC was already implemented via webpack-loader to patch react-dom at build-time or separate package to make it simpler - that's literally 2 lines change in React-dom code - https://github.com/hot-loader/react-dom.

PR #85, opened 2018-11-11
RFC: Marker hook for server-side data

leops:
[View rendered](https://github.com/leops/rfcs/blob/master/text/0000-serializable-data-hook.md)

This RFC proposes a `useSerializable` hook to automate the extraction, and subsequent hydration of serializable server-side data.

### DIFF
--- a/text/0000-serializable-data-hook.md
+++ b/text/0000-serializable-data-hook.md
@@ -1,0 +1,197 @@
+- Start Date: 2018-11-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Introduce a new hook to allow extraction of server-side data and
+subsequent client-side hydration along the component tree.
+
+# Basic example
+
+```js
+// Usage with react-cache
+function ResourceProvider({ children, fetch, hash }) {
+    const resource = useSerializable(
+        // Create a "resource" object acting as a cache
+        // and manager for data requests
+        //
+        // The `data` argument is optional and may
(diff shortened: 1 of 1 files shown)


sebmarkbage:
*Some* way to transfer serializable data from server to the client is certainly in scope for the new server rendering infrastructure that supports Suspense. Unclear what the mechanism to do so is though.

leops:
This is why I tried to design this RFC in isolation, and leave as much as possible "unspecified", since I make no assumption on the specifics of how Suspense on the server will work. Essentially this is just the result of me experimenting with how an SSR framework might work with `React.lazy` by (poorly) polyfilling Suspense for the server, and trying to figure out a way to declaratively mark data for extraction / hydration. In the end I tried to make something that should "just work" unless the final API goes in a drastically different direction.

gaearon:
> the resulting resource object is stored in the hook state chain and returned (this part is essentially doing useMemo(createResource, [])).

This is pretty different from the direction we're going with the Suspense library design, which uses a centralized cache Map to hold data (rather than per-component memory). You can read more about this API here: https://github.com/reactwg/react-18/discussions/25.

There is still a question of whether we'd want the cache to be "transferable" via a built-in React API, or if you have to use Server Components (#188) and rely on transferring only the rendered output. We don't know how that will work yet. But it seems safe to say that the API shape will likely be different so I think we can close this more confidently. Appreciate the exploration though — it's just that we don't have much to say except that it'll likely be different.

PR #89, opened 2018-11-20
Context.write

acdlite:
***This is a very early stage proposal. I'm opening this RFC as a way to explain the problem space. We're still exploring alternative solutions. If this were an ECMAScript proposal it'd be Stage 0.***

## Summary

Proposes an extension to the context API for updating the default context value. This would allow for React-managed state that lives outside the UI tree and is shared across roots.

## Basic example

```js
const Context = React.createContext(initialValue, contextDidUpdate);
// Update the global context value across all roots. Any context consumer that
// is not wrapped in a Provider will re-render with this value.
Context.write(newValue);
// Functional updates also supported for access to the previous value.
Context.write(prevValue => newValue);

function contextDidUpdate(newContext) {
  // Optional callback that fires whenever context changes
}
```

### [Rendered text](https://github.com/acdlite/rfcs/blob/context-write/text/0000-context-write.md)

### DIFF
--- a/text/0000-context-write.md
+++ b/text/0000-context-write.md
@@ -1,0 +1,242 @@
+- Start Date: 2018-11-19
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Proposes an extension to the context API for updating the default context value. This would allow for React-managed state that lives outside the UI tree and is shared across roots.
+
+# Basic example
+
+```js
+const Context = React.createContext(initialValue, contextDidUpdate);
+// Update the global context value across all roots. Any context consumer that
+// is not wrapped in a Provider will re-render with this value.
+Context.write(newValue);
+// Functional updates also supported for access to the previous value.
+Context.write(prevValue => newValue);
+
+function contextDidUpdate(newContext) {
+  // Optional callback that fires whenever context changes
+}
+```
+
+# Motivation
+
+### Caching external data
+
+Where should a React app store data fetched over IO?
+
+It may be tempting to cache it in component state. This is a common pattern for pre-Suspense React apps. But there are some drawbacks. For example, if a Comment component depends on data from the server, consider the implications of storing that data in the Comment component's local state. What if the same comment data is rendered by a different component on another part the page? I shouldn't have to fetch the same comment twice. What if I navigate away and my comment is unmounted, but then I later navigate back? I don't want to refetch the same comment again merely because the old instance was unmounted. The underlying principle here is that data fetched over IO typically does not semantically align with the lifetime of a component used to present that data. The data and its presentation are separate concepts.
+
+Still, although caching in component state isn't ideal, it's arguably "good enough" for many use cases today.
+
+Not so with Suspense. The Suspense model is that the first time React renders an IO-bound component, an exception is thrown when attempting to read the data. While React is waiting for the data to load, it continues rendering the rest of the tree, but it doesn't commit the result; the partially completed tree is _discarded_. Once the data has resolved, React attempts to render the entire tree over again from scratch. It may, as an optimiztion, reuse parts of the previous attempt. But it's not a guarantee. That means using local component state to cache data won't work, because that state will most likely not be persisted. The underlying priciple here is that you can't cache the intermediate results of a computation (server data) on the output of the computation itself (the React tree).
+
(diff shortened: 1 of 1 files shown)


theKashey:
> For each context, React would have to maintain a separate version per root, as well as a separate queue of updates.

I was just thinking about it - per-root changeable defaultValue would allow "automatic Provider injection" effect, making some stuff(like React-Helmet, side-effect, etc) easier.

jfo84:
Great job @acdlite :)

text/0000-context-write.md
```diff
@@ -0,0 +1,242 @@
...
+
+An alternative model is to treat all the roots as siblings and commit them at the same time. This could be viable for single page React apps (where there aren't that many roots, anyway). It doesn't work so well in cases where React is embedded inside another framework (e.g. progressive enhancement of a server rendered app), where temporary inconsistencies may be desirable. For example, the opt-in API for concurrent mode relies on roots committing separately so that you can upgrade some roots to concurrent mode without upgrading the entire app. In an app with mixed synchronous and concurrent roots, a unified commit would mean that every call to `Context.write` has to be synchronous, which probably makes this option a non-starter.
+
+In either of these models, React would need to track a global list of all roots in order to schedule updates on them. This isn't something we do currently, and it means roots would no longer be automatically garbage collected; discarded roots would need to be explicitly unmounted to remove them from the global list.
```
hikire: If by "discarded roots" you mean those which are just removed from DOM without calling ```unmountComponentAtNode()```, why not use ```MutationObserver```?

text/0000-context-write.md
```diff
@@ -0,0 +1,242 @@
...
+
+An external store's state is also not managed by React, so it can't be fully compatible with concurrent rendering. Short of full compatibility, even limited compatibility without tearing is difficult to implement correctly without deep knowledge of React's rendering model. An idiomatic API designed for this use case should at the very least make it possible to lift state out of React without causing tearing and without always falling back to synchronous mode.
+
+# Detailed design
+
+## Motivating examples
+
+### Immutable store (like React Redux)
+
+Implementing immutable stores is straightfoward. Actions are applied in the render phase using the functional form of `Context.write`.
```
mfrachet: There's a point I'm misunderstanding. What does it mean concretely that 

Actions **are applied in the render phase** using the functional form of `Context.write`?

text/0000-context-write.md
```diff
@@ -0,0 +1,242 @@
...
+
+We've come this far without the need for an API like this. But the main reason we need to address this now is because of Suspense and concurrent rendering. Given how important this use case is for Suspense, doing nothing seems unlikely.
+
+## Mount a "shell" component and cache using local state
+
+Another way to move this to user space would be to have a shell component at the root of the app that initially renders with no children. On mount, it schedules a re-render to mount the children. Because the shell is already mounted, the children can cache values in the shell's local state.
+
+This doesn't address the multiple root problem, nor does it work with React's server renderer, which does not have updates.
+
+## Use Hooks
```
timoxley: +1 for more symmetry between component state and context. Ideally, lifting component state & logic up to into a shared context wouldn't require rewriting the whole thing.

diegohaz: Agreed. I've been doing stuff on this direction on https://github.com/diegohaz/constate

text/0000-context-write.md
acdlite: How would that work? AFAIK MutationObserver only works for immediate children. It won't tell you if a tree is disconnected by an ancestor node.

text/0000-context-write.md
acdlite: https://twitter.com/acdlite/status/977291318324948992

rybon:
@acdlite fantastic proposal!

What's the plan? Where do we go from here? Is there a concrete path forward like there is for the Hooks proposal?

zombieJ:
Is that possible to add `freeze` like:
```jsx
const Context = React.createContext(initialValue, contextDidUpdate);
Context.freeze();

Context.write(newValue); // Error. This context has been freeze and can't change defaultValue.
```

dancerphil:
implement partly in https://codesandbox.io/s/react-context-next-xk1rk

mucsi96:
I have submitted an RFC with an alternative solution. I believe more inline with hooks direction.
https://github.com/reactjs/rfcs/pull/130 
Would like to ask for feedback from all of you :)

gaearon:
Looks like this either needs more work and/or won't work at all.

PR #96, opened 2018-12-10
Custom Host Node RFC

sebmarkbage:
***This is just an idea and not likely to land any time soon.***

## Summary

This is an alternative idea to useMutationEffect that was originally part of the hooks RFC but since removed.

## Basic example

```js
import {createManualDOMNode} from "react-dom";
let CustomDOMNode = createManualDOMNode(props => {
  // create
  let node = document.createElement('div');
  node.style.color = props.color;
  return node;
}, (props, node) => {
  // update
  node.style.color = props.color;
}, (props, node) => {
  // clean up
});

function Component(props) {
  return <div><CustomDOMNode color={props.color} /></div>;
}
```

### [Rendered text](https://github.com/reactjs/rfcs/blob/customhostnode-rfc/text/0000-custom-host-node.md)

### DIFF
--- a/text/0000-custom-host-node.md
+++ b/text/0000-custom-host-node.md
@@ -1,0 +1,62 @@
+- Start Date: 2018-12-10
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Custom Host Node
+
+I was thinking about the use case for useMutationEffect. The use case is to implement your own manual updates to a DOM node, or to a DOM node via a legacy imperative API like jQuery.
+
+The idea is that it is better to do before useLayoutEffect and componentDidMount because they might read layout + followed by setState. So if the first sibling reads layout, the second mutates the DOM, the third reads layout, you get two layout passes instead of one.
+
+However, the problem with useMutationEffect is that refs haven't been swapped yet. During initial render, you don't have access to the new element that you want to mutate. During updates, you have access to the old element but not the new one if a ref has swapped instance.
+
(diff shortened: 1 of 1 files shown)


oriSomething:
I like the API. But excluding the advantage of cleaner API, Why not using component with `shouldComponentUpdate: () => false` + ref for creating something similar? The "updates" callback seems cleaner, but I'm not sure if this API, is really must

Andarist:
What about memoization? Or rather firing update callbacks only when inputs has changed? To some extent it can be done inside the update callback itself, but it's hard (impossible?) to store some previous state in order to do the comparison.

jameswomack:
> What about memoization? Or rather firing update callbacks only when inputs has changed? To some extent it can be done inside the update callback itself, but it's hard (impossible?) to store some previous state in order to do the comparison.

@Andarist do you mean "impossible" to do **self-contained**—or *strictly* impossible? Is it not possible to use a collection-type instance to memoize in a closure around the result of `createManualDOMNode` due to some arcane detail of the Hooks API?

Andarist:
With HOC approach we lose access to the component's closure (unless we do some hoops to create that HOC with `useMemo` or smth), with hook approach we'd have to store previous values on the ref or node and do custom inputs check in the updater (instead of relaying on the input array and react doing that work for us like when dealing with all the other hooks)

msand:
You could probably use a weakmap to store state related to the nodes as well, without storing anything on the node/ref

PR #112, opened 2019-04-12
RFC: React.isEvent

jamiebuilds:
### [[Rendered](https://github.com/jamiebuilds/rfcs-2/blob/patch-2/text/000-isEvent.md)]

### DIFF
--- a/text/000-isEvent.md
+++ b/text/000-isEvent.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-A new API `React.isSyntheticEvent(value)` which returns `true` if the value is a `SyntheticEvent` or `false` if it is anything else. 
+A new API `React.isEvent(value)` which returns `true` if the value is a `SyntheticEvent` or an `Event` and returns `false` if it is anything else. 
 
 # Basic example
 
@@ -15,7 +15,7 @@ export function useFormField<T>(defaultValue: T) {
   let [value, setValue] = React.useState(defaultValue)
 
   function onChange(val: T) {
-    if (React.isSyntheticEvent(val)) {
+    if (React.isEvent(val)) {
       throw new Error("Expected `formField.onChange` to be called with a value for the form field, not an event")
     }
     setValue(val)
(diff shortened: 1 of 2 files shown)


jamiekyle-eb:
Also see: https://github.com/taehwanno/is-react-synthetic-event

cc @taehwanno

sebmarkbage:
What’s the actual motivation for using this? It’s not obvious to me. Conversely, when is it a bad idea to use this?

If this is because the event object is reused, that’s something we could just get rid of. It’s probably a premature optimization.

One idea is also to get rid of the synthetic objects and instead use real ones for the standard events. As long as most code just uses the standard properties this gets easier but if code starts taking different paths it gets harder to upgrade.

Liskov’s substitution principle and all that.

jamiebuilds:
The motivation for me is to be able to write an overloaded method which accepts a synthetic event from React as an input.

```ts
function onChange<T>(val: T)
function onChange(val: SyntheticEvent)
```

Where inside I want to write:

```ts
let value

if (val instanceof Event || isSyntheticEvent(val)) {
  value = val.currentTarget.value
} else {
  value = val
}
```

Because `val` can be an arbitrary type `T` I want to filter it down to exactly `T - SyntheticEvent`

Right now the only ways to check if an arbitrary value is a SyntheticEvent are:

- Assert some portion of the shape of SyntheticEvent which another unrelated object is unlikely to match `typeof val === "object" && "currentTarget" in val`
- Create a known SyntheticEvent and use its prototypal object to check with `instanceof`

sebmarkbage:
A more palatable one would be isEvent that returns true for either Event or SyntheticEvent since that allows for us to start sending regular Events.

Technically you could then do `isSyntheticEvent = e => !(e instanceof Event) && isEvent(e);` but at least it encourages thinking of them as at the same.

jamiebuilds:
Sure, I'd be happy with that API. I can update this RFC

trueadm:
There's also the case of the new experimental event API that we're testing internally: https://github.com/facebook/react/issues/15257

Specifically, with the new event API, the concept of "synthetic events" is less defined. In fact, event responder modules (for example [Press](https://github.com/trueadm/react/blob/master/packages/react-events/src/Press.js#L113-L123)) define their own interface of event that differs from how native and synthetic events work today (for example, there is no `preventDefault` or `stopPropagation` on these events).

I'm not sure how this proposal works with this experimental event API. It might be the case in the future that we only have two forms of event objects: native event objects (as @sebmarkbage pointed out) for inline event handlers and event objects defined in event responder modules from the new event API.

gaearon:
I agree that we don't want to encourage differentiating between Events and Synthetic Events, especially as we're moving towards unifying behavior (17 doesn't do pooling, for example). So that seems like a non-goal to me.

As for detecting Event-like things, that seems a bit out of scope to me. Similar to a `.then` check for thenables. 

The canonical solution is probably this:

```js
let value
if (val?.currentTarget !== undefined) {
  value = val.currentTarget.value
} else {
  value = val
}
```

So I will close this.

PR #115, opened 2019-05-05
RFC: context.createProvider (combine hooks and Provider)

Volune:
## Summary

> The purpose of this proposal is to allow developers to create custom context Provider using hooks for state management. It's sort of a merge of *FunctionalComponent* and *ContextProvider*.

### [basic example](https://github.com/Volune/rfcs/blob/proposal-create-provider/text/0000-context-createprovider.md#basic-example)

```
// See *rendered* for full basic example
context.createProvider(function MyProvider() {
  // compute new context value, can use hooks
  return newContextValue;
})
```

### [motivation](https://github.com/Volune/rfcs/blob/proposal-create-provider/text/0000-context-createprovider.md#motivation)

- > reduce the size of the virtual tree
- > simplify how the "changed bits" of a context value are computed

## Prototype

The proposal is simple enough, I was able to prototype it
- [Codesandbox example](https://codesandbox.io/s/n95nk3vpyp)
- [devtool integration example screenshot](https://gist.githubusercontent.com/Volune/c30462df9292864a3df80c8076410964/raw/6cf24465061962482e95b5c9047b63e44416117b/react-devtool-createprovider.jpg)
- [react fork for prototype](https://github.com/Volune/react/tree/proposal-create-provider)

### DIFF
--- a/text/0000-context-createprovider.md
+++ b/text/0000-context-createprovider.md
@@ -1,0 +1,155 @@
+- Start Date: 2019-04-30
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+
+# Summary
+
+The purpose of this proposal is to allow developers to create custom context 
+ Provider using hooks for state management. It's sort of a merge of 
+ *FunctionalComponent* and *ContextProvider*.
+
+
+# Basic example
+
+This example shows how `context.createProvider(...)` could be used to create
+a library that gives the screen height and width.
+
+```js
+import {
+  createContext,
+  useContext,
+  useEffect,
+  useMarkChangedBits,
+  useState
+} from "react";
+
+const WIDTH_BITS = 0b0100;
+const HEIGHT_BITS = 0b1000;
+
+const getScreenSize = () => ({
+  width: window.innerWidth,
+  height: window.innerHeight
+});
+
+const INITIAL_CONTEXT_VALUE = getScreenSize();
+
+const context = createContext();
+
+export const ScreenProvider = context.createProvider(function ScreenProvider() {
+  const [screenSize, setScreenSize] = useState(INITIAL_CONTEXT_VALUE);
+
+  useEffect(() => {
+    const handler = () => {
+      setScreenSize(getScreenSize());
+    };
+    window.addEventListener("resize", handler);
+    return () => {
+      window.removeEventListener("resize", handler);
+    };
+  }, [setScreenSize]);
+
+  useMarkChangedBits(WIDTH_BITS, [screenSize.width]);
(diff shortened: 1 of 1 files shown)


jacobp100:
This proposal would also make it way easier to manage providers that depend on other providers, as the dependencies can be automatically managed. It’s too easy to get wrong when you ha e a pyramid of `<Provider>`s!

gaearon:
Thanks for the RFC, but I don't think we'll go forward with this.

We've already removed "changed bits" from context so that part of the proposal is now not relevant. As for reducing the depth of the tree, I disagree that this proposal helps a lot with that. It doesn't seem far from what you do by extracting a component. Here's the motivating example, rewritten to idiomatic React:

```js
const getScreenSize = () => ({
  width: window.innerWidth,
  height: window.innerHeight
});

const INITIAL_CONTEXT_VALUE = getScreenSize();

const ScreenContext = createContext();

export function ScreenProvider({ children }) {
  const [screenSize, setScreenSize] = useState(INITIAL_CONTEXT_VALUE);

  useEffect(() => {
    const handler = () => {
      setScreenSize(getScreenSize());
    };
    window.addEventListener("resize", handler);
    return () => {
      window.removeEventListener("resize", handler);
    };
  }, [setScreenSize]);

  return (
    <ScreenContext.Provider value={screenSize}>
      {children}
    </ScreenContext.Provider>
  );
}

export const useWidth = () => {
  const { width } = useContext(context);
  return width;
};

export const useHeight = () => {
  const { height } = useContext(context);
  return height;
};
```

The result looks very similar, but more explicit about the tree structure.

In the future, we've been considering changing the context API to be destructutable:

```js
const [ScreenContext, useScreenContext] = createContext()
```

Here, `ScreenContext` itself would be a Provider. 

This would give us:

```js
const getScreenSize = () => ({
  width: window.innerWidth,
  height: window.innerHeight
});

const INITIAL_CONTEXT_VALUE = getScreenSize();

const [ScreenContext, useScreenContext] = createContext();

export function ScreenProvider({ children }) {
  const [screenSize, setScreenSize] = useState(INITIAL_CONTEXT_VALUE);

  useEffect(() => {
    const handler = () => {
      setScreenSize(getScreenSize());
    };
    window.addEventListener("resize", handler);
    return () => {
      window.removeEventListener("resize", handler);
    };
  }, [setScreenSize]);

  return (
    <ScreenContext value={screenSize}>
      {children}
    </ScreenContext>
  );
}

export const useWidth = () => {
  const { width } = useScreenContext();
  return width;
};

export const useHeight = () => {
  const { height } = useScreenContext();
  return height;
};
```

PR #118, opened 2019-06-21
RFC: Lazy Context Propagation

gnoff:
## [View Rendered Text](https://github.com/gnoff/rfcs/blob/lazy-context-propagation/text/0000-lazy-context-propagation.md)

This RFC describes a new Context Propagation implementation which has improved performance characteristics in certain cases by using the work React is already going to do before resorting to explicit context dependencies searches. This lazy implementation also has qualitative differences from the current eager approach which may allow new kinds of context update bailouts to be implemented

 It does not modify the public API of React in any way.

>...unmounts will now happen before context propagation begins.

>...will allow for some interesting new APIs.

## Addendum
Example: https://codesandbox.io/s/react-lazy-context-propagation-67e5j
Implementation: https://github.com/gnoff/react/pull/4/files

### DIFF
--- a/text/0000-lazy-context-propagation.md
+++ b/text/0000-lazy-context-propagation.md
@@ -1,0 +1,104 @@
+- Start Date: 2019-06-19
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This RFC describes a new lazy context propagation implementation which uses the work React is already going to do when possible, before resorting to explicit context dependency searches. This results in improved performance characteristics for some use cases. Additionally it may enable new kinds of context update bailouts to be implemented.
+
+It does not modify the public API of React in any way.
+
+# Motivation
+
+The eager context propagation in place today is executed per ContextProvider. When a ContextProvider is updating to a new value every constituent fiber has its context dependencies checked against the updating Context and work is scheduled if necessary.
+
+This approach has two direct behaviors that can be optimized. First, for any given render, each ContextProvider with a new value will result in the constituent sub-tree for that ContextProvider to be visited. In a simplified scenario, with ContextProviders at the top of the tree, if `N` ContextProviders are updating then the fiber tree will be traversed during propagation exactly `N` times.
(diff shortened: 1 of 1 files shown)


sebmarkbage:
Since you have an implementation now, I wonder if you have some data on how much this helps in real apps?

One thing to consider is that the reason context propagation is acceptable fast when optimized (it's actually unnecessary slow now, thanks to code rot, and we need to optimize it again), is because both code an a lot of the data is very local and benefit a lot from caches. Locality can make brute force surprisingly fast.

By doing this propagation over and over, we might negate some of those benefits. Especially when only a single context updates.

gnoff:
I think you're right that we lose some caching benefits so the imagined speedup isn't very prevalent. however I think the qualitative differences in lazy propagation unlock much more powerful bailouts such as with #119

This RFC hasn't had a lot of engagement and I haven't spent a lot of time using it in real world apps.

As for whether an optimized context propagation is fast enough, i agree. the salient point of this change isn't that it is intrinsically faster but that if work is scheduled to happen anyway, it has a chance to finish. the focus on perf was it should not be slower, it sometimes is faster, and it allows things like context selectors or whatnot that would be impossible (wasteful) with eager propagation

zzz6519003:
like Lazy Eval?

gaearon:
Small update. We've implemented a version of this, and it seems to have worked well. It's currently behind a flag but we'll need to do more testing and rollout before we can switch it on by default. I'm not sure if we have any noticeable perf data since the initial experiment was small.

gaearon:
We’ve essentially implemented a version of this so it’s fair to say that this has been approved. Thank you!

horaciosystem:
hey @gaearon will this be available in React 18? Could we use with the flag?

gaearon:
We'll probably enable it in some 18.x minor release. Until then you can build from main if you'd like to try it.

frontend-winter:
Hopefully this will be available in React version 18

PR #124, opened 2019-10-03
Add React Hooks in Classes RFC

GasimGasimzada:
In this RFC, I am talking about adding Hooks for class based components. Usage will look something like this:

```jsx
class SomeComponent extends React.NewComponent {
    someFn = () => {
        // use this.props here
    }

    state = new State('test');
    effect1 = new Effect(this.effectFn, [this.state]);

    effectFn = () => {
        const val = this.state.getValue();

        someFn();
        // or state.setValue('some other value')
    }

    render() {
        // you can use this.state.getValue() here
        return (

        );
    }
}
```

I believe that a lot of confusing parts of hooks can be alleviates when using OOP due to behaviors that are engrained in classes. This feature is not here to replace functional components but to provide a way to compose components using classes.

EDIT: Here is a demo for the proof-of-concept: https://github.com/GasimGasimzada/class-based-hooks. I am wrapping the class component in `createComponent` function to create a function component but if this is implemented in React space, we will be able to natively create class instances by extending them from something like `React.NewComponent`.

@bvaughn I also forgot to mention you in this RFC after our discussion in https://github.com/facebook/react/issues/16995. I was even able to add a working demo to show the usage.

### DIFF
--- a/text/0000-react-hooks-in-classes.md
+++ b/text/0000-react-hooks-in-classes.md
@@ -1,0 +1,228 @@
+- Start Date: 2019-10-03
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Provide a new Class Based Component API that utilizes React hooks.
+
+# Basic example
+
+```jsx
+class SomeComponent extends React.NewComponent {
+    someFn = () => {
+        // use this.props here
+    }
+
+    state = new State('test');
+    effect1 = new Effect(this.effectFn, [this.state]);
+
+    effectFn = () => {
+        const val = this.state.getValue();
+
+        someFn();
+        // or state.setValue('some other value')
+    }
+
+    render() {
+        // you can use this.state.getValue() here
+        return (
+
+        );
+    }
+}
+```
+
+# Motivation
+
+In my opinion, combined with Hooks, React's composability is an implementation of [Entity-Component-System](https://en.wikipedia.org/wiki/Entity_component_system). Game Developers use this technique to set up entities (e.g a Tree or a Vehicle) in their game world. We are essentially using the same technique to set up entities (React Components) for UI. If we actually map ECS to React, this is how I see it: Entity = React Component, Hooks = Components, System = React Reconciler + Renderer.
+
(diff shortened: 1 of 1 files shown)


GasimGasimzada:
@stephenh I have implemented an approach that works in user land. For the sake of simplicity, I have added support for useState, useEffect, useContext, and custom hooks. The way it works is that, I am using the class instance to create a function component. This way, I can work with all the hooks inside classes. Usage is very simple: just wrap the class component in `createComponent` function. This function creates a function component, then uses function component props to create an instance of the class. Then, all the hooks in the instance are ran. Finally, `render` function of the instance is ran. This way, the scope of the class instance is preserved, meaning `this` from within class hooks and functions. Here is the demo (its a CRA app; so, just install deps and do `yarn start`):

https://github.com/GasimGasimzada/class-based-hooks

In order to use the same logic by ONLY deriving from `NewComponent`, it needs to be handled in the reconciler. The demo that I added is just a proof-of-concept. There might be edge cases that I did not check. My only concern with this approach is Effect hook's dependencies array. Some thought needs to be put into this part to provide a consistent experience for the user. Otherwise, this fixes a lot of issues: functions are created only once, member variables are initialized only once.

mAAdhaTTah:
The core issue with this proposal is that hooks designed for Class components can't be used by Functional components and vice versa. I'm not sure this is all that beneficial.

stephenh:
@mAAdhaTTah well, yes, I agree that option (dual FP/class compatibility) has passed.

It's unfortunate because IMO a few compromises on the hooks API would have made them useable from something very similar to @GasimGasimzada's "use class instance fields to store hooks" approach (basically returning handles/getters instead of straight values, which admittedly is a compromise on the FP UX side), and then we wouldn't have this dichotomy in community of "well, all the new APIs use hooks, so now your components have to be FP to use them".

But, just because we don't/can't do that now, I don't think means this proposal is merit-less.

I.e. I assert there is a large enough subset of React users out there who would like to still use class-based components, that, if given the option to write hook-ish style abstractions (primarily things like Apollo's `useQuery` but from a class component), would choose to do so, but right now they can't.

GasimGasimzada:
@mAAdhaTTah That's understandable. I would like to close this issue if that's okay.

gaearon:
Hi, thank you for the proposal.

Aside from the confusion of having two different parallel "Hooks" systems, this also doesn't help concurrent rendering because it re-introduces reliance on a mutable instance as the state container. This is the problem Hooks solve by relying on closures and capturing. So this is not something we would consider.

PR #125, opened 2019-10-10
Add Context.currentValue

stephenh:
[View formatted RFC here](https://github.com/stephenh/rfcs/blob/add-context-current-value/text/0000-add-context-current-value.markdown)

### DIFF
--- a/text/0000-add-context-current-value.markdown
+++ b/text/0000-add-context-current-value.markdown
@@ -1,0 +1,187 @@
+- Start Date: 2019-10-07
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Add a `Context.currentValue` method as part of Context's public API.
+
+# Basic example
+
+```javascript
+const MyContext = React.createContext({ name: "foo" });
+
+// inside of a class component constructor or lifecycle method
+class FooComponent {
+  constructor(props) {
+    super(props);
+    this.name = MyContext.currentValue.name;
+  }
+}
+```
+
+Or a `useQuery`-ish API that is available to class components:
+
+```
+class FooComponent {
+  constructor(props) {
+    super(props);
+    // newQuery is able to access the ApolloProvider.currentValue 
+    this.query = ReactApollo.newQuery({ ... });
+  }
+
+  render() { 
+    const { ... } = this.query;
+  }
+}
+```
+
+Or as field initializers:
+
+```
+class FooComponent {
+  // newQuery is able to access the ApolloProvider.currentValue 
+  query = ReactApollo.newQuery({ ... });
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() { 
+    const { ... } = this.query;
+  }
+}
+```
+
+# Motivation
+
+The motivation is to provide class components (_and_ code they might invoke) with a non-HOC,
+non-render props, non-`contextType` way to access `Context` values, similar to what hooks
+are able to do with `useContext`.
+
+Currently `contextType` allows a very limited form of this, but couples the component to
+a single context, and does not allow the component to invoke reusable code that, on behalf
+of the class component, can easily access whatever contexts it wants (i.e. the GraphQL
+context or the Redux context).
+
+The intent is basically to "level the playing field" between FP components and class components,
+by providing class components with what (IMO) is the biggest win of hooks: easy, intuitive,
+decoupled access to contexts from utility/infrastructure libraries like react-apollo/etc.
+
+(Well, that and a way to access the component's lifecycle methods in a decoupled manner, but that
+is a separate topic/RFC.)
+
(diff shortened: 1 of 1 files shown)


text/0000-add-context-current-value.markdown
```diff
@@ -0,0 +1,187 @@
...
+
+- It prolongs the "class vs. FP" divide by giving class components a lifeline to
+  hook-level ease-of-use APIs.
+
+  Given the popularity of hooks-style APIs, the "class vs. FP" dictomy will likely
+  very naturally play itself out in favor of FP, which is perhaps the intent of
+  hooks / direction of the React core team in the first place.
+
+- Callers would be responsible for "refreshing" the context value, i.e. if they
+  capture it in the component constructor, and then the context value changes,
```
phryneas: If you want an API like that, why do you want to use context at all? Context ist about re-rendering subscribing components when it changed.
If you want a way of holding a value that can be accessed from anywhere without triggering a re-render, that's a global JavaScript variable (or a singleton, or anything like that). That behaviour you are describing isn't coupled to React at all - so why do you want to use React for it?

stephenh: A global variable is generally brittle as it can only have a single value for the entire process; dependency injection, of passing non-global values, is generally best practice these days in most/all languages, and React's context/provider is the React way of doing dependency injection.

I.e. the context can have different values at different spots in the DOM, or different values for different tests running in parallel, or what not.

> that can be accessed from anywhere without triggering a re-render

Technically I do want a re-render to be triggered, I'm just acknowledging that clients would have to know to not keep the original context value across re-renders, in case the re-render was specifically triggered by a provider value being changed. Apologies if that wasn't clear.

stephenh:
@bvaughn hey Brian; not getting a ton of input on this :-); should I be pinging somewhere? Or just take the lack of input as indicator that no one on the React core team/community is interested in helping flush out/propose alternative approaches to the same goal/etc? Thanks!

bvaughn:
The lack of input just signals that our team is small and trying to do many things at once. It doesn't always make sense to dive into an RFC that's not the core set of things we're working on at the moment.

stephenh:
@bvaughn cool, no worries; didn't mean to come across as antagonistic. I guess just let me know if I should be doing more, like attempting to proof-of-concept things in the React codebase or what not.

(Fwiw I'm currently hesitant to do any prototyping until getting at least some signal of "yeah this doesn't seem dumb", or "here's an obviously easier/better way of doing this", as I'd have some non-trivial learning cost on the React internals that I'm currently procrastinating on paying if I shouldn't bother.)

But I'm not in a rush either, so perfectly fine letting this RFC hang out for a bit, if/until someone on the core/community team has time to focus on it. Thanks!

gaearon:
I don't think this is something we would like to do.

It's not that we want to make working with classes artificially harder by creating barriers. If anything, this is not a new barrier — classes always worked this way. Even legacy context couldn't be read like this implicitly.

Yes, Hooks do make it easier. But they pay for it with an explicit call convention. i.e. if you `useContext`, you can only do this inside a Hook, which forces that thing to also be `useSomething`. Which means that arbitrary helpers can't officially "just read context". This is good because it leaves behind a trail. You can't later get surprised that some random helper inside your render happens to work only during rendering. If something only works during rendering, it has `use` in its name, and you know to not move it out.

If the classes specifically aren't sufficiently ergonomic, one can always have a thin outer layer that reads multiple contexts and passes it down:

```js
function Outer() {
  const a = useContext(A)
  const b = useContext(B)
  const c = useContext(C)
  return <MyClass a={a} b={b} c={c} />
}
```

That wouldn't really require you to "buy into" Hooks 100% but still lets you read context ergonomically.

Sometimes, as you say, you might still think a library "knows better". There exists a hidden secret field you can use to do that. Relay [does it](https://github.com/facebook/relay/blob/597d2a17aa29d401830407b6814a5f8d148f632d/packages/react-relay/readContext.js). Note that if you use it, all bets are off. We don't promise backwards compat there but you can do it at your own risk.

In the longer term, it's clear the ecosystem is moving towards Hooks. So you're right we're mostly leaving classes in maintenance mode. That's another argument to not add an implicit API to them. Especially given that people still holding onto classes tend to favor explicitness.

So I think we can close this. I hope my response makes sense and gives you some pointers.

stephenh:
Cool, thanks for the pointer to relay's approach. I'll see about using it / bookmark it, but right, at this point it seems inadvisable to not just fall in line with hooks. 

> Especially given that people still holding onto classes tend to favor explicitness.

Well, true, although react's context is essentially a thread local variable, which is very common in OO/class/etc. environments and, while "not something you use on a whim", is commonly used for the same places that react's context is. :man_shrugging: 

Nonetheless/also just thanks in general for the mass gardening of RFCs. I really appreciate the time you took to give useful thoughts on each one.

> I hope my response makes sense and gives you some pointers.

+1

PR #139, opened 2020-01-11
Profiler should measure commit phase durations

bvaughn:
### [View formatted RFC](https://github.com/bvaughn/rfcs/blob/profiler-commit-durations/text/0000-profiler-measure-commit-durations.md)

Implementation: https://github.com/facebook/react/pull/17910

### DIFF
--- a/text/0000-profiler-measure-commit-durations.md
+++ b/text/0000-profiler-measure-commit-durations.md
@@ -70,3 +70,8 @@ Both new `Profiler` callback props will be optional. Existing callsites will con
 # How we teach this
 
 I will write a React block post (or similar) with examples of how to use the new APIs and what they mean.
+
+# Unresolved questions
+
+* Will the proposed implementation be able to handle cascading updates with renders and commits?
+* Is it okay to not report time spent in clenaup functions for certain unmount scenarios?
--- a/text/0000-profiler-measure-commit-durations.md
+++ b/text/0000-profiler-measure-commit-durations.md
@@ -116,5 +116,5 @@ I will write a React block post (or similar) with examples of how to use the new
 
 # Unresolved questions
 
-* Will the proposed implementation be able to handle cascading updates with renders and commits?
-* Is it okay to not report time spent in clenaup functions for certain unmount scenarios?
+* It would be very difficult to report time spent in cleanup functions for passive effects in certain umount cases (with nested Profilers) given the current unmount implementation. We plan to refactor this code as part of the concurrent mode release. The question here is: is it okay to land an initial version of this API that does not handle this case (and fix it as part of the refactor)?
+* Should we add interaction tracing support to these new methods? Tracing already adds a lot of complexity and (arguably) little value.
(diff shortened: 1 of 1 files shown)


bvaughn:
### Should this RFC drop passive effect timing?

As I dig into actually implementing this, I've realized there are a few challenges presented by passive effects. The biggest is that we can't call `onPostCommit` during the commit phase tree traversal because:
1. Passive effects won't have been processed by this time (at least not for the render work we're currently committing). Duration values will represent the previous round of passive effects.
1. We *could* call the previous callback at this time, although that presents additional problems:
   1. The `memoizedInteractions` set corresponds to the current batch of render work, so it would be incorrect.
   1. The `memoizedProps.id` of the Profiler also corresponds to the current batch of render work. (It may have changed since the render our passive effects are part of.)

I think this means that we would need to do the following:
1. Add a traversal pass to `flushPassiveEffectsImpl` that:
   1. Calls `onPostCommit` for any Profilers with this prop registered
   1. Bubbles durations to the next nearest parent Profiler
1. Add a new DevTools hook (so the Profiler actually has a way to read these transient values if it wants to report them in the Profiler somehow). Normally DevTools reads values in its commit hook, but this won't work for passive effect durations.

Adding a new DevTools hook is probably fine, but I think an added traversal pass is probably not a great idea. Do passive effect durations actually provide enough value to offset this?

bvaughn:
### Should we measure cleanup functions?

This is tricky in a similar way as passive effects, since cleanup functions don't get run until an update (or unmount)- after we've already reported durations.

We have a couple of options here:
1. Don't measure these durations. Typically cleanup functions are small (e.g. removing an event listener). This would be a continued blindspot in the Profiler API though.
1. Delay reporting commit durations until the *next* commit (after cleanup functions have been run) so the durations are inclusive of both. This has similar complexity downsides as mentioned above for passive effects though. It would also the last profiled commit in a session would always be missing that info.
1. Measure cleanup functions, but include their durations as part of the current commit. An argument could be made for this being the *best* way to think of them, since their durations impact the *current* commit (not the previous one), although it might be confusing since e.g. interactions are for the current commit.
1. Add new callbacks for cleanup durations (e.g. `onCommitCleanup`? `onCommitDestroy`?) and report them separately. This would provide the most insight into where slowness comes from, but I also think this might be overkill.

My current vote is for option 3 above (include their durations as part of the current commit).

### Should we also measure "before mutation" code?

Currently this only includes the class component `getSnapshotBeforeUpdate` lifecycle, but we will probably add a before-mutation hook at some point too.

Given the currently proposed callbacks (`onRender`, `onCommit`, `onPostCommit`) it might make sense to add a new Profiler hook (`onPreCommit`?) for "before mutation" lifecycles. Could be done as part of this RFC, or as part of a follow up.

I have not decided at this point which option seems best.

bvaughn:
I've updated the RFC to incorporate the above comments.

bvaughn:
Implementation PR has been accepted and merged, so I'm going to merge the RFC.

PR #145, opened 2020-02-12
New RFC: Mechanism to substitute the react build.

Gaelan:
This is fairly borderline in terms of whether it needs to be an RFC—if this is better suited for the primary react repo, I'm happy to move it over there.

### DIFF
--- a/text/0000-development-injection.md
+++ b/text/0000-development-injection.md
@@ -1,0 +1,96 @@
+- Start Date: 2020-02-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Provide a mechanism for DevTools (or another browser extension) to switch a
+page to a development build of React.
+
+# Motivation
+
+Every once in a while, all of us end up debugging a production build of our
+apps. Maybe there's a bug that only shows up in a production app, or maybe
+you're troubleshooting an issue a user's experiencing that you can't
+reproduce locally.
+
+React recommends developing against a development build for a reason—a lot
+of weird React bugs are detected and warned about, but only in development
+builds. If there was a quick way to run "the prod build of my site, but
+with a development build of React," debugging some of these issues could
+be substanially easier.
+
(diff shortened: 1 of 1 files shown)


text/0000-development-injection.md
```diff
@@ -0,0 +1,96 @@
...
+you're troubleshooting an issue a user's experiencing that you can't
+reproduce locally.
+
+React recommends developing against a development build for a reason—a lot
+of weird React bugs are detected and warned about, but only in development
+builds. If there was a quick way to run "the prod build of my site, but
+with a development build of React," debugging some of these issues could
+be substanially easier.
+
+The specific issue that inspired my to file this RFC was an issue with
```
gaearon: >The specific issue that inspired my to file this RFC was an issue with
Gatsby SSR: Gatsby only uses SSR in production, and we had a bug that
caused SSR to generate a slightly different component tree than the
client, causing a hard-to-track-down issue with hydration.

That sounds like an important issue for Gatsby to solve. We assume that people using SSR develop with it enabled too. Is there an issue I can read about this?

gaearon:
I think maybe @sophiebits suggested something similar at some point.

sophiebits-humu:
Yes. Still think this would be useful.

sophiebits:
Yes. Still think it’s a good idea.

text/0000-development-injection.md
Gaelan: I think this is the relevant Gatsby issue: https://github.com/gatsbyjs/gatsby/issues/17914

gaearon:
Tbh this seems more oriented towards GH repo with a proof of concept.

gaearon:
I think the really annoying part here is that the versions need to match, and that's a bit tricky. I guess DevTools could try to download a matching version from npm. Then there's a question of: do we also replace `react` (or `react-dom` only), what about `scheduler`, what to do when they get optimized/inlined by the bundler, and so on. I think an RFC for this would need to go into a lot more detail about how the hard parts are solved.

PR #148, opened 2020-02-15
ES Module as allowed component type

penx:
If `MyComponent` is an ES Module, then `<MyComponent>` should behave the same as `<MyComponent.default>`

(e.g. transpile to `React.createElement(MyComponent.default)`).

This is primarily to support syntax such as the following, while ensuring that tree shaking still will still function (see diff for details):

```jsx
 import ControlPanel from './control-panel';

 export default = () => <ControlPanel>
     <ControlPanel.Button />
     <ControlPanel.Button />
 </ControlPanel>;
 ```

### DIFF
--- a/text/0000-module-component-type.md
+++ b/text/0000-module-component-type.md
@@ -1,0 +1,133 @@
+- Start Date: 2020-02-15
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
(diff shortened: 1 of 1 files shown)


penx:
closing this as it will likely require a runtime overhead to solve a semantic and fairly uncommon case

PR #150, opened 2020-02-28
RFC: Speculative Mode

gnoff:
## Summary

This RFC describes an approach to making all work done in React lazier. This was inspired by @sebmarkbage 's comments in response to #118 and #119 where the lazy approach seems good but the whole React work process could and should benefit from it

In a few sentences Speculative Mode is a feature where if React bails out of updates it will no longer clone child fibers and but will actually start work on the current tree. It defers the creation of workInProgress until it has actually confirmed an update is required. Only then does the current tree get copied into the workInProgress tree

## [See Rendered Text](https://github.com/gnoff/rfcs/blob/js-speculative-work-rfc/text/0000-speculative-work.md)

Implementation: https://github.com/gnoff/react/pull/7
Example: https://codesandbox.io/s/speculative-mode-dr89t

Find me on twitter ([@joshcstory](https://twitter.com/joshcstory)) or tag me (@gnoff) in a comment if you have any questions

### DIFF
--- a/text/0000-speculative-work.md
+++ b/text/0000-speculative-work.md
@@ -4,23 +4,11 @@
 
 # Summary
 
-This RFC proposes a change to the fiber work implementation to defer creating
-work-in-progress fibers until and unless it is certain we need to. It builds on
-ideas initially outlined in #118 and #119 where context propagation was done
-lazily. This general idea is extended to all kinds of work and is summarized as
-follows.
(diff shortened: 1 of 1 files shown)


gaearon:
I'll keep this open but wanted to note for readers that parts of the ideas from here have been upstreamed into experiments, while the rest is too complicated to fit into the current architecture — but we'll keep it in mind for the next time we revamp it.

PR #158, opened 2020-04-19
useKeyedGroups; bending the Rules of Hooks with "keys for hooks"

Nathan-Fenner:
`useKeyedGroups` makes it possible for libraries and application writers to composably and declaratively consume custom hooks that depend on dynamic props in cases that are currently forbidden by the Rules of Hooks.

# [View formatted RFC](https://github.com/Nathan-Fenner/rfcs/blob/master/text/0000-keyed-hooks.md)

See [React issue #15893](https://github.com/facebook/react/issues/15893) for a brief overview of how this differs from similar older proposals. This proposal ensures that custom hooks remain composable and independent from one another: it focuses on allowing developers to do more with the custom hooks that already exist, without needing to totally restructure their applications.

### DIFF
--- a/text/0000-keyed-hooks.md
+++ b/text/0000-keyed-hooks.md
@@ -1,0 +1,299 @@
+- Start Date: 2020-04-18
+- RFC PR:
+- React Issue:
+
+# Summary
+
+A new `useKeyedGroups` primitive hook makes it possible to composably consume existing (custom) hooks conditionally or in loops.
+
+# Motivation & Example
+
+A basic example follows, which shows the shortcomings of the existing Rules of Hooks.
+Then, a demonstration of how `useKeyedGroups` solves these shortcomings.
+
+## The Setup
+
+Consider the following existing custom hook, which enables any component to listen to use the `EventSource` listener API:
+
+```javascript
+function useEventSource(url, callback) {
+  const callbackRef = React.useRef(callback);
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks for the RFC! This is _really_ well-written.

>We either have to split our one component into multiple, which may involve a restructuring of our application, or we have to come up with a brand-new (much, much more complicated) useMultipleEventSource custom hook.

In general, splitting a component into multiple _is_ the idiomatic way to do a lot of things in React. It doesn't make everyone happy (e.g. #197) but it's the way it goes, and all features are built with the assumption that eventually you'll want/need to split it up.

The feature of "keyed Hooks" is something we considered early on and even (if I recall correctly) implemented. However, we've thought of it only as a compiler target (to support component inlining) and not a user-facing API. It is, in our opinion, a bit too complicated and advanced to be used widely, and not worth it if it's used rarely.

In our experience, by the time you want something like this and you also _can't_ split a component up, there's a big risk that having those Hooks stay isolated wouldn't have worked out anyway. E.g. you might begin to want to control them together in some way, accumulate information between them, compose their implementation details, and so on. So we think the subset of cases that would be well-served by this abstraction is just to small to justify the complexity (both in implementation and API surface) that it brings. That's the reason we rejected it last time, and I don't think anything made this more compelling yet.

PR #170, opened 2020-06-14
Component Behaviour approach (more flexible logic reuse without hooks or with hooks)

sergeysibara:
From https://github.com/facebook/react/issues/17792 issue.
I suggest considering a several changes in structure of React components.
1. Make component a configurable container without hard-coded render function and hard-coded custom logic.
2. Add ability adding and removing logical blocks in components (at least before component creating).

I made an implementation on classes with examples. It is described in the Detailed design section.
Example of using implementation:
```js  
class CounterBehaviour extends BaseBehaviour {
  defaultState = { count: 0 };
  passedToRender = {
    setCount: value => {
      this.setState({ count: value });
    }
  };
}

const CounterExample = createContainerComponent("CounterExample", {
 behaviours: [{ behaviour: CounterBehaviour }],
 render: ({ count, setCount }) => (
   <>
     <h3>Counter Example</h3>
     <p>You clicked {count} times</p>
     <button onClick={() => setCount(count + 1)}>Click me</button>
   </>
 )
});
```
It’s hard to see a benefits on an example of simple component. Below is a link to the implementation with examples of more complex components:
https://codesandbox.io/s/entity-component-behaviour-approach-in-react-zo2yy

Even if you will refuse both suggests, there are several other interesting ideas in my implementation (using logical blocks as directives, grouping props by logical blocks).
I hope they will be useful.

### DIFF
--- a/text/0000-entity-component-behaviour-approach-implementation-description.md
+++ b/text/0000-entity-component-behaviour-approach-implementation-description.md
@@ -3,7 +3,8 @@ I do not propose a concrete implementation. You can to implement it using object
 Here is a description of my implementation with using class-based React API.
  
 Link to my implementation with examples using the Entity Component-Behaviour approach:   
-https://codesandbox.io/s/entity-component-behaviour-approach-in-react-zo2yy
+https://codesandbox.io/s/entity-component-behaviour-approach-in-react-zo2yy   
+[Github: source code](https://github.com/sergeysibara/entity-component-behaviour-approach-in-react/tree/master/example)
  
 **All code of implementation is inside the `core` folder.**   
 **Other folders contain only examples.**
--- a/text/0000-entity-component-behaviour-approach-implementation-description.md
+++ b/text/0000-entity-component-behaviour-approach-implementation-description.md
@@ -267,7 +267,6 @@ createContainerComponent("ConcreteComponent", {
      initData: { // any custom data that will passed to init method of a behaviour.
        name: “MyBehaviourName”, // name for this behaviour. (optional)
        defaultState: {}, // data that will passed in defaultState of Behaviour. (optional)
-       wrapRenderData: wrapperFunction; // (optional)
        //any custom data
      }
    },
--- a/text/0000-entity-component-behaviour-approach-implementation-description.md
+++ b/text/0000-entity-component-behaviour-approach-implementation-description.md
@@ -1,0 +1,317 @@
+# Detailed design (part 2)
+I do not propose a concrete implementation. You can to implement it using objects, classes API, hooks API.
+Here is a description of my implementation with using class-based React API.
+ 
+Link to my implementation with examples using the Entity Component-Behaviour approach:   
+https://codesandbox.io/s/entity-component-behaviour-approach-in-react-zo2yy
+ 
+**All code of implementation is inside the `core` folder.**   
+**Other folders contain only examples.**
+ 
+**Main programming entities:**
+- `Container` component;
+- `config` (configurable object). Used in `Container` components; 
+- `Behaviour` logical block;
+- `render` function.
+
+**Data flow diagram:**
+
+![](https://raw.githubusercontent.com/sergeysibara/entity-component-behaviour-approach-in-react/master/content/data-flow.png)
+___
+ 
+This document also provides examples of solutions for some react problems and some useful features, such as :
+- creating `Container` for logical blocks (`Description` paragraph 1-3)
+- moving `render` function outside component (`Description` paragraph 6)
+- creating logical blocks (`Description` paragraph 9-11)
+- removing/adding logical blocks in runtime (`Description` paragraph 8)
+- props grouping by logical blocks (`Description` paragraph 12)
+- code reuse like Vue directives - adding logical block through prop in JSX (`Description` paragraph 12-13)
+
+## Description
+**1.** I created the `ContainerComponent` class which inherits from `React.Component` that to change the behavior of all future custom components.   
+```jsx
+ContainerComponent extends React.Component
+```
+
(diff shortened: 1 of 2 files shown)


leodr:
The code in your example is not valid JavaScript code, the variable name is missing.

```
const createContainerComponent("CounterExample", {
    ...
});
```

This pattern also seems pretty similar to [higher-order components](https://reactjs.org/docs/higher-order-components.html), but correct me if I'm wrong.

You could extract the code from your sandbox into a NPM package, why should it be part of the React core?

sergeysibara:
Thank you!
I added missed variable.


> This pattern also seems pretty similar to higher-order components, but correct me if I'm wrong.   

You are wrong. HOC adds new components to a component hierarchy.
This pattern does not add new components to a component hierarchy.
This pattern let to use reusable objects similarly as custom hooks let to use reusable functions.

### For a clearer explanation.
**HOС**  - used vertical / (multi layered) composition. Its like using the decorator pattern many times for one component.
- ComponentAWithLogic (HOC1)

- - ComponentBWithLogic (HOC2)

- - - ComponentC

**Custom hooks** - used horizontal composition (no new components are added to the component hierarchy). It's like using the list of the strategy pattern. But hooks cannot be replaced.
- ComponentA
- - **Unchanged list** of  reusable functions (custom hooks):
  FunctuonWithLogic1 (hook1)
  FunctuonWithLogic2 (hook2)

**This pattern**  - used horizontal composition (no new components are added to the component hierarchy). It's like using the list of the strategy pattern. **Any reusable object can be replaced (including in runtime).**
- ComponentA
- -  **Dynamic list** of  reusable objects (behaviours):
     ObjectWithLogic1 (behaviour1)
     ObjectWithLogic2 (behaviour2)


**from documentation:** https://reactjs.org/docs/hooks-intro.html#its-hard-to-reuse-stateful-logic-between-components
_Hooks allow you to reuse stateful logic without changing your component hierarchy._ 
And this pattern complements the statement above:
_Behaviours allow you to reuse stateful logic without changing your component hierarchy **and without changing your component**._

> You could extract the code from your sandbox into a NPM package, why should it be part of the React core?

I don't have enough free time to maintain NPM package. I just suggested looking at this approach and its benefits. And I wanted to show that a powerful composition can be used in class-based components. There are only about 300 lines of code in my implementation. I think more code was changed/added to implement hooks in React.
It would be great if someone made an experimental NPM package. Then the community will get a composition in class-based components, similar to the custom hooks.

Ayc0:
I think what's you're proposing is very close to the [mixins React use to support](https://reactjs.org/docs/react-without-es6.html#mixins). But they were removed because they were [considered harmful](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html). One of the reasons is because, when looking at your example, you're creating 2 values `count` and `setCount` in the render. But if you were adding another `behavior`, this new behavior could also define those, overriding the previous values, and making those 2 behaviors completely unusable together.

sergeysibara:
I knew mixins is harmful before React was created)
Mixins look similar, but implementation of the behaviors is different from implementation of mixins.
Behaviors is not mixins and is not HOC) 
Behaviors is most likely an extended variation of the Strategy pattern, which uses a list of strategies with a common interface.

Mixins do not work well because they mix fields and methods from several objects to one object.
Behaviors and hooks use composition like the Strategy pattern.
Component + mixins = one big object with name conflicts.
Component + hooks or behaviors = multiple separate functions / objects where name conflicts can be easily fixed.

By the way, a similar approach was used in Marionette framework:
https://marionette.gitbooks.io/marionette-guides/content/en/behaviors/index.html

> One of the reasons is because, when looking at your example, you're creating 2 values count and setCount in the render. But if you were adding another behavior, this new behavior could also define those, overriding the previous values, and making those 2 behaviors completely unusable together.    

You can write a function to rename conflicting names before passing any data to render function using mapToRenderData option.
A solution to the problem is described in text from files of this pull request:
> **17.** `mapToRenderData` option in `config` object of component.   
The `mapToRenderData` is like to 'optionMergeStrategies' в Vue:   
https://vuejs.org/v2/guide/mixins.html#Custom-Option-Merge-Strategies
> 
>Usually `mapToRenderData` no need to use. In this case all data from a behaviours will merged like mixins.   
To avoid a names conflict you can to use the prepared function `mapToIsolatedRenderDataObjects` from `mapToRenderDataStrategies.js` or to write own function.   
> 
> ```jsx 
> import { mapToIsolatedRenderDataObjects } from "../core/mapToRenderDataStrategies";
> createContainerComponent("ConcreteComponent", {
>  render: ...
>  behaviours: ... // multiple behaviours   
>  mapToRenderData: mapToIsolatedRenderDataObjects,
> }
> ```
> 
> then in the `render` instead of simple arguments like
> 
> ```jsx 
> const render = ({
>  x,
>  y,
>  isLoading,
>  title,
>  body,
>  props
> }) => (
> ```
> 
> will passed objects (the object name is taken from the `name` field of a Behaviour), for example:
> 
> ```jsx 
> const render = ({
>  behaviour1Data,
>  behaviour2Data,
>  props
> }) => (
> ```
> 
> For more complex cases, you can to write own mapToRenderData function.

gaearon:
If you like this style of coding, I suggest you to look at [recompose](https://github.com/acdlite/recompose).

From the examples I'm seeing in this RFC, you're trying to break the logic down in small pieces. However, that introduces a lot of indirection. The React philosophy is to try to avoid such indirections in rendering code, and to prefer code that's direct and "to the point". 

There are use cases where a "behavior composition" system like this might make sense. This reminds me of some of the code I've seen that manages a highly customizable modal window framework (which has many behaviors, like focus behavior, window resize behavior, and so on). It worked well there because of the specific use case, but such generalization is not broadly useful in the product code. So when you _need_ to use it, a userland solution like yours is probably the best bet. I don't see a need or a benefit to building this into React.

PR #172, opened 2020-06-30
Anti-tamper mechanism

uncttao:


theKashey:
That's more sounds like the job for CSP and some information added into the dev console to prevent it from misuse.

uncttao:
> That's more sounds like the job for CSP and some information added into the dev console to prevent it from misuse.

How exactly to prevent the risk? Could you elaborate more?

theKashey:
CSP is not letting to injects or evaluate third party scripts and styles, keeping only one vector of attack - dev console, which could be managed in non-technical ways, just asking a customer to close it.

uncttao:
> CSP is not letting to injects or evaluate third party scripts and styles, keeping only one vector of attack - dev console, which could be managed in non-technical ways, just asking a customer to close it.

As far as I know, there is no clean way to detect console opening. Plus, scammers can just do some editing to delete the warning message or blocker---whatever you put there---and continue the scam. This is, of course, can be a quick working patch: Maybe you can flash a recurring warning to the customer, surely making the scammers' job much more difficult.

But it would just be nice this security vulnerability can be addressed at the code level once in for all.

phryneas:
When someone has active control over the user's machine, they can
* just install a keylogger
* open the URL of a similar-looking site (nobody who would be at this point giving a scammer control would notice that)
* add a proxy to modify code on transmission, potentially stripping out/replacing the hardened copy of react with a more modifiable one
* quickly start a program that paints over the browser
* simply download the website as a static html file, modify that and open that in the browser

And I'm not even getting started on possibilities, that's just what I can think of in seconds. A scammer would have better ideas.

What this would prevent against would probably be self-XSS: some people copy/pasting code into the devtools they are told to. And in that case, the warning message in the console should be quite persuasive. As an example, go to facebook and open the devtools there. If I do, I am greeted with this:
image


On the other side, a website hardened like this would probably break tools like password safes, a11y helpers and other browser plugins like in-site translators.

In the end: This is not something that a web framework can prevent. Google would have to start shipping a browser version without devtools and without extensions - and that's not very likely.

crypt096 (approved):

gaearon:
I agree this is out of scope for React. There is also a technical reason for why this is difficult to do: React *can't actually recover* from a DOM that's been tampered with. Reconciliation doesn't compare DOM against the tree; it compares the previous tree with the next tree. This is in some cases required (reading from DOM can be lossy), is faster, and avoids a lot of complexity that we would otherwise have to add (even if it were behind some flag). In addition, manual DOM manipulation is still common (for edge cases), both in product code and in third party component. So a mode like this would likely be turned off by the developer anyway because it would prevent them from doing regular legit tasks. I agree that displaying a prominent message in the console is probably the most pragmatic way to approach this today.

PR #183, opened 2020-10-17
RFC: Declared Dependencies

dantman:
This RFC is for the proposal/discussion of a potential common convention.

I regularly see devs (myself, library authors, etc) disabling the `react-hooks/exhaustive-deps` rule on specific hooks because they want to explicitly declare the conditions which one of the deps actually changes in a meaningful way under. There are also some scenarios where you still want the value to change but do not want effects to re-run.

I would like to create alternative solutions to disabling the lint rule for an entire deps list just to declare the change conditions of one dep.

### [View formatted RFC](https://github.com/dantman/rfcs/blob/declared-dependencies/text/0000-declared-dependencies.md#summary)

### DIFF
--- a/text/0000-declared-dependencies.md
+++ b/text/0000-declared-dependencies.md
@@ -1,0 +1,122 @@
+- Start Date: 2020-10-17
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Provide a hook that can be used to declare the relevant condition's a value's "equality" should change from the perspective of hooks using that value.
+
+This RFC is about a potential common convention. I do not anticipate this RFC making a change to React. However I do wish to start a discussion about whether this pattern is a good idea, or what the best solution to these issues are.
+
+# Basic example
+
+```js
+// Example hook representing any auth system someone may use, returns the current user's auth token and userId
+const { token, userId } = useLoginState();
+
+// Proposed hook
+// - apiTokenRef will always contain the current value in `token`
+// - however apiTokenRef will be referentially identical unless `userId` changes
+const apiTokenRef = useDependencyDeclaration(token, [userId]);
+
+// Crude example representing an API fetching effect
+// Because apiTokenRef does not change as long as userId is the same, fetchSelf will not be re-executed if the token is just refreshed
+// However because it does change when userId changes, fetchSelf will re-execute if the user signs in to a different account
+const [user, setSelf] = useState(null);
+const fetchSelf = useCallback(async () => {
+  const response = await fetch('/me', {
+    headers: {
+      Authorization: `Bearer ${apiTokenRef.current}`
+    }
+  });
+  setSelf(await response.json());
+}, [apiTokenRef]);
+useEffect(() => {
+  fetchSelf();
+}, [fetchSelf]);
+```
+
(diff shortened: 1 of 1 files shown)


Ayc0:
I think this proposal breaks one of react's base principle:
> `useRef` returns a mutable ref object whose `.current` property is initialized to the passed argument (`initialValue`). **The returned object will persist for the full lifetime of the component**.

And in your RFC, you're suggesting to return "refs" but that are recreated when one (or more) of the dep changes. IMO as it breaks this base principle, it shouldn't be called a ref.

dantman:
> I think this proposal breaks one of react's base principle:
> 
> > `useRef` returns a mutable ref object whose `.current` property is initialized to the passed argument (`initialValue`). **The returned object will persist for the full lifetime of the component**.
> 
> And in your RFC, you're suggesting to return "refs" but that are recreated when one (or more) of the dep changes. IMO as it breaks this base principle, it shouldn't be called a ref.

The principle you are referring to only applies to the [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) hook. It is simply a declaration of that specific hook's behaviour, not a principle that refs follow. It does not apply to other object refs like those created by [`createRef`](https://reactjs.org/docs/react-api.html#reactcreateref).

theKashey:
The root cause is the "all or nothing" presence of eslint rule. It's or working, complaining about wrong dependencies, or not working at all.
What is the root cause of it? How one can solve it in a better way:
- of one uses dependency, which is not used inside a function, and there are certain use cases when it's usefull, I want only one dep to be __explicitly added__ as an extra dep.
- of one is not declaring used dep, I want to mark it in a special way
  - in some cases React known about stability of used variables (like `setState`). It would be nice to be mark some _user-space_ variables in the same way

While the result might contain twice as much comments, it can lead to a better understand of the result, and reasons behind for both for eslint rule and the end developer.
```tsx
config = React.useMemo(
  () => configToObject(config),
  [JSON.stringify(config)]
)

// ⬇️
config = React.useMemo(
   // react-ignore-usage comment: do not react on variable change
  () => configToObject(config),
  [
   // react-additional-dependency comment: we use string representation of config to make result less reactive to some changes
  JSON.stringify(config)
  ]
)

// ⬇️
/**
 * @stable this variable is "stable" and don't have to be added into any dependency
 */
const config = ...
config = React.useMemo(
  () => configToObject(config), // config is "stable"
  [
   // react-additional-dependency
  JSON.stringify(config)
  ]
)
```

dantman:
@theKashey Sure, that is an alternative I listed (no. 3).

I will note though. One of the original idea for hooks was supposed to be that they could eventually be automatic. i.e. Instead of an eslint rule you have to run, you write code without the deps array at all and the JS compiler or bundler would automatically insert the deps.

Although it looks like instead of going that route of only having auto deps for the useMemo/etc... recently ([React Forget](https://youtu.be/lGEMwh32soc)) it looks like auto-memoizing everything might be the direction things go.

Either way. The one thing to note is that adding comments to the deps array to declare dependencies in some way gets in the way of those auto-deps/memoization plans. But the RFC's way of declaring special cases to deps with a hook instead of comments does play well with auto-deps and auto-memoization.

PR #186, opened 2020-11-22
RFC: contextFor

brandonbloom:


gaearon:
While the Next.js problem seems hacky, it's hard for me to tell what's fundamental about this. It seems somewhat specific to webpack. Is there a writeup somewhere that explains the technical difficulties in more detail than these source links and this issue? In general, React relies on identity in other ways (e.g. component identity as a "compatible type" reconciliation check). If we use a module system we generally rely on it to provide guarantees. It may be difficult for webpack and/or Next to provide these guarantees, but this doesn't seem to be a problem that needs to be solved at the React level.

brandonbloom:
Yeah, even though I wrote this proposal, I'm not convinced that this isn't a Webpack issue. I mostly wrote this to illustrate the problem.

There are also potentially related issues if you do things like render components in a web worker or some other context where something like [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) is utilized. This necessarily serializes data with the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), losing reference identity.

These problems may or may not apply to [Realms](https://github.com/tc39/proposal-realms/blob/main/explainer.md). Search that page for "serialization" and "transfer".

gaearon:
Got it. I think for the normal use case, we can close this, but maybe we can revisit if some other use case comes up. E.g. we do have a remote environment with Server Components (#188). But those will likely necessitate adding some sort of Server Context which itself needs to be serializable. So we'll be drafting that and discussing that separately. If this problem comes up in other scenarios, we should look at them in isolation.

Issue #137, opened 2020-01-06
[Hooks] Abortable useEffect


sag1v:
When ever we use effect with a state update inside an async callback (or after an `await`) we probably want to check that we are not trying to update an unmounted component or worse, we are in a race condition and updating the state with the wrong data.

With hooks, this is fairly simple to achieve using a boolean and a closure:

```jsx
useEffect(() => {
  let current = true;
  asyncOp(arg).then(res => {
    current && setState(res);
  });

  // cleanup
  return () => current = false
}, [arg]);
```  

But doing that in each effect, every time we are dealing with a state update in an async operation can become tedious. 

I played around and created an "abortable" effect, this effect is just a wrapper over `useEffect` and its providing a `status` indicator so the consumer can check if the effect was "cleaned":

```jsx
function useAbortableEffect(effect, dependencies) {
  const status = {}; // mutable status object
  useEffect(() => {
    status.aborted = false;
    // pass the mutuable object to the effect callback
    // store the returned value for cleanup
    const cleanUpFn = effect(status);
    return () => {
      // mutate the object to signal the consumer this effect is cleaning up
      status.aborted = true;
      if (typeof cleanUpFn === "function") {
        // run the cleanup function
        cleanUpFn();
      }
    };
    // not the best way to pass dependencies, but what other options do we have
  }, [...dependencies]);
}
``` 

And the usage:

```jsx
useAbortableEffect(status => {
  asyncOp(arg).then(res => {
    !status.aborted && setState(res);
  });
}, [arg]);
```

I wish that `useEffect` would provide this information out of the box instead of us writing it in "user land".

courtneynguyen:
Shouldn't `status` and  `effect` be included in the dependencies array? Also, since `dependencies` is not being used inside the side effect itself, it shouldn't be included.

sag1v:
@courtneynguyen 
> Shouldn't `status` and `effect` be included in the dependencies array? 
Well `status` is mutable so i'm not sure how effective it will be to include it as dependency, i think its the same as a `ref`.

As for `effect`, again this won't make much sense as it will be a new function reference on each cycle because most if not all the time we will pass `effect` as an inline function.

> Also, since `dependencies` is not being used inside the side effect itself, it shouldn't be included.

We are basically wrapping `useEffect` here so we should keep the same API, if we won't use `dependencies` then the consumers won't be able to sync it with dependencies like they can with the builtin `useEffect`.

ajwhite:
[...]

PR #202, opened 2021-08-25
useEffect status RFC

sag1v:
A status parameter for useEffect/useLayoutEffect callbacks.

There are some discussions in #137  

### [View formatted RFC](https://github.com/sag1v/rfcs/blob/master/text/0000-effect-status.md)

### DIFF
--- a/text/0000-effect-status.md
+++ b/text/0000-effect-status.md
@@ -1,0 +1,92 @@
+- Start Date: 2021-08-25
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Add a status argument to the useEffect/useLayoutEffect callback.
+
+# Basic example
+
+```js
+useEffect(status => {
+  asyncOp(arg).then(res => {
+    !status.aborted && setState(res);
+  });
+}, [arg]);
+```
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


sag1v:
Interesting idea indeed. Thanks

dantman:
> * As AbortSignal is part of the platform and spreading, other libraries will likely start adopting AbortSignal and when they do manual cleanup code can be replaced with just passing the standard AbortSignal to them

😅 Actually, I think I underestimated the spread of this feature. Apparently Chrome 88 already supports AbortSignal on addEventListener.

https://css-tricks.com/using-abortcontroller-as-an-alternative-for-removing-event-listeners/

i.e. This code:

```js
useEffect(() => {
  function paste(e) {
    console.log('Pasted', e);
  }
  window.addEventListener('paste', paste);

  return () => {
    window.removeEventListener('paste', paste);
  };
}, []);
```

If `status` was actually an `AbortSignal`, could be replaced with the following at least for browsers that support it (hopefully there's a polyfill).

```js
useEffect((signal) => {
  function paste(e) {
    console.log('Pasted', e);
  }
  window.addEventListener('paste', paste, { signal });
}, []);
```

dantman:
Looks like #204 includes the following in addition to the AbortSignal arg:
> Accept generator functions so that React can take charge of how async things runs inside the effect.

Which IMHO is enough reason to keep this one around even if this one switches to AbortSignal too. Cause that kind of change is highly likely to stop the RFC from moving on.

PR #204, opened 2021-09-05
RFC: Cancel async effects

Jack-Works:
This proposal proposes to 

- Provide [`AbortSignal`](https://mdn.io/AbortSignal) when executing effects.
- Provide a way to stop async effects automatically.

```js
useEffect(function* () {
    setLoading(true)
    // React automatically stops if the deps array outdated.
    setText(yield getData(props.id))
}, [props.id])
```

```js
useEffect(signal => document.addEventListener('event', () => something(), { signal }))
```

[View rendered RFC](https://github.com/Jack-Works-forks/rfcs/blob/cancel-async-effects/text/0000-cancel-async-effects.md)

### DIFF
--- a/text/0000-cancel-async-effects.md
+++ b/text/0000-cancel-async-effects.md
@@ -29,7 +29,7 @@ function Component(props) {
 Advanced solution: Automatic cancel with [co](https://www.npmjs.com/package/co)-like API
 
 ```js
-useEffect(function* () {
+useEffect(function* (_signal) {
     setLoading(true)
     // React automatically stops if the deps array outdated.
     setText(yield getData(props.id))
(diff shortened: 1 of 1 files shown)


bglgwyng:
I utilized a similar methodology within [this repository](https://github.com/gwyng-dev/react-eff-hook) and ended up here while delving into its history. Another notable benefit we now possess, as opposed to when this RFC was initially suggested, is our alignment with the `using` syntax in TypeScript 5.2 and upcoming ECMAScript versions. I don't believe there's currently a viable means of seamlessly integrating the using syntax with the existing `useEffect` approach.

Jack-Works:
closed because looks like there are no interest of this for one year

PR #206, opened 2021-09-14
#: The setState method that can directly modify the immutable data

zdu-strong:


rickhanlonii:
Hey @zdu-strong, thanks for the submission. It looks like you didn't fill out the rest of the RFC including the proposed design, so I'm going to close this but let us know if you update this to follow the [rfc process](https://github.com/reactjs/rfcs#what-the-process-is) and we can re-open.

PR #209, opened 2021-10-29
RFC: Inject prop to components

behnammodi:
# Summary

Inject prop to all components or specific components

# Basic example

Sometimes we import some modules into every component's file ex: i18n translation module or history module or tracker/logger module

# Motivation

Motivation about this RFS is improvement bundle size by injecting prop, also simply component's file by reducing any unnecessary imports

# Detailed design

Imagine we have something like this for injecting prop to all components

```js
import React, { inject } from 'react';
import history from 'history-package-or-module';
import t from 'i18n-package-or-module';

inject({
  // history helper methods
  push: (...args) => history.push(...args),
  // translation helper method
  t: (...args) => t(...args),
  // GA helper method
  tracker: (...args) => GA.dataLayer.push(...args),
})
```

So we can use this method without importing its modules

```jsx
function Component({ push, t, tracker }) {

  const handleSubmit = () => {
    push('/somewhere');
    tracker('user is going to somewhere')
  }

  return (
    <button onClick={handleSubmit}>{t('submit')}</button>
  );
}
```

Also we can have `excludes` or `includes` as option for just inject to specific component by name like this:

```js
inject({
  // history helper methods
  push: (...args) => history.push(...args),
  // translation helper method
  t: (...args) => t(...args),
  // GA helper method
  tracker: (...args) => GA.dataLayer.push(...args),
}, {
  includes: ['Home', 'SignUp', 'SignIn'] // all of them is component names
})
```

# Drawbacks

I'm not sure about implementation cost, I think this is not hard to learn and also this is no need for any migration plan

# Alternatives

Nothing, or maybe I found nothing

# Adoption strategy

Lucky this hasn't any breaking change

# How we teach this

The term about that is `Injection` also all developer familiars to this concept

# Unresolved questions

Nothing

### DIFF
--- a/text/0000-inject-prop.md
+++ b/text/0000-inject-prop.md
@@ -1,0 +1,85 @@
+- Start Date: 2021-10-30
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Inject prop to all components or specific components
+
+# Basic example
+
+Sometimes we import some modules into every component's file ex: i18n translation module or history module or tracker/logger module
+
+# Motivation
+
+Motivation about this RFS is improvement bundle size by injecting prop, also simply component's file by reducing any unnecessary imports
+
+# Detailed design
+
+Imagine we have something like this for injecting prop to all components
+
+```js
+import React, { inject } from "react";
+import history from "history-package-or-module";
+import t from "i18n-package-or-module";
+
+inject({
+  // history helper methods
+  push: (...args) => history.push(...args),
+  // translation helper method
+  t: (...args) => t(...args),
+  // GA helper method
+  tracker: (...args) => GA.dataLayer.push(...args),
+});
+```
+
+So we can use this method without importing its modules
+
+```jsx
+function Component({ push, t, tracker }) {
+  const handleSubmit = () => {
+    push("/somewhere");
+    tracker("user is going to somewhere");
+  };
+
+  return <button onClick={handleSubmit}>{t("submit")}</button>;
+}
+```
+
+Also we can have `excludes` or `includes` as option for just inject to specific component by name like this:
+
+```js
+inject(
+  {
+    // history helper methods
+    push: (...args) => history.push(...args),
+    // translation helper method
+    t: (...args) => t(...args),
+    // GA helper method
+    tracker: (...args) => GA.dataLayer.push(...args),
+  },
+  {
+    includes: ["Home", "SignUp", "SignIn"], // all of them is component names
+  }
+);
+```
+
+# Drawbacks
+
+I'm not sure about implementation cost, I think this is not hard to learn and also this is no need for any migration plan
+
+# Alternatives
+
+Nothing, or maybe I found nothing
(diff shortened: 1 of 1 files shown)


dantman:
Adding universal default props this way has a number of drawbacks:
- TypeScript types for this are completely broken.
- The injected props will by default apply to 3rd party components too and this could cause things to break if they spread props or happen to share the name of a prop and have a behaviour for when the parent component doesn't pass that prop (the inject will result in it always acting as if the parent passed the prop even when the parent does not pass a prop of that name).
- Component names are not a reliable way of targeting components. Different components can share the same display name. And production compilation can change the display name causing the code to break in production.

> Motivation about this RFS is improvement bundle size by injecting prop, also simply component's file by reducing any unnecessary imports

I don't see how the suggestion reduces bundle size in any way compared to just using context to do the same thing.

```js
import { createContext, useContext } from 'react';
import history from 'history-package-or-module';
import t from 'i18n-package-or-module';

export const InjectContext = createContext()

// ...
<InjectContext.Provider value={{
  // history helper methods
  push: (...args) => history.push(...args),
  // translation helper method
  t: (...args) => t(...args),
  // GA helper method
  tracker: (...args) => GA.dataLayer.push(...args),
}}>{children}</InjectContext.Provider>
```

```js
import { InjectContext } from './app'

function Component() {
  const { push, t, tracker } = useContext(InjectContext)

  const handleSubmit = () => {
    push('/somewhere');
    tracker('user is going to somewhere')
  }
  
  return (
    <button onClick={handleSubmit}>{t('submit')}</button>
  );
}
```

The only space you appear to save is reducing the `import { InjectContext } from './app'` and the use of `createContext`. Which is of little value. In either case you're still importing `history-package-or-module` and `i18n-package-or-module` so the bulk of things are still in the bundle.

Meanwhile you may actually de-optimize or increase your bundle size. By importing `history` and `t` in a core part of your main bundle you've now made it required in order to run the main bundle even if it's not used. If the `Component` file were to import these itself and your `Component` happened to only be used in something imported by `import()` then your bundler could decide to split them out and would be able to finish executing the main bundle without having to first download those two packages.

Of course because you are discarding the contextual nature of context there doesn't appear to be any value to using component props for this. You could avoid imports and get basically the exact same thing with similar issues by just using a global variable.

```js
import history from 'history-package-or-module';
import t from 'i18n-package-or-module';

window.injected = {
  // history helper methods
  push: (...args) => history.push(...args),
  // translation helper method
  t: (...args) => t(...args),
  // GA helper method
  tracker: (...args) => GA.dataLayer.push(...args),
}
```

```js
function Component() {
  const { push, t, tracker } = window.injected

  const handleSubmit = () => {
    push('/somewhere');
    tracker('user is going to somewhere')
  }
  
  return (
    <button onClick={handleSubmit}>{t('submit')}</button>
  );
}
```

behnammodi:
@dantman Thanks a lot, your solution is awesome and I agree 👍🏻  , I will close this RFC

PR #210, opened 2022-02-06
RFC: Inner Hooks

tkow:
This rfc is child hooks extended can be passed from a special prop by named inner Hooks.

[view RFC](https://github.com/tkow/rfcs/blob/main/text/0000-inner-hooks.md)


tkow:
I also made library realized this concept.
See if you interested in it see [this link](https://www.npmjs.com/package/react-inner-hooks-extension).

PR #211, opened 2022-02-07
RFC: useDynamicStateFactory

tkow:
This rfc is about child's state can be defined in children components using method in [this innerHooks proposal](https://github.com/reactjs/rfcs/pull/210) and easily obtain controlling whole state by dynamically generating state and handler as much as you need as you add more components. It is also easily to remove because a partial state of whole state mutation is not depend on any other components without itself in my proposal way.

My proposal is  simulated by hoc in [Code Sand Box](https://codesandbox.io/s/react-typescript-forked-5hq21?file=/src/App.tsx).

[view RFC](https://github.com/tkow/rfcs/blob/feat/use-dynamic-state-factory/text/0000-use-dynamic-state-factory.md)

### DIFF
--- a/text/0000-use-dynamic-state-factory.md
+++ b/text/0000-use-dynamic-state-factory.md
@@ -1,0 +1,406 @@
+- Start Date: 2022-0207
+- RFC PR:
+- React Issue:
+
+# Summary
+
+A new hook create component state and it can be passed decendants components via hooks.
+
+# Basic example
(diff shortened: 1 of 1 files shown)


tkow:
I close this issue because this proposal's rejected in the past.
I found [this article](https://overreacted.io/why-do-hooks-rely-on-call-order/).

PR #212, opened 2022-03-23
RFC: Intent to ship React 18

rickhanlonii:
This RFC describes the changes we intend to ship in React 18 🎉

### [View formatted RFC](https://github.com/reactjs/rfcs/blob/react-18/text/0000-react-18.md)

### DIFF
--- a/text/0000-react-18.md
+++ b/text/0000-react-18.md
@@ -46,7 +46,7 @@ setTimeout(() => {
   // React will render twice, once for each state update (no batching)
 }, 1000);
 
-`// After: updates inside of timeouts, promises,
+// After: updates inside of timeouts, promises,
 // native event handlers or any other event are batched.`
 setTimeout(() => {
   setCount(c => c + 1);
--- a/text/0000-react-18.md
+++ b/text/0000-react-18.md
@@ -63,7 +63,7 @@ Transitions will opt in to concurrent rendering, which allows the update to be i
 
 ### New Suspense features
 
-These features are outlined in a separate RFC.
+[These features are outlined in a separate RFC](https://github.com/reactjs/rfcs/pull/213).
 
 ### New client and server rendering APIs
 
--- a/text/0000-react-18.md
+++ b/text/0000-react-18.md
@@ -1,0 +1,189 @@
+- Start Date: 2022-03-23
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# RFC: Intent to ship React 18
+
+## Summary
+
+This RFC describes the changes we intend to ship in React 18 including:
+
+* New feature: automatic batching
+* New feature: transitions
+* New Suspense features
+* New client and server rendering APIs
+* New Strict Mode behaviors
+* New hooks
+
+*This RFC is different than the process we typically do because it is the result of years of research into concurrency, Suspense, and server rendering. **We'd like to get one final round of broad public feedback from the community before shipping in case there are new concerns that have not been discussed before.** All of what is posted here was designed and discussed over the last year in the React 18 Working Group ([available here](https://github.com/reactwg/react-18/)). You can consider the Working Group to be a part of the RFC, so please feel free to quote and discuss any content from it when commenting on the RFC here.*
+
+## Motivation
+
+The goal of React 18 is to release the new concurrent rendering APIs that we are building all future React features on top of including Suspense, transitions, streaming server rendering, server components, and more. This release does not include all of the features on day one, but by upgrading your apps fully to React 18 using the new client and server rendering APIs, you will be ready to gradually adopt new future features as they become available.
+
+While the focus of this release is on upgrading to the new rendering engine, we’re also making a few long-needed improvements (such as automatic batching, Suspense on the server, and `useId`), adding new hooks for libraries to support concurrent rendering (such as `useSyncExternalStore` and `useInsertionEffect`), and a few new concurrent features (such as `useTransition` and `useDeferredValue)`.
+
+## Detailed Design
+
+### New feature: Automatic Batching
+
+Batching is when React groups multiple state updates into a single re-render for better performance. Without automatic batching, we only batched updates inside React event handlers. Updates inside of promises, setTimeout, native event handlers, or any other event were not batched in React by default. With automatic batching, these updates will be batched automatically:
+
+```js
+// Before: only React events were batched.
+setTimeout(() => {
+  setCount(c => c + 1);
+  setFlag(f => !f);
+  // React will render twice, once for each state update (no batching)
+}, 1000);
+
+`// After: updates inside of timeouts, promises,
+// native event handlers or any other event are batched.`
+setTimeout(() => {
+  setCount(c => c + 1);
+  setFlag(f => !f);
+  // React will only re-render once at the end (that's batching!)
+}, 1000);
+```
+
+* [Automatic batching for fewer renders in React 18](https://github.com/reactwg/react-18/discussions/21)
+
+### New feature: transitions
+
+Transitions allow users to tell React that the contained updates are less urgent than other work. For example, handling user text input is more urgent than UI transitions. To support these feature, we’re adding two new APIs:
+
+* `useTransition`: a hook to start transitions, including a value to track the pending state.
+* `startTransition`: a method to start transitions when the hook cannot be used.
+
+Transitions will opt in to concurrent rendering, which allows the update to be interrupted. If the content re-suspends, transitions also tell React to continue showing the current content while rendering the transition content in the background (see the Suspense RFC for more info).
+
+* [New feature: startTransition](https://github.com/reactwg/react-18/discussions/41)
+* [Real world example: adding startTransition for slow renders](https://github.com/reactwg/react-18/discussions/65)
+* [Patterns for startTransition](https://github.com/reactwg/react-18/discussions/100)
+
+### New Suspense features
+
+These features are outlined in a separate RFC.
+
+### New client and server rendering APIs
+
+In this release we took the opportunity to redesign the APIs we expose for rendering on the client and server. These changes allow users to continue using the old APIs in React 17 mode while they upgrade to the new APIs in React 18.
+
+On the client, these APIs are:
+
+* `react-dom/client`
+    * `createRoot`
+    * `hydrateRoot`
+* `react-dom/server`
+    * `renderToPipeableStream`
+    * `renderToReadableStream`
+
+Additionally, we’ve added limited support for Suspense to:
+
+* `renderToString`
+* `renderToStaticMarkup`
+
+And have deprecated:
+
+* `renderToNodeStream`
+
+Note: these APIs are designed to be upgraded at the same time. For example, upgrading to `createRoot` on the client, but keeping `renderToString` on the server is unsupported.
+
+* [Replacing render with createRoot](https://github.com/reactwg/react-18/discussions/5)
+* [Moving create/hydrateRoot to react-dom/client](https://github.com/reactwg/react-18/discussions/125)
+* [Upgrading to React 18 on the client](https://github.com/reactwg/react-18/discussions/6)
+* [Upgrading to React 18 on the server](https://github.com/reactwg/react-18/discussions/22)
+
+### New Strict Mode behaviors
+
+In the future, we’d like to add a feature that allows React to add and remove sections of the UI while preserving state. For example, when a user tabs away from a screen and back, React should be able to immediately show the previous screen. To do this, React would unmount and remount trees using the same component state as before.
+
+This feature will give React apps better performance out-of-the-box, but requires components to be resilient to effects being mounted and destroyed multiple times. Most effects will work without any changes, but some effects assume they are only mounted or destroyed once.
+
+To help surface these issues, React 18 introduces a new development-only check to Strict Mode. This new check will automatically unmount and remount every component, whenever a component mounts for the first time, restoring the previous state on the second mount.
+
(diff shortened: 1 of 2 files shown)


text/0000-react-18.md
```diff
@@ -0,0 +1,198 @@
...
+
+```js
+// Before: only React events were batched.
+setTimeout(() => {
+  setCount(c => c + 1);
+  setFlag(f => !f);
+  // React will render twice, once for each state update (no batching)
+}, 1000);
+
+`// After: updates inside of timeouts, promises,
```
eps1lon, line 49:
```suggestion
// After: updates inside of timeouts, promises,
```

michaelwarren1106:
So since i'm not seeing any info about native support for web components in the RFC, does that mean that web component native support wont ship in React 18? If so, is there a rationale for that decision?

gaearon:
>So since i'm not seeing any info about native support for web components in the RFC, does that mean that web component native support wont ship in React 18? If so, is there a rationale for that decision?

It's available in the `@experimental` builds (and will continue to be available there), but we are going to put it into the next major stable release together with *other* DOM-related breaking changes that we've meant to do for years. It makes sense to batch them together. See https://github.com/facebook/react/issues/11347#issuecomment-1063490040 for more details.

holloway:
If a library wants to support React 16/17/18 is there a recommended way of providing a `useId` fallback? Would it look something like (untested code)?:

```javascript
import { useId } from 'react';

const useMyId = () => useId ? useId() : useState(uuid());
```

text/0000-react-18.md
```diff
@@ -0,0 +1,198 @@
...
+Additionally, we’ve added limited support for Suspense to:
+
+* `renderToString`
+* `renderToStaticMarkup`
+
+And have deprecated:
+
+* `renderToNodeStream`
+
+Note: these APIs are designed to be upgraded at the same time. For example, upgrading to `createRoot` on the client, but keeping `renderToString` on the server is unsupported.
```
eps1lon, line 99: This is new to me. I was under the impression that `renderToString` + `hydrateRoot` would work (assuming you meant `hydrateRoot` not `createRoot`).

How would I deploy a statically rendered site that's hydrateable with new features like selective hydration?

What does "unsupported" mean here concretely?

text/0000-react-18.md
```diff
@@ -0,0 +1,198 @@
...
+`useInsertionEffect` is a new hook that allows CSS-in-JS libraries to address performance issues of injecting styles in render. Unless you’ve already built a CSS-in-JS library we don’t expect you to ever use this. This hook will run after the DOM is mutated, but before layout effects read the new layout. This solves an issue that already exists in React 17 and below, but is even more important in React 18 because React yields to the browser during concurrent rendering, giving it a chance to recalculate layout. 
+
+* [Library Upgrade Guide: <style> (most CSS-in-JS libs)](https://github.com/reactwg/react-18/discussions/110)
+* [Add useInsertionEffect](https://github.com/facebook/react/pull/21913)
+
+#### useId
+
+`useId` is a new hook for generating unique IDs on both the client and server, while avoiding hydration mismatches. This solves an issue that already exists in React 17 and below, but it’s even more important in React 18 because of how our streaming server renderer delivers HTML out-of-order. 
+
+* [Intent to Ship: useId](https://github.com/reactwg/react-18/discussions/111)
```
pacocoursey, line 162: The IDs generated by `useId` are not valid as CSS selectors, so one cannot target it using `.querySelector` or inside of CSS (`getElementById` still works):

CleanShot 2022-03-23 at 16 26 47@2x

Are there plans to update the format to be compatible with CSS selectors, or are consumers expected to transform it themselves?

rickhanlonii:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1076784999

> If a library wants to support React 16/17/18 is there a recommended way of providing a useId fallback? 

There are libraries that do this pre-18, such as [`auto-id`](https://reach.tech/auto-id/) that you can use in 17 and below. We introduced `useId` in 18 because it's difficult for those libraries to support out-of-order streaming in 18. I would anticipate that existing libraries will switch to using `useId` if it's available so that all versions are supported.

text/0000-react-18.md
acdlite: > What does "unsupported" mean here concretely?

It mostly means Suspense doesn't work. Anything that suspends will get replaced by a fallback and de-opt to client rendering. Whereas in the new APIs, server rendering will continue once the data resolves.


> How would I deploy a statically rendered site that's hydrateable with new features like selective hydration?

The new APIs can be used for static site generation. Although the main use case is to stream the response, you can also buffer that response and server it from a cache on subsequent requests.

acdlite: The key part that enables selective hydration is that we output out-of-order HTML chunks and stitch them together using inline script tags. This works really well with streaming, but it still works even if the HTML is delivered from a cache. Usually you'll want to stream the cached response, anyway, so in practice it's not even that different except for the fact that the data loads much faster.

danmouracunha (approved):

sonatard:
It looks like the Strict mode change is for the Offscreen API, will the Offscreen API be released in React 18?

gaearon:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1076996808

> It looks like the Strict mode change is for the Offscreen API, will the Offscreen API be released in React 18?

Not in 18.0 but it’s one of the things we’ll be focusing on after 18.0 is out. So it will be out likely in some 18.x release if everything goes well.

VanTanev:
Sorry if this is addresed and I failed to find it, but is there a part of the RFC that expands on `startTransition()` vs `useDeferredValue()` - when to reach for one over the other and best practices for `useDeferredValue()`? I feel like `startTransition()` is explained very well, but `useDeferredValue()` (which I think in a sense is an alternative to `startTransition()` for simple usecases) hasn't been expanded upon.

moshfiqrony:
I really liked the `Automatic Batching` feature. Is it something that we can turn or off for a project through `package.json` or it will be enable always? Can there be any situation where it will act like a bug?

Moshyfawn:
> I really liked the `Automatic Batching` feature. Is it something that we can turn or off for a project through `package.json` or it will be enable always? Can there be any situation where it will act like a bug?

You can use the `flushSync` `ReactDOM` API to opt-out of batching.

From [Automatic batching for fewer renders in React 18](https://github.com/reactwg/react-18/discussions/21) write up:

> ## What if I don’t want to batch?
> 
> Usually, batching is safe, but some code may depend on reading something from the DOM immediately after a state change. For those use cases, you can use `ReactDOM.flushSync()` to opt out of batching:
> ```js
> import { flushSync } from 'react-dom'; // Note: react-dom, not react
> 
> function handleClick() {
>   flushSync(() => {
>     setCounter(c => c + 1);
>   });
>   // React has updated the DOM by now
>   flushSync(() => {
>     setFlag(f => !f);
>   });
>   // React has updated the DOM by now
> }
> ```

gaearon:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1077357914

>is there a part of the RFC that expands on startTransition() vs useDeferredValue() - when to reach for one over the other and best practices for useDeferredValue()? 

No, this is an accidental omission. We'll try to write up something soon. The quick tldr is that `useDeferredValue` is similar to `startTransition` but with an API similar to userland `useDebouncedValue` helpers. It's useful when the value comes "from above" and you don't actually have control over the corresponding `setState` call. It is less clunky to use than `startTransition` but it has pitfalls similar to userland `useDebouncedValue` helpers (e.g. you have to use memoization for child components to get any benefit, and you need to be cautious to not pass always-new values into it to avoid loops).

gaearon:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1077485530

>I really liked the Automatic Batching feature. Is it something that we can turn or off for a project through package.json or it will be enable always?

No, you can't configure it. As @Moshyfawn described in https://github.com/reactjs/rfcs/pull/212#issuecomment-1077817390 though, you can opt out of it for a specific `setState` call if you need to observe the modified DOM synchronously.

nafg:
Is there any way to get the benefits of the server integration without a node.js backend? I guess the applies to traditional SSR but is there a wire protocol spec (not sure if this is just a matter of a JSON schema or something more)? If not I think it would be great if there was. Then there could be backends in other languages that support "isomorphic apps." For instance I develop in Scala, using Scala.js to compile the frontend and scalajs-react to build React UIs. But right now the only way I know of to get SSR in Scala involves a complicated setup. If the new server features, as well as traditional SSR, would have a wire spec then alternative backends could be written in other languages.

gaearon:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1078140656

The main value proposition of both SSR and Server Components is that you get to run *the same own component code* on the backend. In principle, nothing stops you from emulating the payload React generates (e.g. by running Scala versions of your components and reimplementing the React server renderer), but there are no current plans to standardize the payload.

nafg:
> Re: [#212 (comment)](https://github.com/reactjs/rfcs/pull/212#issuecomment-1078140656)
> 
> The main value proposition of both SSR and Server Components is that you get to run _the same own component code_ on the backend. In principle, nothing stops you from emulating the payload React generates (e.g. by running Scala versions of your components and reimplementing the React server renderer), but there are no current plans to standardize the payload.

If it was, the Scala vdom DSL could easily be implemented in a way that it would do different things on the JVM backend vs. the JS frontend. So this benefit could still be achieved with a much simpler setup if only the payload was specified.

gaearon:
>If it was, the Scala vdom DSL could easily be implemented in a way that it would do different things on the JVM backend vs. the JS frontend. So this benefit could still be achieved with a much simpler setup if only the payload was specified.

We don't specify the exact payload protocol because there is a lot of benefit to being able to change it midway without bumping versions. So I don't see that happening. In practice, the HTML output we produce is relatively stable and usually matches what you'd expect, aside from a few special instructions (like for Suspense). The most pragmatic solution, if you really wanted to do something like this, is to take our test suite and run your implementation against it. Since you're probably going to want to recreate other behaviors (like actual rendering of components according to React rules!) too. So if you're running against our test suite, you can consider that to be the specification. Then as long as you use the same version of React as the version that the tests were written against, it should work well. Tests are the source of truth for us, too.

text/0000-react-18.md
```diff
@@ -0,0 +1,198 @@
...
+  * Effects are created.
+```
+
+* [Adding Strict Effects to Strict Mode](https://github.com/reactwg/react-18/discussions/19)
+* [How to support strict effects](https://github.com/reactwg/react-18/discussions/18)
+* [Testing with strict effects](https://github.com/reactwg/react-18/discussions/17)
+
+### New hooks
+
+#### useSyncExternalStore
```
jhuleatt, line 142: Is the `useSubscription` hook mentioned in the guide for library maintainers (https://github.com/reactwg/react-18/discussions/70) part of React 18? [ReactFire](https://github.com/FirebaseExtended/reactfire) uses Observables at its core and we're wondering if we should use `useSubscription` or try to integrate with `useSyncExternalStore`.

gaearon: `useSyncExternalStore` is the preferred approach that replaces `useSubscription`.

gaearon:
We've added https://github.com/reactwg/react-18/discussions/129 with an explanation of `useDeferredValue`, and https://github.com/reactwg/react-18/discussions/129#discussioncomment-2440646 with a small concrete example.

gaearon (approved):

rickhanlonii:
Update: I've added `useDeferredValue` to this RFC, my mistake for forgetting it!

rickhanlonii:
Also included here are all the docs we are adding and changing on the website: https://github.com/reactjs/reactjs.org/pull/4499.

For a preview, see: https://18-preview.reactjs.org.

trhinehart-godaddy:
Is there a pattern or plans for adding `startTransition()` into `Suspense` to mimic a "delay" before showing a fallback? Other libs like [react-loadable](https://github.com/jamiebuilds/react-loadable#avoiding-flash-of-loading-component) and [@loadable/component](https://loadable-components.com/docs/delay/) try to address the issue but could only render `null` instead of use a feature like `startTransition()` to keep the existing DOM intact and interactive.

gaearon:
Re: https://github.com/reactjs/rfcs/pull/212#issuecomment-1080828400

>Is there a pattern or plans for adding startTransition() into Suspense to mimic a "delay" before showing a fallback? Other libs like [react-loadable](https://github.com/jamiebuilds/react-loadable#avoiding-flash-of-loading-component) and [@loadable/component](https://loadable-components.com/docs/delay/) try to address the issue but could only render null instead of use a feature like startTransition() to keep the existing DOM intact and interactive.

`statrTransition` *does* work together with `<Suspense>`, as described [here](https://github.com/reactjs/rfcs/blob/suspense-18/text/0000-suspense-in-react-18.md#new-feature-using-transitions-to-avoid-hiding-existing-content). However, you can't have it "wait" and then "switch". Instead, we decide on the behavior depending on whether the `<Suspense>` boundary is _newly created_ or _already shown_. If it's newly created during a transition, we show fallback immediately. But if it was already shown before, we wait for it to complete because we don't want to "hide" already shown content.

There is no way to get a behavior where we first "wait" and _then_ show the fallback after a timeout. We used to provide that ability but removed it. We found that in practice there's no concrete meaningful number one can specify (one second? three seconds? ten? how to choose?) and it ends up being a confusing user experience (the switch to fallback implies some kind of a change but there was no actual change). We think the current heuristic with consistent behavior for "new" vs "existing" boundaries works better in practice.

PR #213, opened 2022-03-23
RFC: Suspense in React 18

gaearon:
In this RFC, we describe a few changes to Suspense that we intend to ship in React 18.

### DIFF
--- a/text/0000-suspense-in-react-18.md
+++ b/text/0000-suspense-in-react-18.md
@@ -104,7 +104,7 @@ Previously, React would do this in a sequence that goes like this:
 1. Fire the `Panel` effects because technically it has "mounted" (even though not fully).
 1. Wait for the `Comments` to be ready.
 1. Then, attempt rendering again.
-1. Remove the `Spinner` content form the DOM.
+1. Remove the `Spinner` content from the DOM.
 1. Place the `Comments` content into the `Panel` content that was already in the DOM.
 1. Remove `display: none` from the `Panel` content.
 
--- a/text/0000-suspense-in-react-18.md
+++ b/text/0000-suspense-in-react-18.md
@@ -210,9 +210,9 @@ function Photos() {
   return (
     <>
       <MyPhotos />
-+      <Suspense fallback={<PhotosGlimmer />}>
++     <Suspense fallback={<PhotosGlimmer />}>
         <TaggedPhotosVerySlow />
-+      </Suspense>
++     </Suspense>
     </>
   )
 }
--- a/text/0000-suspense-in-react-18.md
+++ b/text/0000-suspense-in-react-18.md
@@ -128,7 +128,7 @@ Read https://github.com/reactwg/react-18/discussions/7 for more details on the p
 
 ## New feature: Server-side rendering support with streaming
 
-Previously, if you a component suspends during server rendering, React would throw a hard error. In practice, it meant that apps using server rendering (or built with an SSR framework like Next.js or Remix) could not use Suspense for code splitting in a supported way.
+Previously, if a component suspends during server rendering, React would throw a hard error. In practice, it meant that apps using server rendering (or built with an SSR framework like Next.js or Remix) could not use Suspense for code splitting in a supported way.
 
 We are adding a new server renderer that supports streaming HTML out-of-order. Unlike the old server renderer that synchronously produces a string, the new server renderer produces a stream. That stream starts with the initial HTML that can be flushed early. However, the new renderer is also fully integrated with Suspense, which means that it's able to "wait" for parts of the tree that are not ready, and emit fallback HTML (e.g. spinners) for them. When the content is ready, React emits the content HTML in the same stream along with a small inline `<script>` to insert it in the right place in the original DOM structure. As a result, even if some part of the page is slow on the server, the user sees a progressively loading page with all intentionally designed intermediate loading states — even before client JS loads.
 
--- a/text/0000-suspense-in-react-18.md
+++ b/text/0000-suspense-in-react-18.md
@@ -1,0 +1,276 @@
+- Start Date: 2022-03-23
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This RFC describes several changes that we'd like to make to the behavior of the `<Suspense>` component:
+
+- Behavior change: **Committed trees are always consistent**
+- New feature: **Server-side rendering support with streaming**
+- New feature: **Using transitions to avoid hiding existing content**
+- Behavior change: **Layout effects re-run when content reappears**
+
+The Suspense API itself does not change.
+
+This RFC *does not* add support for data fetching. The changes in this RFC are prerequisites for that future step.
+
+*This RFC is different than the process we typically do because it is the result of years of research into concurrency, Suspense, and server rendering. **We'd like to get one final round of broad public feedback from the community before shipping in case there are new concerns that have not been discussed before.** All of what is posted here was designed and discussed over the last year in the React 18 Working Group ([available here](https://github.com/reactwg/react-18/)). You can consider the Working Group to be a part of the RFC, so please feel free to quote and discuss any content from it when commenting on the RFC here.*
+
+# Basic example
+
+Suspense lets you declaratively specify the loading state for a part of the component tree if it's not yet ready to be displayed:
+
+```js
+<Suspense fallback={<Spinner />}>
+  <Comments />
+</Suspense>
+```
+
+Suspense makes the "UI loading state" a first-class declarative concept in the React programming model. This lets us build higher-level features on top of it. This RFC does not change the Suspense API itself, but it **refines its semantics and adds some new features** on top.
+
+# Motivation
+
+With the initial release in [React 16.6.0](https://reactjs.org/blog/2018/10/23/react-v-16-6.html), Suspense only supported a single use case: [code splitting on the client with `React.lazy`](https://github.com/reactjs/rfcs/pull/64). So even though you could add `<Suspense>` boundaries in your component trees, they were not being used by React for any other purposes. Moreover, you couldn't use them with server rendering. This made their usefulness very limited.
+
+The full motivation has always been to extend the support so that eventually, the same declarative Suspense fallback can handle any asynchronous operations (loading code, data, images, etc). You can learn more about the full vision for Suspense in the [React 18 Keynote](https://youtu.be/FZ0cG47msEk?t=409).
+
+This set of changes is the first step towards making Suspense more powerful.
+
+# Detailed design
+
+We'll first briefly recap how Suspense works, and then describe the changes.
+
+## Recap: How Suspense works
+
+Suspense lets you declaratively specify what React should show when a part of the tree is not yet ready to render:
+
+```js
+<Suspense fallback={<PageGlimmer />}>
+  <RightColumn>
+    <ProfileHeader />
+  </RightColumn>
+  <LeftColumn>
+    <Suspense fallback={<LeftColumnGlimmer />}>
+      <Comments />
+      <Photos />
+    </Suspense>
+  </LeftColumn>
+</Suspense>
+```
+
+Conceptually, you can think of `Suspense` as being similar to a `catch` block. However, instead of catching errors, it catches components "suspending". Any component in the tree can "suspend", which means that it's not ready to render. (The reason is arbitrary, but usually it could be due to missing code, data, etc.)
+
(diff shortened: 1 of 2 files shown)


text/0000-suspense-in-react-18.md
```diff
@@ -0,0 +1,285 @@
...
+
+Consider code like this:
+
+```js
+<div>
+  {showComments && (
+    <Suspense fallback={<Spinner />}>
+      <Panel>
+       <Comments /> 
+     </Panel>
```
eps1lon, line 91: nit: intendation is slightly off
```suggestion
      <Panel>
        <Comments />
      </Panel>
```

text/0000-suspense-in-react-18.md
```diff
@@ -0,0 +1,285 @@
...
+
+We don't think this concern is relevant anymore for several reasons. `UNSAFE_componentWillMount` was marked as "unsafe" in 2018, and most popular open source libraries have long migrated away from it. This issue also only affects components "between" the `<Suspense>` node and the component that actually suspends. So it is very local in scope, and is easy to fix. Finally, Hooks have become a popular alternative to classes, and don't have the same issue. On the other hand, the current behavior [has been causing issues](https://github.com/facebook/react/issues/14536) for using popular component libraries with Suspense. This is why we think now is a good idea to make that change.
+
+Read https://github.com/reactwg/react-18/discussions/7 for more details on the proposed new behavior.
+
+## New feature: Server-side rendering support with streaming
+
+Previously, if you a component suspends during server rendering, React would throw a hard error. In practice, it meant that apps using server rendering (or built with an SSR framework like Next.js or Remix) could not use Suspense for code splitting in a supported way.
```
eps1lon, line 131:
```suggestion
Previously, if a component suspends during server rendering, React would throw a hard error. In practice, it meant that apps using server rendering (or built with an SSR framework like Next.js or Remix) could not use Suspense for code splitting in a supported way.
```

devknoll:
One issue that I saw with Next.js is that some folks are already using Suspense on the client, and such APIs might not work well with server rendering without additional changes.


Would it make sense to officially recommend something like having existing client-only suspense APIs start throwing an error when rendered on the server so that they fallback to client-rendering, until the API can be updated to properly support server-side Suspense?

gaearon:
Re: https://github.com/reactjs/rfcs/pull/213#issuecomment-1076763161

>One issue that I saw with Next.js is that some folks are already using Suspense on the client, and such APIs might not work well with server rendering without additional changes.

Is it the case where people do conditional rendering like `{typeof window !== 'undefined' && <Suspense>...</Suspense>}`, or is this some other case? Currently, it throws on the server if rendered unconditionally. Rendering different content on the client and server based on a condition was never officially supported, so I'm not sure which scenario this refers to.

If it's the scenario I'm thinking of, the fix is usually to do rendering in two passes:

```js
const [mounted, setMounted] = useState(false);
useEffect(() => {
  setMounted(true);
}, []);

<div>
  {mounted && <Suspense>...</Suspense>}
</div>
```

It's not great but would have the same behavior.

>Would it make sense to officially recommend something like having existing client-only suspense APIs start throwing an error when rendered on the server so that they fallback to client-rendering, until the API can be updated to properly support server-side Suspense?

I'm not sure I understand what you're proposing. Is it to make `renderToString` throw on Suspense nodes? Or something else? Thanks!

text/0000-suspense-in-react-18.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Providing immediate feedback
+
+Doing something asynchronous without any feedback may also be confusing. For this reason, React will provide a `useTransition` Hook which returns a tuple like `[isPending, startTransition]`. You can then use `isPending` to reflect to the user that something is happening. The UI stays completely interactive — for example, the user is able to switch back to the `'photos'` tab if they'd like to.
+
+```js
+const [isPending, startTransition] = useTransition();
+
+function handleClick() {
+  setTab('comments');
```
thysultan, line 177: Give this:

>In the above example, when you click, the user will still see `Photos` for a bit, but the parent `div` will have 0.8 opacity, signaling a transition.

Shouldn't this be:

```js
  startTransition(() => {
    setTab('comments');
  });
```

text/0000-suspense-in-react-18.md
```diff
@@ -0,0 +1,285 @@
...
+
+Previously, React would do this in a sequence that goes like this:
+
+1. Place `Panel` content into the DOM, but with a "hole" instead of `Comments` content.
+1. Add `display: none` to the incomplete `Panel` content so that it does not appear visible.
+1. Add the `Spinner` content into the DOM.
+1. Fire the `Panel` effects because technically it has "mounted" (even though not fully).
+1. Wait for the `Comments` to be ready.
+1. Then, attempt rendering again.
+1. Remove the `Spinner` content form the DOM.
```
penspinner, line 107:
```suggestion
1. Remove the `Spinner` content from the DOM.
```

text/0000-suspense-in-react-18.md
gaearon: yes.

sebmarkbage:
@gaearon The case is not something that happens by just upgrading to newer React. It's when you first upgrade to newer React and then later start using existing Suspense-y APIs on the server and `<Suspense>` "just works". Except it doesn't just work because the result isn't serialized in the stream so it might change and cause hydration errors.

So there's no conditional Suspense.

gaearon:
>existing Suspense-y APIs on the server

Do you mean third-party libs that throw Promises today?

gaearon:
OK yeah got it. I thought this was about using `<Suspense>` for code splitting, but I see now the question was about using libraries that try to use Suspense for data fetching (which is not fully supported yet), and the fact that these libraries used to throw on the server, but now they may just subtly be broken. I agree it makes sense for them to throw explicitly until there is a way to transfer the data to the client.

text/0000-suspense-in-react-18.md
```diff
@@ -0,0 +1,287 @@
...
+
+We don't think this concern is relevant anymore for several reasons. `UNSAFE_componentWillMount` was marked as "unsafe" in 2018, and most popular open source libraries have long migrated away from it. This issue also only affects components "between" the `<Suspense>` node and the component that actually suspends. So it is very local in scope, and is easy to fix. Finally, Hooks have become a popular alternative to classes, and don't have the same issue. On the other hand, the current behavior [has been causing issues](https://github.com/facebook/react/issues/14536) for using popular component libraries with Suspense. This is why we think now is a good idea to make that change.
```
Friss, line 125: Is there a timeline for when these `UNSAFE_` methods would go away? This seems to be the first concrete reason to move away from these APIs.

jasonflorentino: FYI there’s a typo on line 123, after the link for [described]:
`ot` should be `to`
strategies ot migrate -> strategies to migrate
Love all this btw!

gaearon: For what it’s worth, componentWillMount potentially firing multiple times was _the_ original reason that it was renamed to UNSAFE in 2018. So I wouldn’t quite say that it’s “the first concrete reason” now. This was _the_ original reason for the rename four years ago. We didn’t anticipate that the whole journey to actually taking advantage of this would take so long, but this reason was exactly why we suggested to move away from it back then.

There is currently no plan to “remove” the UNSAFE methods. They’ll keep firing. It’s just that if you use Suspense, UNSAFE_componentWillMount may potentially execute multiple times without corresponding DidMount/WillUnmount if it’s in a component between Suspense and the part that suspends. Since many old componentWillMount code is not resilient to that, “UNSAFE” helps highlight that this code can be a problem.

voluntadpear:
Are the layout effects to re-run when content reappears effectively only those defined using `useLayoutEffect`? Although not recommended due to visual inconsistency, DOM mutations and reading using `useEffect` also happen (folks sometimes move to `useLayoutEffect` only after noticing blinks, layout shift, etc.) Are those effects also re-running?

text/0000-suspense-in-react-18.md
Friss: Gotcha, this to me just feels like the first step towards a world without these UNSAFE methods being available at all (with the language of its been 4 years and libraries have moved away etc) so I was curious if there was a timeline for that. 

For context at my company we have > 500 instances of just `UNSAFE_componentWillMount` and are currently undergoing a migration to move away from all UNSAFE methods. But overall its always easier to prioritize such work when they are being removed and a hard blocker for the upgrade vs potentially subtle bugs / opt into R17 behavior.

gaearon: I see. Our internal policy was to forbid them in any new code (or any code imported into new code) by using Strict Mode which complains about them. And then for old code (of which there is plenty), the policy is to exercise caution before using new features like Suspense. So it could still be added there but there are possible new memory leaks etc.

gaearon:
Re: https://github.com/reactjs/rfcs/pull/213#issuecomment-1077053718

>Are the layout effects to re-run when content reappears effectively only those defined using useLayoutEffect?

Yes, as well as ref callbacks.

>DOM mutations and reading using useEffect also happen (folks sometimes move to useLayoutEffect only after noticing blinks, layout shift, etc.) Are those effects also re-running?

No, `useEffect` does not re-run. If it's visually important then `useLayoutEffect` is the right tool to use, so this builds on the existing distinction (like you said, otherwise you might have shifts anyway). Not re-running `useEffect` is important because it leaves you a workaround for things that should _not_ reset. It also avoids performance issues with running too much setup/cleanup code when hiding/showing stuff.

bennettdams:
This sounds trivial, but: What are the current scenarios that allow a component to suspend in general?

The whole RFC gives explanations for what happens when a component suspends, but how do I, as a developer, enable a component to suspend? With that I mean `throw`ing a promise, `startTransition`, and so on.
Or, in other words: What are the "tools" we can use in our code so `<Suspense>` (and its fallback) is actually used?

I saw "_The exact protocol a component should use to tell React that it's suspended._" at the "_Unresolved questions_" - maybe I'm lost in translation, but I read this as "How a component lets React know while it is suspending" (=> "loading state"). Sorry for asking if it was not meant like this.

gaearon:
Re: https://github.com/reactjs/rfcs/pull/213#issuecomment-1077953067

>The whole RFC gives explanations for what happens when a component suspends, but how do I, as a developer, enable a component to suspend? 

Indeed, that's the "unresolved question" you're mentioning. The pragmatic answer is that today it works by throwing a Promise. **However** that's not an "official" or "final" API, and it's also not a complete description of the contract. You can't just throw arbitrary promises and expect it to work. Rather, it has to be backed by some cache outside of your components (rather than create during rendering), you need to throw *the same* Promise while it's pending (rather than creating a new one every time), you have to store its *result* outside of your components, you can't mutate the received data, and so on. These requirements are somewhat difficult to satisfy on your own, which is why we don't consider Suspense for data fetching officially supported yet. https://github.com/reactwg/react-18/discussions/25 is part of the work to make this simpler.

IDrissAitHafid:
I have a question, with this new behavior, if `Panel` is a big component that takes time to render it's better to put it outside the fallback like this:
```jsx
<Panel>
  <Suspense fallback={<Spinner />}>
    <Comments />
  </Suspense>
</Panel>
 ```
Because with the previous behavior `Panel` is already mounted, thus we don't have to wait for it to mount again once the `Comments` times out?

However, I am more in favor of the new approach because, in addition to its simplicity and all the reasons you have mentioned, I think it also simplifies react-devtools code by removing many of the Suspense edge cases, thanks to the fact that Suspense now behaves the same as regular fibers from the devtools perspective, right?

gaearon:
Re: https://github.com/reactjs/rfcs/pull/213#issuecomment-1078166879

>Because with the previous behavior Panel is already mounted, thus we don't have to wait for it to mount again once the Comments times out?

Re-rendering always starts from the `Suspense` node, so even in the old behavior, `Panel` would get re-rendered when `Comments` completes.

wintercounter:
I recently had proplems with Suspense during hydration. I had a text mismatch warning and then that immediatelly failed the Suspense root and switched to client side rendering. If it's "just" a warning, I wouldn't expect the whole thing to break. If it's an error, why is it only a warning? :) I tested with `suppressHydrationWarning` just out of curiousity, and my app didn't switch to client render anymore. Is this really the expected behavior?

gaearon:
Re: https://github.com/reactjs/rfcs/pull/213#issuecomment-1078667103

Yes, we’re treating hydration mismatches as errors now. They cause a fully client render at the granularity of the closest Suspense boundary above the mismatch. I will post more details about this later today. The correct fix is usually to _fix_ the mismatch but you’re right that for things like timestamps, the “suppress warning” prop now actually has production behavior too. This is unfortunate and we might rename it in a future release.

sebmarkbage:
We've effectively introduced a new form of Error that's kind of between a Warning and an Error... "Recoverable Error". It just means it can recover by client rendering so the user won't see it but it's an error so worth fixing.

gaearon:
More on that: https://github.com/reactjs/rfcs/pull/215

rickhanlonii (approved):

PR #215, opened 2022-03-25
RFC: Server Rendering Errors in React 18

gaearon:
In this RFC, we describe a mechanism in React 18 that allows apps to recover from server rendering errors automatically by wrapping parts of the tree into `<Suspense>` boundaries. Additionally, React does not attempt to "patch up" individual nodes on hydration mismatches and instead relies on the same mechanism to retry rendering a part of the tree on the client.

## [View the formatted RFC.](https://github.com/reactjs/rfcs/blob/server-errors/text/0000-server-errors-in-react-18.md)

### DIFF
--- a/text/0215-server-errors-in-react-18.md
+++ b/text/0215-server-errors-in-react-18.md
@@ -1,5 +1,5 @@
 - Start Date: 2022-03-25
-- RFC PR: (leave this empty)
+- RFC PR: https://github.com/reactjs/rfcs/pull/215
 - React Issue: (leave this empty)
 
 _Note: This RFC is closer to an "intent to ship" and is different than the
--- a/text/0000-server-errors-in-react-18.md
+++ b/text/0000-server-errors-in-react-18.md
@@ -1,0 +1,128 @@
+- Start Date: 2022-03-25
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+_Note: This RFC is closer to an "intent to ship" and is different than the
+process we typically do because it is the result of years of research into
+concurrency, Suspense, and server rendering. All of what is posted here was
+designed and discussed over the last year in the React 18 Working Group
+([available here](https://github.com/reactwg/react-18/discussions)) and iterated
+on in public experimental releases since 2018. **We'd like to get one final
+round of broad public feedback from the community before shipping in case there
+are new concerns that have not been discussed before.** You can consider the
+Working Group to be a part of the RFC, so please feel free to quote and discuss
+any content from it when commenting on the RFC here._
+
+# Summary
+
+This RFC describes a new recovery mechanism for server errors in React 18. It lets you recover from errors thrown on the server by adding `<Suspense>` boundaries.
+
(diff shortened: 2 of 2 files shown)


acdlite (approved):

somewhatabstract:
Is there any way to make the formatted RFC visible to everyone?

text/0215-server-errors-in-react-18.md
```diff
@@ -0,0 +1,128 @@
...
+</Suspense>
+```
+
+If `Comments` or something inside of `Comments` throws on the server, React will include the `<Spinner />` HTML output in the server rendered HTML along with a special marker. On the client, during hydration, React will automatically retry rendering `<Comments />`.
+
+Hydration mismatches will also use the same mechanism to retry a client client render.
+
+# Motivation
+
+There used to be no way to handle a rendering error on the server, so any component error was fatal and deopted the entire app to a full client render (or, depending on how your app is set up, would completely break it). On the client, errors are handled by error boundaries. But error boundaries don't work on the server because they rely on state. Although it's possible there will later be a separate error boundary API for the server, in the meantime, this change provides a natural way for the app to recover from error. Since server and client run the same code, it is possible that the client will be able to succeed (e.g. if the error was caused by some server-only problem).
```
tenorok, line 44: > On the client, errors are handled by error boundaries.

But [`componentDidCatch`](https://reactjs.org/docs/react-component.html#componentdidcatch) not catching synchronously errors, which can be throws on first render when hydration run.

For example:
```jsx
function MyComponent() {
    // Explicit example, it can be any error.
    throw new Error('Undefined is not a function);

    return <div>Hello world!</div>;
}
```

If we wraps `MyComponent` in `ErrorBoundary` the error above will crash application entirely.

Do you have any solution for this case or may be I do something wrong?

gaearon: Please file an issue with a reproducing case?

dante01yoon:
This might be outdated question, but I wonder we don't need using [lodable components](https://loadable-components.com/) anymore for using Suspense in SSR?

tranvansang:
There is a minor case that is only achievable by using loadable components  (disclose: I am the maintainer of [@react-loadable/revised](https://github.com/react-loadable/revised))

With loadable components: we know which CSS files are required and insert them to `<head>`.
Otherwise, we need to either load all CSS files or lazily load them all.

Because the effect is small, I drop react-loadable completely in my projects and replace with react native methods (lazy, suspense) and load all CSS files in every request regardless of whether they are required or not. Hope that this helps.

tranvansang:
> Isn't it css code splitting could be acquired without using lodable components?

yes, css code splitting could be acquired without using lodable components

> Do you mean importing css file directly in components, not making classes like normal css-in-js, tailwindcss?

That is the case I mentioned. If you use tailwindcss, it is irrelevant.
For css-in-js, I do not have much experiment with SSR + css-in-js.
Previously, when using loadable components, with CSS import, CSS files are included in chunks and I identified which CSS files need for a request.

dante01yoon:
I guess not only css files, but also all static files might be loaded

PR #219, opened 2022-04-23
RFC: injectToStream

brillout:
[View formatted RFC](https://github.com/brillout/rfcs/blob/main/text/0000-inject-to-stream.md).

### DIFF
--- a/text/0000-inject-to-stream.md
+++ b/text/0000-inject-to-stream.md
@@ -1,0 +1,140 @@
+- Start Date: 2022-04-24
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Today, library authors cannot inject chunks into the React SSR stream in a way that works across all SSR frameworks (Next.js, Hydrogen, vite-plugin-ssr, ...).
+
+The problem is that it is the SSR framework that integrates the React SSR stream. This means that there is no way for a library author to access the React SSR stream (without cooperation from each SSR framework).
+
+This RFC proposes a new React API `injectToStream(chunk: string)` (along with a new hook `useStream()`), so that library authors can use it to inject HTML chunks to the React SSR stream. The point here is that `injectToStream()` works accross all SSR frameworks.
+
+Note that the current situation is highly problematic. Without something like `injectToStream()`, library authors cannot leverage React SSR streaming: we cannot expect each library author to coordinate with each SSR framework.
+
+That's why I believe this topic to be of high importance and high urgency.
+
+# Basic example
+
+```jsx
+// New `useStream()` hook
+import { useStream } from 'react'
+
+function SomeComponent() {
+  const stream = useStream()
+  if (stream === null) {
+    // No stream available. (Client-side, or when there isn't any SSR stream at all.)
+    // ...
+  }
+  const { injectToStream } = stream
+  injectToStream('<script type="application/json">[{"some":"data"}]</script>')
+  // ...
+}
+```
+
+# Motivation
+
+I'm the author of an RPC tool [Telefunc](https://telefunc.com/) and I'm working on integrating it with the React SSR stream, so that the user can use Telefunc to fetch data:
+
+```jsx
+// TodoList.jsx
+// Environment: Browser & Node.js server (SSR)
+
+// Telefunc transforms `TodoList.telefunc.js` into a thin HTTP client.
+import { fetchTodoItems } from './TodoList.telefunc'
+import { useTelefunc } from 'telefunc'
+
+function TodoList() {
+  const todoItems = useTelefunc(async () => {
+    const todoItems = await fetchTodoItems()
+    return todoItems
+  })
+  return (
+    <ul>
+      { todoItems.map(todoItem =>
+        <li>{todoItem.text}</li>
+      )}
+    </ul>
+  )
+}
+```
+
+```jsx
+// TodoList.telefunc.js
+// Environment: Node.js server
+
+export async function fetchTodoItems() {
+  // This works because `fetchTodoItems()` is always run on the server-side. (Telefunc makes
+  // an HTTP request when `fetchTodoItems()` is called remotely from the client-side.)
+  const todoItems = await query("SELECT text FROM todo_items;")
+  return todoItems
+}
+```
+
+The current working prototype uses [react-streaming](https://github.com/brillout/react-streaming).
+
+For users that manually integrate React into their stacks (e.g. with [vite-plugin-ssr](https://vite-plugin-ssr.com/)), Telefunc can require the user to use `react-streaming`. So that Telefunc can access the SSR stream over `react-streaming`.
+
+But, with other SSR frameworks such as Next.js, it not the user but Next.js that integrates the React SSR stream. This means that there is no way for Telefunc to access the SSR stream.
+
+The same goes for all others React frameworks (Hydrogen, Remix, etc.).
+
+
+# Detailed design
+
+This RFC proposes a new API `injectToStream()` with two ways to access it:
+ 1. With `useStream()`:
+    ```jsx
+    import { useStream } from 'react'
+
+    function SomeComponent() {
+      const stream = useStream()
+      if (stream === null) {
+        // No stream available.
+      }
+      const { injectToStream } = stream
+    }
+    ```
+    Enabling all kinds of higher-level hooks, such as [react-streaming](https://github.com/brillout/react-streaming)'s `useSsrData()` and `useAsync()`, and Telefunc's `useTelefunc()`.
+ 1. With `renderToPipeableStream()` (or `renderToReadableStream()`):
+    ```jsx
+    import { renderToPipeableStream } from 'react-dom/server'
+    const { pipe, injectToStream } = renderToPipeableStream(<App />)
+    ```
+    Enabling SSR frameworks such as Next.js to inject scripts to the HTML.
+
+For more details see [`react-streaming`'s implementation of `injectToStream()`](https://github.com/brillout/react-streaming/blob/e90e29690d1029050ec06f923877fdaaa1209f5d/src/renderToStream.ts).
+
+# Drawbacks
+
+None AFAICT.
+
+# Alternatives
+
+> What other designs have been considered?
+
+Higher-level hooks such as [react-streaming](https://github.com/brillout/react-streaming)'s `useSsrData()` and `useAsync()`.
+
+But `injectToStream()` is lower-level and fundamentally more flexible. Libraries like `react-streaming` can leverage `injectToStream()` to implement and provide higher-level hooks. (This is actually already the case: `useSsrData()` and `useAsync()` are based on top of `injectToStream()`.)
+
+Library authors (Telefunc, React Query, CSS-in-JS tools, ...) can then use the higher-level hooks provided by tools like `react-streaming`.
+
+That said, I think it could make sense to make `useAsync()` a React built-in hook, but this would be another (low-priority & non-urgent) RFC.
+
+> What is the impact of not doing this?
+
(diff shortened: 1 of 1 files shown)


sebmarkbage:
I agree about the problem statement and urgency. I considered this and intentionally excluded it from the MVP but doesn't mean it can't be added.

The main thing that makes a hook like this less powerful is that I don't think features using it will be future compatible with new features that we have in mind. There are way more things we can do with SSR that we haven't discussed yet. We've also explicitly decided not to add something like this for Server Components.

Therefore our focus has been on adding higher level replacements for each use case.

I expect any solution using this to be temporary in the interim before we add new features that wouldn't work well with them anyway and until we add the higher level alternatives.

That's not to say that adding this in the interim wouldn't be valuable. That's why we have things like useEffect in the first place. Tbh, not adding this is more of a prioritization of work - which effectively shifts the burden to others to deal with the intermediate step. This is unfortunate but the reality.

Another proposal is to just have every SSR library add an API like this. It would require agreement and coordination but so would adding it to React. It just side-steps the React team as the bottleneck and maintainers.

To get a sense for other RFCs coming that you can expect will touch of this problem space:

- Inserting deeply nested `<head>` (not intended for data, styling or script injection).
- Managing preloading scripts and external CSS files.
- Inserting preloads/early loads for `React.lazy` scripts discovered during SSR.
- Inserting preloads of images discovered during SSR.
- Inserting `<link rel="stylesheet" />` tags in a Suspensey way both on the server and client.
- Inserting data that was fetched during SSR into the HTML for hydration.
- Special treatment of Server Components streams when they're the data.

These would also work from Server Components.

A notable missing one is CSS-in-JS that use dynamic `<style>` tag generation. I see a lot of issues with how that plays with some future features (the first one being Server Components). We haven't figured out an API that I think works well. We already added `useInsertionEffect` on the client but we don't have an equivalent on the server. We'd have to add something though even if it has downsides when used with other features. If it doesn't work well with other features anyway, then maybe it's fine that it's a low level API such as this one.

None of this is to say that we can't/shouldn't add this but this is just why we haven't done it yet. If we do, it's urgent since it's more as an interim feature.

brillout:
Thanks for the insightful answer, I appreciate it.

Ok, I see the reluctance because of potential future breaking changes, but the way I see it is that this RFC (or any alternative that enables library authors to inject to the stream) doesn't make the situation worse: React's 18 discussion board already recommends injecting tags to the stream such as style tags, data `<script type="application/json">` and JS `<script src="script.js">` tags. This means that library authors are already injecting to the stream, and this RFC doesn't introduce any potential future breaking changes that don't already exist today.


> Another proposal is to just have every SSR library add an API like this. Technically you don't need the React team to add this. It would require agreement and coordination but so would adding it to React. Each of them would just have to commit to an upgrade path. It just side-steps the React team as the bottleneck and maintainers.

The thing I'm worried about is the feasibility to align all SSR frameworks.

If all SSR frameworks (Next.js, Hydrogen, ...) were to use `react-streaming` (or an alternative) then that we solved the problem: library authors (Telefunc, GraphQL, ...) would then use `react-streaming`'s implementation of `injectToStream()`. But I can't really convince all SSR frameworks to use [`react-streaming`](https://github.com/brillout/react-streaming).

Instead of all SSR framework agreeing on using `react-streaming`, the more likely outcome is that each SSR frameworks ends up implementing their own `injectToStream()`. This means that library authors will have to use `n` `injectToStream()` variants where `n` is the number of frameworks.

We can expect SSR frameworks to spend time to implement their own `injectToStream()`, but I don't think we should expect library authors, which are much more resource limited, to need to use `n` different `injectToStream()` interfaces to be able to inject to the stream.


> Tbh, not adding this is more of a prioritization of work - which effectively shifts the burden to others to deal with the intermediate step. This is unfortunate but the reality.

I see. So I guess it boils down to how we can design the interm solution so that:
 1. It enables tools like `react-streaming` to seamlessly work accross all SSR frameworks.
 2. It's quick for React to implement.

Library authors (Telefunc, GraphQL, ...) can then use higher-level abstraction such as react-streaming to easily inject what they need to the stream, without worrying about framework integration.

This RFC is a solution for 1. but I can't say if there is a better solution for 2.

sebmarkbage:
@brillout If we didn't do this and picked one higher level use case to prioritize. If you had to pick one use case, which one would you pick?

brillout:
This one:

> * Inserting data that was fetched during SSR into the HTML for hydration.

By far. Simply because it's required to have suspended SSR boundaries.

Without passing SSR data to the client-side, we cannot achieve suspended SSR boundaries for data fetching. Which defeats the purpose of SSR streams.

These other use cases seem like low-priority nice-to-have optimizations:

> * Inserting deeply nested `<head>` (not intended for data, styling or script injection).
> * Managing preloading scripts and external CSS files.
> * Inserting preloads/early loads for `React.lazy` scripts discovered during SSR.
> * Inserting preloads of images discovered during SSR.
> * Inserting `<link rel="stylesheet" />` tags in a Suspensey way both on the server and client.

As for:

> * Special treatment of Server Components streams when they're the data.

I'm not all too familiar with Server Components, so I can't say. But since Server Components are not widely adopted yet, I'm thinking it's probably low priority as well.

<br/>

Also, I'm thinking maybe something like `react-streaming`'s [`useAsync()`](https://github.com/brillout/react-streaming#useasync) and [`useSsrData()`](https://github.com/brillout/react-streaming#usessrdata) could be high-level contracts between React and the users / library authors.

It seems to me that having `useAsync()` built into React to be quite useful for users and, AFAICT, `useAsync()` is a future-safe feature.

Same for `useSsrData()` but for library authors. (E.g. React Query would want to control the cache key.)

Let me know if you have questions, I'm happy to elaborate.

gaearon:
Some initial work on the stylesheet case is being done in https://github.com/facebook/react/pull/24886

brillout:
> work on the stylesheet case

👍 Also about the data fetching use case: https://github.com/acdlite/rfcs/blob/first-class-promises/text/0000-first-class-support-for-promises.md. I didn't read the whole RFC yet, but it seems like it completely replaces the need for [react-streaming](https://github.com/brillout/react-streaming). That's neat.

phryneas:
This is something we would really need in Apollo-Client to add good support for the new `renderToReadableStream`/`renderToPipeableStream` SSR (not RSC, that's a different set of problems!).

Right now, we are considering doing the following:

* Whenever request data is received on the server, we queue it using NextJS's `ServerInsertedHTMLContext` (an implementation detail of [`useServerInsertedHTML`](https://beta.nextjs.org/docs/styling/css-in-js#configuring-css-in-js-in-app)), which is meant to insert css during SSR.
* We replay that request data on the browser for cache entries to be added to Apollo Client

(Additionally, we also "snapshot" the return value of specific hooks to rehydrate them on the client side with the right value even if the client cache has already advanced to the future, but that's not relevant to this issue.)

Our current approach has a few issues.

* It's next.js specific, and we would like to support other SSR frameworks without creating an extra package for every framework (and possibly spending weeks on finding out how to support them best).
* If flushes data to the client every time a new suspense boundary is done. That can often be too late and result in race conditions - the client might have made a similar request after the server made it, and replaying the response the server forwarded to the browser with a long delay might set a the whole app "back in time". The ability to stream data "as soon as possible" and not "when the next suspense boundary is done" would be incredibly helpful here to minimize this risk.
* If we were to try and build a solution ourselves that used something like `AsyncLocalStorage`/`AsyncContext` to pass down the stream to inject into, we would run into a chicken-and-egg problem for calling `createReadableStream`, and even with `createPipeableStream` we would have to add a lot of logic to detect when React is currently streaming data, as not to inject data in-between React chunks. With time passing on, this would stay a "moving target" as React keeps adding more features - React has more access to the internal state, so that would seem to be the right place to implement this.

Is there any way for this to be reconsidered/reprioritized? 
If it's just about implementation work: I would be happy to help with that, but I would need some guidance on how to implement this best.

dac09:
Great write out and RFC! 

It does seem like an important omission from the React's streaming capabilities at the moment, and it seems to me (with my limited knowledge) like an antipattern for data-fetching library authors to have to fork special versions for each framework.

The workarounds sadly aren't documented anywhere that I can see! Is there a safe way to inject to the stream with an intermediate stream? It doesn't seem that way to me right now as there's no guarantee that each of Reacts chunks are "complete" . 

Right now in Redwood, we will have to use the "ServerHtmlContext" workaround used in react-streaming and NextJS, but an indication from the React team of the direction we would be going in (and some idea if this is being prioritised) would be great. :) 

FWIW - we are looking at this without using RSC capabilities stand point at first - just streaming+ssr, before we move into the RSC world.

jimrandomh:
In order to use Suspense with SSR and any sort of server-side data fetching, you need to somehow inject the result of fetches into the response stream. If you can't do this, the choices are no SSR or no data-fetching. The only thing the user-facing documentation says about this is "use a framework". As far as I can tell there is no framework-author-facing documentation about this anywhere.

When you use `renderToPipeableStream`, you get a stream into which chunks of HTML get written. You can mix this with other content (ie, the stuff you want to inject) by passing it through a `TransformStream` or extending `Writable`. These are standard JS APIs which operate on chunks (byte arrays). The problem is that you can't inject just anywhere; if you insert a chunk containing a `<script>` tag as soon as you know you need it, you're likely to be in the middle of some HTML at a spot where this isn't syntactically valid HTML because it's in the middle of some other tag. And if you inject into the stream too late, the component that wanted the data will have already been hydrated (with a mismatch).

So to find a safe insertion spot, you need to either convert the byte-stream back to UTF-8, parse the HTML, find a safe insertion point inside it, and re-serialize, or else find a secret React property that lets you identify safe insertion spots more cheaply.

The answer, it turns out, is that React will write chunks that do not correspond to safe insertion points, but will only yield to the event loop at safe insertion spots. So after seeing a call to `useServerInsertedHtml` or whatever you want to call it in your framework, you yield to the event loop exactly once with `setImmediate`, and that leaves you at a safe insertion point. (Reconciling this with the `TransformStream` API left as an exercise for the reader.)

This essential fact is not documented anywhere. When I went source-diving into apollo-client to figure out how they do it, I found that they... [reverse-engineered how nextjs does it](https://github.com/apollographql/apollo-client-integrations/issues/325#issuecomment-2205375796).

It looks to me like a dubious architectural decision was made here that makes it a lot more difficult to use React directly without a framework, and to make it more difficult to create new frameworks. And I have to say, the process of figuring this out, and finding this RFC, and seeing that the RFC was rejected by someone who now works for Vercel, has left a pretty bad taste in my mouth and leaves me worried about the long-term health of the React ecosystem.

phryneas:
> This essential fact is not documented anywhere. When I went source-diving into apollo-client to figure out how they do it, I found that they... [reverse-engineered how nextjs does it](https://github.com/apollographql/apollo-client-integrations/issues/325#issuecomment-2205375796).

Said Apollo Client maintainer here 👋 
At this point I've reverse engineered integrations into Next.js, React Router and TanStack Start. Having to keep up with the changes to internals of three frameworks which all have their own workarounds and hacks for this is ... not fun. So yeah, this general problem still exists and still is in need of a good solution.

> seeing that the RFC was rejected by someone who now works for Vercel, has left a pretty bad taste in my mouth and leaves me worried about the long-term health of the React ecosystem.

That part is just FUD and doesn't do anyone involved justice.
The React team is known for adopting only patterns that they are sure make sense in their long-term-plan for the library. I've discussed this a lot with them and I can vouch that they are sincerely trying to find a perfect solution here that they can maintain a long time and won't settle for less. 
I can't say that I'm perfectly happy about this, as perfect is sometimes the enemy of "anything at all" and it can mean that things sometimes don't move for a long time, but I can assure you that there is definitely no maliciousness here.

PR #221, opened 2022-06-06
RFC: Test selectors

bvaughn:
This proposal introduces a set of APIs to help with automated testing of React applications.


text/0000-test-selectors.md
```diff
@@ -76,7 +76,7 @@ While useful, most of these existing APIs generally have downsides:
 * They do not do enough to discourage brittle test selectors from being written (e.g. `"ProfilerHeader NavLink label a"`).
   * Test ids provide an alternative to this but they [have scaled very poorly in practice](https://gist.github.com/bvaughn/d3c8b8842faf2ac2439bb11773a19cec#gistcomment-3966895).
 
-A first-party React selector API could be adopted by these frameworks so they are no longer be renderer-specific, (although this might require imposing constraints on what the selector syntax supported). It could also make open source libraries like `[react-testing-library](https://github.com/testing-library/react-testing-library)` more powerful.
+A first-party React selector API could be adopted by these frameworks so they are no longer be renderer-specific, (although this might require imposing constraints on what the selector syntax supported). It could also make open source libraries like [`react-testing-library`](https://github.com/testing-library/react-testing-library) more powerful.
```
bvaughn, line 79: Agh, thanks.

text/0000-test-selectors.md
```diff
@@ -0,0 +1,816 @@
...
+
+## `createTestNameSelector()`
+
+Matches host elements with a specific test name data attribute (e.g. `data-testname="LoginFormButton"`).
+
+```js
+createTestNameSelector('LoginFormButton')
+```
+
+## `createHasPseudoClassSelector()`
```
motiz88, line 163: Bikeshedding, but this one is the only selector whose name doesn't make intuitive sense to me. Maybe something like `createHasContentsSelector` could work?

motiz88: Actually, I meant to propose `createContainsSelector`.

bvaughn: Agreed that the naming of this one is a mouthful. The concept (and so the name) is inspired by [CSS pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes). I don't feel strongly about this name being _the one we must have_ though. The important thing is that the name makes it clear that this one works differently than the others: It doesn't select the element(s) at the end of its nested selectors; it just refines the criteria for where it's at in the higher level selection tree.

bvaughn: I am not strongly opposed to the name `createContainsSelector` exactly, but I do worry it's potentially confusable with `createTextSelector` (since people might think of elements as _containing_ text).

motiz88: Are the semantics close enough that we can make them _both_ overloads of `createContainsSelector`? i.e. pass a string to Contains -> get the Text behaviour. (Then it's a separate design choice whether to keep the Text selector around separately for convenience/readability.)

bvaughn: I think that would be bad because the behavior of those selectors is different:
* Text selector works like a normal selector, in that if you put it at the end of a selector chain– it selects the lowest level elements that match it.
* This selector we're talking about acts to refine the sequence it's a part of, but doesn't select anything inside of it. (It just looks ahead and excludes things from the chain it's in– kind of like a regex lookahead.)

The text selector name also has some precedent with other testing libraries.

bvaughn: Maybe `createLookaheadSelector` 🤪

motiz88: I think the closest CSS analogue is `:has()`, right? So maybe `createHasSelector`, except that it scans as "has selector" which is awkward. As a compromise from there, `createHasElementSelector`?

motiz88: Last one of this batch of ideas: `createHasMatchSelector`

bvaughn: I jokingly mentioned the lookahead name, but there is a pretty strong similarity. Of the names we've discussed, I think maybe that one gives the strongest indicator of what it does / how it works relative to some other established concept.

motiz88: My 2 cents on "lookahead", while we're here, is that it may be an established term, but still seems niche / context-specific. It's also more imperative than declarative, if that makes sense: it tells the selector matcher what to do instead of what element to look for.

motiz88: In other words, "Lookahead" would be an improvement over "HasPseudoClass" for me as a reader but I'd still worry about it being too opaque to a general audience.

PR #222, opened 2022-08-01
Grammatical corrections

Akhigbe-E:
Corrections for improved comprehension


rickhanlonii:
Thanks for reviewing this doc! Unfortunately we don't tend to merge these types of changes because many times it's subjective and creates a lot of noise for maintainers.

PR #224, opened 2022-08-25
RFC: useGlobalState

mushan0x0:
This is the global state implemented using batch updates. The usage is similar to useState, simpler than context, and the usage scenarios are different.

[Here is my implementation and usage example.](https://codesandbox.io/s/dumi-demo-forked-73u0pv?file=/useGlobalState.tsx)

### DIFF
--- a/text/0000-use-global-state.md
+++ b/text/0000-use-global-state.md
@@ -1,0 +1,44 @@
+- Start Date: 2022-08-25
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+Implement an easy-to-use global state hook with unique keys and batch updates. No need for createContext, provider, useContext, it is as simple as using useState, the difference is that there is a unique key.
+
+# Basic example
+
+If the proposal involves a new or changed API, include a basic code example.
+Omit this section if it's not applicable.
+
+# Motivation
+
+Context is easy to be abused. In actual development, developers may put a lot of things in it, which makes management difficult and performance difficult to optimize. The design of useGlobalState is consistent with useState, which can avoid this problem.
+
+The current react needs to use the context api to achieve state sharing. This may not be suitable and easy in terms of semantics and usage.
+
+It is applied to pages or components that require the same state, and does not require context and provider reuse scenarios.
+
+It can be implemented using unique keys and batch updates, which is much simpler, as easy as using useState, the only difference is an additional key parameter.
+
+# Detailed design
+
+![architecture diagram](https://bojuematerial-prudcut-public.oss-cn-guangzhou.aliyuncs.com/fdsafdsfdggfhfhgfgds.png)
+
+# Drawbacks
+  
+- keys in use may conflict.
+- when several places are used at the same time, the setting of the initial value may be confusing to the user.
+- the user does not know the difference with the context.
+
+# Alternatives
+
+The saved state needs to be cleaned up immediately, otherwise it will cause a memory leak. Consider adding a throw error when the key is duplicated.
+
+# Adoption strategy
+
+This is a new api, the usage method is similar to useState, it does not require too much learning cost, and there is no historical burden.
+
+# How we teach this
+
+Tell the user that useGlobal has no provider, cannot be reused, and how to use it differently from context.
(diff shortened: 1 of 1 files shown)


Zachiah:
Why not have the keys be `Symbol`s? That would mean you basically couldn't conflict. The downside is that you would have to import the `Symbol` from another module or make a wrapper hook and import it, however, I think this is really a plus not a downside as it makes it so much clearer where the state sharing is

iMrDJAi:
I had a similar but a slightly different approach. Check my implementation of [useSync()](https://github.com/iMrDJAi/useSync).

Do you think that it is worth creating an RFC for?

iMrDJAi:
@wmzy You'll still have to import the hook into different components in order to use it, which is not ideal. A subscription based one with a unique identifier (either a string or a symbol) will be way handier and will cause less headache when trying to remember which hook to import and where to import it from.

bennobuilder:
Libraries like [`feature-state`](https://github.com/inbeta-group/monorepo/blob/develop/packages/feature-state/README.md), [`jotai`](https://github.com/pmndrs/jotai), and [`nanostores`](https://github.com/nanostores/nanostores) already provide implementations for creating global states.

How would `useGlobalState()` differ?

### Example using [`feature-state`](https://github.com/inbeta-group/monorepo/blob/develop/packages/feature-state/README.md) (< 1KB minified)

`store/tasks.ts`
```ts
import { createState } from 'feature-state';

export const $tasks = createState<Task[]>([]);

export function addTask(task: Task) {
    $tasks.set([...$tasks.get(), task]);
}
```

`components/Tasks.tsx`
```tsx
import { useGlobalState } from 'feature-state-react';
import { $tasks } from '../store/tasks';

export const Tasks = () => {
    const tasks = useGlobalState($tasks);

    return (
        <ul>
            {tasks.map(task => <li key={task.id}>{task.title}</li>)}
        </ul>
    );
}
```

iMrDJAi:
@bennoinbeta And why going for third party solutions if React already had this built-in? Given how crucial this is and how simple it is to implement, it makes sense to have it included in core.

sarath263:
I can suggest another way, like a selector ( a bit looks like useSelector in redux)

`const [value, setGlobalState] = useGlobalState(s => s.name);`

Returns: `[value, setGlobalState]`

- `value`: The current value of the selected slice. `undefined` if the slice doesn't exist.
- `setGlobalState`: Function to update the global state by merging the provided object.

**Note**: `setGlobalState` merges the provided object into the global state. It can update any part of the state, not just the selected slice.

I have a full implementation of this. Since most people agreed that we need a `useGlobalState` hook for sure, I believe we can immediately decide on an implementation plan and do it ASAP.

vezaynk:
This seems unnecessary given that you can simply 

1. write your own state management abstraction
2. put an instance of into a context
3. access via context
4. query it exactly as you do now


Indeed, Zustand+Context achives exactly that, so you don't even need to write it yourself: https://tkdodo.eu/blog/zustand-and-react-context

sarath263:
Isn't it cool to use a global state just like useState 

const [name, setGlobalState] = useGlobalState(s => s.name);

Set name to global state:
 setGlobalState({name:"Kevin"})

Simple and humble.. right?

The problem with context is that it is not that optimized on render control. Update one value in context, all the components which uses any other part of context too will re render.

sarath263:
Hi React team,
Please check this implementation. 

```
import { useNativeState } from 'native-state-react';

function App() {
  const [name, setName] = useNativeState('s.name', 'Andrew');

  const updateName = () => {
    setName("George");
  };

  return (
    <div>
      <p>Name: {name}</p>
      <button onClick={updateName}>Update Name</button>
    </div>
  );
}
```
For [full implementation](https://github.com/sarath263/native-state/blob/main/src/nState.js), [example ](https://github.com/sarath263/native-state/tree/main/example) and [demo](https://codesandbox.io/p/sandbox/global-state-native-state-react-c4j9yc) . 
[NPM Library](https://www.npmjs.com/package/native-state-react).

This uses just the [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) from react , so the concurrent rendering is solved.
@gaearon @acdlite  , kindly take a look.

PR #226, opened 2022-09-11
RFC: experimental_use(customHook)

dai-shi:
[View formatted RFC](https://github.com/dai-shi/rfcs/blob/patch-2/text/0000-experimental-use-custom-hook.md)

### DIFF
--- a/text/0000-experimental-use-custom-hook.md
+++ b/text/0000-experimental-use-custom-hook.md
@@ -45,6 +45,8 @@ hooks implementation is just arrays. https://www.swyx.io/hooks/#not-magic-just-a
 So, implementation-wise, it will be `Map<customHook, array>` instead of just an array.
 The referential equality of `customHook` is important.
 
+(A little more precicely, `experimental_use` can be nested, so `Map<customHook, array>` is a recursive structure forming a tree.)
+
 A question is if this is an acceptable concept, but a similar approach is once proposed.
 https://github.com/reactwg/react-18/discussions/25
 > The "create" function acts as the type/key/identity for the cache.
--- a/text/0000-experimental-use-custom-hook.md
+++ b/text/0000-experimental-use-custom-hook.md
@@ -1,0 +1,76 @@
+- Start Date: 2022-09-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+In the React repository, there are a few PRs to add `experimental_use`.
+- [experimental_use(promise)](https://github.com/facebook/react/pull/25084)
+- [experimental_use(context)](https://github.com/facebook/react/pull/25202)
+
+The `experimental_use` hook is a special hook that can be called conditionally.
+https://twitter.com/acdlite/status/1558186162518478850
+
+As the name implies, it can accept some other values.
+This PR proposes `experimental_use(customHook)`.
+
+# Basic example
+
+```jsx
+import { experimental_use, useState } from 'react';
+
+const useCountState = () => useState(0);
+
+const Component = ({ mode }) => {
+  if (mode === 'simple') {
+    return <div>Just nothing</div>;
+  }
+  const [count, setCount] = experimental_use(useCountState); // This doesn't violate the React hooks rule.
+  return <div>{count}<button onClick={() => setCount(c => c + 1)}>+1</button></count>
+};
+```
+
+# Motivation
+
+There is a rule with React hooks, which is hooks can't be called conditionally.
+Some considered it kind of a bad design and other frameworks chose different design decision.
+It would be so nice, if we can provide a solution to it.
+`experimental_use` seems like a nice fit.
+
+# Detailed design
+
+My understanding of React internals is very limited, but at high level
+hooks implementation is just arrays. https://www.swyx.io/hooks/#not-magic-just-arrays
+
+So, implementation-wise, it will be `Map<customHook, array>` instead of just an array.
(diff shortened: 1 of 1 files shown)


MrFlashAccount:
Shall we change the current naming conventions and consider removing the ‘use’ prefix?

‘experimental_use(useState(...)) -> experimental_use(state(...))’

dai-shi:
As for naming convention, this would be possible.
```js
const myState = () => useState();

  // inside component/hook
  experimental_use(myState);
```

But, I'm not sure if it's good. We need to consider about eslint rule too to detect bad practice.

> `experimental_use(useState(...))`

Note that it's not possible like that. Needs a custom hook defined outside.

MrFlashAccount:
Oh, I see. Wrapping into a function outside of a component is essential for referential stability.

satya164:
I think using the function reference for identification could be problematic:

1. I regularly see people create components during render and run into issues, I'd assume people would do the same thing with this and get confused why it doesn't work as they expect
2. React can't warn about this in runtime, coz the reference could just change due to conditionals, and not because it's a mistake - which makes it harder to catch errors. It can be linted against but unfortunately, many beginners don't use the lint rule so it'd be better to avoid relying on it for new features
3. This assumes custom hooks would need to be created for each usage, however since existing custom hooks would work with this as well, it's possible that in bigger components people may use the same hook multiple times causing unexpected results

dai-shi:
I agree it's hard to do runtime checks.

What would happen if we do `use(Promise.resolve(1))` ? Wouldn't it be the same problem?

> people may use the same hook multiple times causing unexpected results

I'm not sure if I understand how using it multiple times causes unexpected results.

satya164:
> What would happen if we do `use(Promise.resolve(1))` ? Wouldn't it be the same problem?

Yes, I don't necessarily like that either. But there can be the argument that a component suspending is very visible. However in this case it's just silent and needs more testing to find out that something is wrong.

> I'm not sure if I understand how using it multiple times causes unexpected results.

If the hook is identified by reference, and someone uses the same hook multiple times (like they would use `useState`), then the naive expectation is to think they are all different, but they'd actually be the same one. It's a user mistake but an easy mistake to make with bigger components.

dai-shi:
> the naive expectation is to think they are all different, but they'd actually be the same one.

I see. It's the case if we use the same custom hook more than once at the same level. I'm not sure if the naive expectation is they are different (some may assume they are the same). But, I understand the pitfall. We could limit the usage just once at the same level, which can be detected at runtime.

dai-shi:
> React can't warn about this in runtime, coz the reference could just change due to conditionals

Hm, because render has to be idempotent, we can detect such a change with double render in StrictMode, I think.

sebmarkbage:
I have some ideas for using this syntax in a similar way but there’s nothing new about the syntax here that unlocks any capabilities.

The reason context and promises can be read unconditionally is that they don’t represent state. Nothing has changed about the stateful part so there’s nothing new here.

There are other techniques we can do but it doesn’t have anything to do with this syntax and as far as I can tell there’s no new mechanism in this RFC that wasn’t already considered in the original Hooks RFC.

dai-shi:
> The reason context and promises can be read unconditionally is that they don’t represent state.

Implementation-wise, I think I understand that.
This RFC's proposal is to add such a capability (= reading state unconditionally), from the user perspective.

> There are other techniques we can do but it doesn’t have anything to do with this syntax

I agree there could be some other syntax to allow reading state unconditionally.
The motivation behind this syntax is, it's _somewhat_ close to the mental model of `experimental_use(promise/context)`.

> there’s no new mechanism in this RFC that wasn’t already considered in the original Hooks RFC.

Maybe so. But, there wasn't any discussion of the special syntax of `experimental_use()` that can be called unconditionally, no?
So, this RFC is born because `experimental_use()` is being added.
If it were `useThenable()` and `useContext()` that can be called unconditionally, this RFC wouldn't have been considered.

sebmarkbage:
It might be worth waiting on a `use()` RFC to understand what it actually does first.

Technically `useContext` could already be called unconditionally and `readContext` is just the same thing without the lint warning. It's just that if you wrapped it in a custom Hook it didn't play nicely because other Hooks can't be called unconditionally so by convention it can't be called unconditionally to play nicely with the rest. The introduction of `use()` doesn't change this for the other Hooks nor custom Hooks.

We've shown in other experiments (since Prepack days) that we can support conditional Hooks and in keyed loops using a compiler.

dai-shi:
Yeah, this RFC was too rush.

I didn't mean to base on `use()` implementation. So, this mixing two things is probably too confusing implementation-wise and mental-model-wise.

Basically, this is to revisit the "rejected" idea in the original Hooks RFC.
We can close this RFC if there are no further discussions.

> We've shown in other experiments (since Prepack days) that we can support conditional Hooks and in keyed loops using a compiler.

The difference between that ☝️ and this RFC would be a) using function reference as a key, and b) not a compiler but runtime. `use()` syntax isn't essential (although, it's a motivation behind this proposal.)

---

On the other hand, I'd still like to explore to open up some new possibilities with `use()`, like `use(usable(() => use(context)...))`, maybe after we have the `use()` RFC.

sebmarkbage:
Yea using function identity or even location of the source call function was evaluated early on but that suffers from the issues pointed out above. It would share state with other usage of that same function.

It is possible if the thing created has identity before it’s passed to use:

```
const [store] = useState(createStore());
if (c)
  use(store);
```

I’ll close this for now.

PR #227, opened 2022-10-11
RFC: React Server Module Conventions v2

sebmarkbage:
This is a new take on [the original v1 of this proposal](https://github.com/reactjs/rfcs/pull/189). See the discussion there for deeper context on the reason for the changes made.

Please watch [the React Server Components intro talk](https://reactjs.org/server-components/) and high level [React Server Components RFC](https://github.com/reactjs/rfcs/pull/188) for some background context on this RFC.

### DIFF
--- a/text/0000-server-module-conventions.md
+++ b/text/0000-server-module-conventions.md
@@ -29,7 +29,7 @@ We also propose using package.json conditional exports to allow an import to for
 import Component from "./Component.js";
 function Parent() {
   // Component is of type Reference<T> where T is the type
-  // of the default export of Component.client.js
+  // of the default export of Component.js
   // Because of this React knows that it can't render it on
   // the server and instead will leave it as a placeholder
   // to later be rendered on the client.
@@ -43,11 +43,12 @@ function Parent() {
 
 function Child() {
   // Component is of type Reference<T> where T is the type
-  // of the default export of Component.client.js
+  // of the default export of Component.js
   // Because of this React knows that it can't render it on
   // the server and instead will leave it as a placeholder
   // to later be rendered on the client.
-  return <Component />;
+  let [state, setState] = useState(false);
+  return <Component onClick={() => setState(true)} value={state} />;
 }
 ```
 
--- a/text/0000-server-module-conventions.md
+++ b/text/0000-server-module-conventions.md
@@ -26,7 +26,7 @@ We also propose using package.json conditional exports to allow an import to for
 
 ```js
 // Parent.js
-import Component from "./Component.client.js";
+import Component from "./Component.js";
 function Parent() {
   // Component is of type Reference<T> where T is the type
   // of the default export of Component.client.js
--- a/text/0000-server-module-conventions.md
+++ b/text/0000-server-module-conventions.md
@@ -1,0 +1,178 @@
+- Start Date: 2022-10-11
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This is a new take on [the original v1 of this proposal](https://github.com/reactjs/rfcs/pull/189). See the discussion there for deeper context of the reason for the changes made.
+
+Please watch [the React Server Components intro talk](https://reactjs.org/server-components/) and high level [React Server Components RFC](https://github.com/reactjs/rfcs/pull/188) for some background context on this RFC.
+
+[React Server Components](https://github.com/reactjs/rfcs/pull/188) doesn't have much of an API other than integration into meta-frameworks. Instead, we propose that the main way you make choices about what runs on the server and client is defined by its file extension.
+
+We also propose using package.json conditional exports to allow an import to fork its implementation between server and client usage.
+
+# Changes Since v1
+
+- No more file extension conventions.
+- A `"use client"` directive at the top of a file defines that it's a boundary between server and client.
+- Server and Client Components no longer have to explicitly state so. Everything is implicitly Shared until proven otherwise, and then the build errors.
+- `import "server-only"` and `import "client-only"` can be used to poision a module so it can only be imported in that environment.
+- TypeScript Convention for enforcing serializable boundaries.
+
+# Basic example
+
+## `"use client"` Directive
+
+```js
+// Parent.js
+import Component from "./Component.client.js";
+function Parent() {
+  // Component is of type Reference<T> where T is the type
+  // of the default export of Component.client.js
+  // Because of this React knows that it can't render it on
+  // the server and instead will leave it as a placeholder
+  // to later be rendered on the client.
+  return <Component />;
+}
+```
+
+```js
+// Component.js
+"use client";
+
+function Child() {
+  // Component is of type Reference<T> where T is the type
+  // of the default export of Component.client.js
+  // Because of this React knows that it can't render it on
+  // the server and instead will leave it as a placeholder
+  // to later be rendered on the client.
+  return <Component />;
+}
+```
+
+When a Component with a `"use client"` directive (similar to `"use strict"`) is imported in a "React Server" environment its exports gets replaced with a special "Reference" object. This object can't be accessed directly in that file but it can be passed into React as if it was a plain component.
+
+React, together with the bundler, knows how to send this reference to the client. On the client it's rendered as a Client component. The real file never actually gets imported on the server.
+
+## `"react-server"` conditional exports
+
+```js
+// package.json
+{
+  ...,
+  "exports": {
+    ".": {
+      "react-server": {
+        "browser": "./debugger-polyfill.server.js",
+        "node": "./native-impl.server.js"
+      },
+      "default": {
+        "browser": "./browser-impl.client.js",
+        "node": "./node-polyfill.client.js"
+      }
+    }
+  }
+}
+```
+
+A package can choose to fork its exports depending the environment it executes in across two dimensions.
+
+The `"react-server"` condition applies only React Server Component environments. However, these can execute either in a `"browser"` during development for debugging purposes, or it can execute in a `"node"` environment in production.
+
+React Client Component environments are the default and they typically execute in a `"browser"` environment but they can also execute in a `"node"` environment for SSR purposes.
+
+Note: This condition can only be used for conditional `imports` within the project itself.
+
+Possible addition: A possible addition to this can be to also support the `"server-only"` condition as a less React specific alternative that could be adopted by other frameworks that also choose this strategy.
+
+# Poisoned Imports
+
+The `"react-server"` condition can be used to generate poisoned imports that can only be used by in one environment or another.
+
+I've already published these packages:
+
+`server-only/package.json`
+```
+"exports": {
+  ".": {
+    "react-server": "./empty.js",
+    "default": "./index.js"
+  }
+}
+```
+
+In this case `index.js` throws an error, so if you import this from a Client Component - even in SSR - you get an error.
+
+`client-only/package.json`
+```
+"exports": {
+  ".": {
+    "react-server": "./error.js",
+    "default": "./index.js"
+  }
+}
+```
+
+In this case `error.js` throws an error, so if you import this from a Server Component - you get an error.
+
+This lets you mark certain packages as only working in one environment or another using:
+
+```
+import "server-only";
+```
+or
+```
+import "client-only";
+```
+
+This can help enforce things statically with early error messages before you do a mistake. Especially since these conventions has such a strong uncanny valley. It can also enforce security or performance by ensuring that certain keys are only available to the server or that data fetching only happens from Server Components.
+
+The nice thing about this convention is that it works outside of Server Components compatible environments by ensuring that `server-only` throws in those. It also lets you configure your own custom error messages for libraries since it's just a convention.
+
+# TypeScript Convention
+
+Types between the Server and Client boundary mostly just work because they behave the same as if there was no boundaray. One limitation is that if those types includes a non-serializable type - according to the React Server Components specification, you get an error.
+
+The correct way to model this is to model Client References for exports at the boundaries and then have the callsite of JSX accept a Client Reference as a component and enforce the serialization at the callsite. This ensures that you can still pass client references from server to client in other forms than JSX. In fact, this is how Flow already implements it and it could be implemented in TypeScript too. However, this requires a bit more involved type checking from the TypeScript side.
+
+It's also not necessarily the best enforcement because it doesn't prevent a component from adding a non-serializable prop as inputs unless it can also type check against all possible usages. Which might not be best for error messages and not be viable if you're publishing a shared component that's used in other repos.
+
+Therefore, we propose another convention to files with a `"use client"` directive:
+
+The conventions apply if a module exports a function that is identified as a React Component, according to some definition specified by the [React Compiler](https://reactjs.org/blog/2022/06/15/react-labs-what-we-have-been-working-on-june-2022.html#react-compiler) and the [Rules of Hooks Lint rule](https://reactjs.org/docs/hooks-rules.html). Basically if it contains a capital name and containers JSX or Hooks.
+
+In that case, such function's arguments must all be serializable according to the Serialization rules of [React Server Components](https://github.com/josephsavona/rfcs/blob/server-components/text/0000-server-components.md) (basically JSON + Cycles, Binary Data, Promises, JSX etc).
+
+This can be enforced by a framework using code gen, a TypeScript plugin or just built-in to TypeScript. This doesn't need any additional information from the graph to be enforced.
+
+The implications of this is also that you shouldn't add the `"use client"` directive to all Client Components. Only the ones intended to be used directly by the Server. Because indirect ones are still allowed to have non-serializable props.
+
+# Drawbacks
+
+The main drawback of this new approach is that we further the uncanny valley. It's hard to tell when you jump into any given file whether that is going to get used as a Shared, Client and Shared component. That's the big downside of the whole Server Components design that we're taking a bet that it's worth learning this one thing to unlock its possibilities. To help combat this, we'll also propose that, as a convention Server Components are mostly written as "Async Functions" form instead of "Hooks Functions". More on that soon. While Shared and Client Components would be written as "Hooks Functions". This distinction is not enforced though.
+
+Additionally, since it's not enforced, it's difficult for tooling to tell if a file is a Shared, Client or Server Component. The only thing where you can tell that for sure is at the boundaries that are annotated. Otherwise you need to inspect the dependency graph. A lot of early errors might only be detected by a bundler that has the graph available to it.
+
+It also places constraints on the bundler. To find all Client Components to put in the client bundle, you now need to do a partial pass to parse them. It's not enough to just bundle all `.client.js` files. However, since in practice, most bundling has to follow the dependency graph to tree shake files that are not used you probably have to do that anyway.
+
+Because of the `"react-server"` export condition, the correct way to implement a module graph requires duplicating module resolution on the server. Not necessarily duplicating the bytes in each module but the resolution has to happen twice. If module A imports module B which is forked based on `"react-server"`, then there has to be two copies of module A. One for SSR and one for RSC. Since `"react"` itself is forked on `"react-server"` it means that any file that imports `"react"` needs to be initialized twice too. Naively you end up needing two separate bundles for SSR and for RSC. This can be good for the client navigation case since it's only have to load the RSC bundle then but worse for the SSR case. This is probably the most controversial part of the design.
+
+The benefit is that we can now statically and early error when you try to import something itself imports `useState` or `useEffect` and give better early error messages for easy mistakes.
+
+Another benefit is that this kind of fork is needed for the [React Compiler](https://reactjs.org/blog/2022/06/15/react-labs-what-we-have-been-working-on-june-2022.html#react-compiler) to be able to optimize SSR separately from RSC. E.g. it doesn't make sense to lower things to "concatenating pieces of HTML" for generating the RSC payload. So if you want to take advantage of this, you'll need something like it anyway.
+
+Right now dereferencing a Client Reference wouldn't be a TypeScript error.
+
+```
+// server-file.tsx;
+import {Foo} from 'client-file.tsx';
+
+Foo.hello();
+```
+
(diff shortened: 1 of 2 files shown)


text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+This is a new take on [the original v1 of this proposal](https://github.com/reactjs/rfcs/pull/189). See the discussion there for deeper context of the reason for the changes made.
+
+Please watch [the React Server Components intro talk](https://reactjs.org/server-components/) and high level [React Server Components RFC](https://github.com/reactjs/rfcs/pull/188) for some background context on this RFC.
+
+[React Server Components](https://github.com/reactjs/rfcs/pull/188) doesn't have much of an API other than integration into meta-frameworks. Instead, we propose that the main way you make choices about what runs on the server and client is defined by its file extension.
```
hamlim, line 11:
```suggestion
[React Server Components](https://github.com/reactjs/rfcs/pull/188) doesn't have much of an API other than integration into meta-frameworks. Instead, we propose that the main way you make choices about what runs on the server and client is defined by a directive.
```

text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+- `import "server-only"` and `import "client-only"` can be used to poision a module so it can only be imported in that environment.
+- TypeScript Convention for enforcing serializable boundaries.
+
+# Basic example
+
+## `"use client"` Directive
+
+```js
+// Parent.js
+import Component from "./Component.client.js";
```
hamlim, line 29:
```suggestion
import Component from "./Component.js";
```

text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+
+It's also not necessarily the best enforcement because it doesn't prevent a component from adding a non-serializable prop as inputs unless it can also type check against all possible usages. Which might not be best for error messages and not be viable if you're publishing a shared component that's used in other repos.
+
+Therefore, we propose another convention to files with a `"use client"` directive:
+
+The conventions apply if a module exports a function that is identified as a React Component, according to some definition specified by the [React Compiler](https://reactjs.org/blog/2022/06/15/react-labs-what-we-have-been-working-on-june-2022.html#react-compiler) and the [Rules of Hooks Lint rule](https://reactjs.org/docs/hooks-rules.html). Basically if it contains a capital name and containers JSX or Hooks.
```
hamlim, line 143:
```suggestion
The conventions apply if a module exports a function that is identified as a React Component, according to some definition specified by the [React Compiler](https://reactjs.org/blog/2022/06/15/react-labs-what-we-have-been-working-on-june-2022.html#react-compiler) and the [Rules of Hooks Lint rule](https://reactjs.org/docs/hooks-rules.html). Basically if it contains a capital name and contains JSX or Hooks.
```

text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+
+This can be enforced by a framework using code gen, a TypeScript plugin or just built-in to TypeScript. This doesn't need any additional information from the graph to be enforced.
+
+The implications of this is also that you shouldn't add the `"use client"` directive to all Client Components. Only the ones intended to be used directly by the Server. Because indirect ones are still allowed to have non-serializable props.
+
+# Drawbacks
+
+The main drawback of this new approach is that we further the uncanny valley. It's hard to tell when you jump into any given file whether that is going to get used as a Shared, Client and Shared component. That's the big downside of the whole Server Components design that we're taking a bet that it's worth learning this one thing to unlock its possibilities. To help combat this, we'll also propose that, as a convention Server Components are mostly written as "Async Functions" form instead of "Hooks Functions". More on that soon. While Shared and Client Components would be written as "Hooks Functions". This distinction is not enforced though.
```
hamlim, line 153:
```suggestion
The main drawback of this new approach is that we further the uncanny valley. It's hard to tell when you jump into any given file whether that is going to get used as a Shared, Client and Server component. That's the big downside of the whole Server Components design that we're taking a bet that it's worth learning this one thing to unlock its possibilities. To help combat this, we'll also propose that, as a convention Server Components are mostly written as "Async Functions" form instead of "Hooks Functions". More on that soon. While Shared and Client Components would be written as "Hooks Functions". This distinction is not enforced though.
```

text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+// Component.js
+"use client";
+
+function Child() {
+  // Component is of type Reference<T> where T is the type
+  // of the default export of Component.client.js
+  // Because of this React knows that it can't render it on
+  // the server and instead will leave it as a placeholder
+  // to later be rendered on the client.
+  return <Component />;
```
hamlim, line 50: Is it worth updating this snippet to `useState` or `useEffect` to better communicate why it's a Client component?

hamlim (commented):
Noticed a few small typos/carryovers from the previous iteration of the RFC

text/0000-server-module-conventions.md
sebmarkbage: Yea, that makes sense.

sebmarkbage:
Thank you! ☺️

text/0000-server-module-conventions.md
sebmarkbage: Done.

redbar0n:
> The implications of this is also that you shouldn't add the "use client" directive to all Client Components. Only the ones intended to be **used** directly by the **Server**.

This is confusing. Isn’t it effectively a «use server» directive?

Rather than telling the server to use a client component, the perspective would be to denote a client component for use on the server. The result is the same. But when looking at a client component it makes sense that it is the point of focus (not the server).

redbar0n:
ah, I see that my confusion was merely woth the wording of the aforementioned paragraph. Since:

> The real file never actually gets imported on the server.

So the «use client» directive is only for the server for the purpose of passing the component reference to the client where the conponent is executed. So «use client» kinda makes sense then.

kylemh:
> Interestingly, none of these use file extensions to separate the environments. Why is that? It would be equally useful to determine whether some code is intended to be used in Node.js, browsers, react-native or either by purely the file extension. For most npm packages there's not really any formal indication either way unless there's specific fallback implementations.

You mentioned this in the other thread, but - in my experience - in those scenarios either the project is all one environment or a folder separates the different environments allowing Webpack to bundle each file differently.

bobaekang:
Instead of `"use client"` directive, how about a wrapper for client components that cross boundaries in the manner of `React.memo()` or `React.lazy()`? Maybe `React.serializable()`? Whatever it's called, a wrapper would look more familiar as API than a directive yet be no less viable for statically enforcing the rule.

jplhomer:
Following up on my comments from the previous thread:

- I really like this approach, because it's dramatically simpler and does not involve file name conventions.
- This will make it easier for developers to migrate to RSC — most components will "just work."
- The "client" wording issue is still unsolved, but I think the community might just get used to it. We haven't been able to come up with a better name.
- The only remaining concern is with knowing where you are in the module graph while editing a single component. But with poison pills and good compiler errors, the feedback to developers should be really fast when they introduce a hook they cannot use. Plus, if server components are largely async, that's yet another indicator about where you are in the tree.

Great work getting this pushed through!

gaearon (approved):

gaearon:
We're going to go ahead with this. Very thankful to everyone for the discussion on the original issue.

oljimenez:
> Following up on my comments from the previous thread:
> 
> * I really like this approach, because it's dramatically simpler and does not involve file name conventions.
> * This will make it easier for developers to migrate to RSC — most components will "just work."
> * The "client" wording issue is still unsolved, but I think the community might just get used to it. We haven't been able to come up with a better name.
> * The only remaining concern is with knowing where you are in the module graph while editing a single component. But with poison pills and good compiler errors, the feedback to developers should be really fast when they introduce a hook they cannot use. Plus, if server components are largely async, that's yet another indicator about where you are in the tree.
> 
> Great work getting this pushed through!

The problem is that now i usually creating my components with .client.tsx or .server.tsx because I need a ez way of difference them. Maybe this not need to be the core feature of React, but the "use client" is a little annoying. Have to be a better way. But after all is a amazing work 20/10.

zigang93:
I am agreed that "use client" quite annoying..
what if I need the same components use in client and server side? I need write everything twice ?
prefer something like `<SomeComponents client />` .. 

this open out flexibility that I can decide use same components to render under client or server side without WET.

devongovett:
Could a compiler automatically do this without needing a `"use client"` directive at all? For example, it could statically analyze the module and if there are any hook calls, automatically mark it as a client component. Seems like this could provide some benefits, such as automatically moving things to the server if the code is refactored without manually adding/removing a directive.

sebmarkbage:
The key is that when you move it, the props have to be serializable. This doesn't just mean that they use only the subset of types that are supported but also that those props aren't going to be an explosion in data transfer. There's benefit sometimes to move something to the client to make it a reusable template. Dev tooling will highlight this more.

It's also important that you teach patterns around what goes where for predictability.

I think some kind of "auto-fix" would be nicer to start.

That yea, it's totally possible. In fact, this idea spawned from work on Prepack to do exactly that. It just didn't work well for existing code that wasn't written with this code flow in mind.

devongovett:
That makes sense. I wonder if "use client" could just be a way of opting out of the automatic analysis. Seems like moving code to the server is usually better, and dev tooling could tell you when your serialized data is large. Just wonder how common it'll be to refactor some code (maybe deep in a custom hook) but forget to remove "use client".

sebmarkbage:
It's also about stability across tooling.

0xcadams:
Do you have any recommendations for using the `use client` directive in a bundled environment? I didn't see any mention of function-level directives in the RFC, and adding the `use client` module directive in a bundle would apply to all other modules in the bundle.

Related: https://github.com/rollup/rollup/issues/4699#issuecomment-1299770973

devongovett:
Another reason why I think automatic analysis would be a huge win: at the moment, tons of libraries are broken when rendered by a server component (the default in Next.js 13). Anything that uses state, context, etc. causes this. That's almost every component in a design system, for example. Unless these libraries are updated to add `"use client"` manually, the developer will get an (at the moment, quite cryptic) error. If the compiler analyzed these modules automatically, updating every library in the ecosystem wouldn't be necessary.

brillout:
> tons of libraries are broken when rendered by a server component

The workaround is to import React libraries from `"use client"` user-land components.

Since components with `"use client"` behave exactly as components currently behave today, there is 100% coverage of current libraries (as long as the user wraps them with `"use client"`). Or am I missing something here?

This makes me think that, although it'd be neat, automatic analysis isn't *that* much needed.

brillout:
As for the problem that the user doesn't know in what environments components are loaded (the uncanny valley problem), I think the "ultimate solution" would be IDE plugins showing in what env the component is loaded. For example a plugin changing the background color of the current opened file: green background => Server Component, blue background => Client Component, yellow background => Shared Component.
 
So I'm thinking there isn't any DX issues left, or am I missing something? (Personally I do think that "use client" solves the naming problem, since it describes exactly what's going on.)

kylemh:
> For example a plugin changing the background color of the current opened file: green background => Server Component, blue background => Client Component, yellow background => Shared Component.

Something like that doesn't exist, so I'd say it's a pretty gaping problem.

devongovett:
> The workaround is to import React libraries from "use client" user-land components.

This may require quite some refactoring if you have a server component and then add a library that doesn't have "use client" in it. You'd get an error, and need to manually wrap that library yourself. That will probably lead to lots of unnecessary issues getting filed against libraries across the ecosystem. I think the compiler could do this automatically and avoid all of this, plus keep apps more optimal over time as they change.

sebmarkbage:
Appreciate the thoughtful feedback. For things that only affect the current rollout, it's a little late though. It's possible an automatic compiler can work in the future but it likely won't happen as part of this wave. Therefore the effect on the ecosystem would still happen. E.g. the expectation for libraries to add `"use client"` would still happen and once that has passed, it's less of an issue anymore.

In the original discussion, we mentioned that one of the issues we were trying to address is that there is a lot of existing code that should just work. There's components but mainly it affects other types of imports as well like utility function. That's part of the reason why the original proposal used `.client.js` instead of just `.js` to annotate these.

If we made an opt in using something in package.json would that would make all utility functions in any JS library also "client modules". Since the split isn't just components but on the module level. Making that work another way is much more radically different approach than what we have here. Some of the earliest iterations did something more granular but there are other problems with that.

One thing we're trying to promote to tackle the uncanny valley is to, by convention, use async functions for Server Components. That helps a little bit the "background color" problem. It would be nice to have in a more specific form too. It needs to tie into the environment that tracks the root entry points to the system though. E.g. Next.js can have a plugin like this because it knows which files are root entry points to the system.

brillout:
> expectation for libraries to add `"use client"`

React libraries transpile JSX to plain JS. Is there a plan/thoughts for enabling library authors to transpile their library from RSC to plain JS?

Many ecosystems don't transpile the server-side code of libraries. Also, an increasing number of ecosystems (e.g. Vite) deliver code to the browser directly without any transpilation.

Forcing bundlers to process large React libraries (e.g. a large React components library) can have profound development speed degradation. (E.g. Vite's architecture is based on lazy-transpiling while externalizing as much code/libraries as possible.)

jaslong:
> Do you have any recommendations for using the `use client` directive in a bundled environment? I didn't see any mention of function-level directives in the RFC, and adding the `use client` module directive in a bundle would apply to all other modules in the bundle.
> 
> Related: [rollup/rollup#4699 (comment)](https://github.com/rollup/rollup/issues/4699#issuecomment-1299770973)

+1 to this. For example, suppose I have a package with a few files like this:

```tsx
// server.js
export function fetchData() { /* ... */ }

// client.js
import { useState } from "react"
export function DisplayData() {
  const [state, setState] = useState("this cannot run on RSC")
  /* ... */
}

// index.js
export * from "./server"
export * from "./client"

// package.json
{
  "name": "my-lib",
  "exports": "./dist/index.js"
}
```

And it all gets bundled into 1 `dist/index.js` file. Because of the bundling, it would not be possible for an RSC to `import { fetchData } from "my-lib"`, because `dist/index.js` includes the code from `client.js` and imports `useState`. The consequence here is that "use client" files can never be mixed with non-"use client" files in the same package for a [main entry point export](https://nodejs.org/api/packages.html#main-entry-point-export). However, I think [subpaths exports](https://nodejs.org/api/packages.html#subpath-exports) could work around this issue. Is this intended?

brillout:
> React libraries usually (always?) transpile JSX to plain JS before publishing the library. Is there a plan/thoughts for enabling library authors to transpile their library from RSC to plain JS?

If we do end up having JSX living in `node_modules/` then, at least, the JSX code should live in files ending with `.jsx` (and not `.js`). Otherwise that'd be significant performance regression for the entire Vite ecosystem. (Vite is based on lazy-transpilation as explained in my previous comment).

Vite transforms JSX only if strictly needed and uses the `.jsx` file extension to determine whether a file contains JSX.

bombillazo:
Hey, how could one opt-put of React Server Components when using a library like Next, which is enabling them by default and forcing refactoring to add `use client` for components?

gaearon:
@bombillazo `'use client'` is only needed at the entry point. So if you're using Next but want to opt out for your entire app, you can split each page into a server and a client part. Add `'use client'` to the client part and then from that point on, anything imported from that file will also be treated as client without an explicit annotation.

bombillazo:
@gaearon Thank you for the help! And thanks for the quick and prompt response!

HenriqueLimas:
Hi team! Really wanted to thank you for the new proposal from file extension to the `use ...` directive, it is clear and easier to understand! Learned a lot by reading the discussion, I really appreciate that you make this public.

I think that the name `React Server Components` and `React Client component` are confusing though. As my understanding RSC in the future can also run as part of build time not only on server. And also that `client` actually runs on the Server when doing SSR. I think that developers might start using `window` and any browser API on the render function when they see `use client` causing confusion and errors. 

My initial though (wrong one) of RSC was that they would render on the server and the hydrate/render would load it and put in the right place of the React application. But then after looking on the [initial demo](https://github.com/reactjs/server-components-demo) I noticed that it make use of a **Protocol** defined to serialize a component.

A different name could help preparing everyone for a different mental model of the current **standard** React components:

- React Protocol Components -> `'use protocol'`
- React Standard Component-> `'use standard'`
- React Shared Component -> `'use shared'`
- `"exports": { "react-protocol": "./..." }`

The reason to use `Protocol` (inspired by [protobuf](https://protobuf.dev/)) is to have something different than what we have today (SSR) to not cause confusion, so it force people having a different mental model.

Again, thank you for the huge work! RSC is really exciting technology!

HenriqueLimas:
But researching more, it seems that the community is more lean toward Server and Client components anyway where other frameworks also use Server/Client separation and introducing a new name might bring even more confusion to everyone.

jamiter:
I like the idea of simply adding "use client" to a file, but currently bundlers seem to strip away the "use client" directive automatically, causing issues in monorepos like mentioned here: https://github.com/nrwl/nx/issues/13194

If someone has an idea on how to work around this, please point me in the right direction.

Thanks for all the hard work put into this!

gaearon:
I messaged an Nx maintainer so they might be able to get a look at this.

gaearon:
>But researching more, it seems that the community is more lean toward Server and Client components anyway where other frameworks also use Server/Client separation and introducing a new name might bring even more confusion to everyone.

Yeah. The names aren't perfect (sometimes "Data Components" vs "Interactive Components" seems to make more sense) but in practice I think they have stuck.

juristr:
Hey, @jamiter a fix is released in Nx `v15.8.7`. Thanks for reaching out 🙏

thebuilder:
React libraries are starting to add `use client` to the bundled code they publish to NPM. This week [react-query](https://github.com/TanStack/query/pull/5161) added it - This is nice if you are running a Next.js application with support for RSC, but it impacts the output from all other bundlers like Rollup/Vite. 
Since RSC is not supported in Vite, logging that as a warning during build is not helpful. The user can't do anything to fix it.

```
packages/react-query/src/useMutation.ts (1:0)
1: 'use client'
   ^
2: import * as React from 'react'
3: import { useSyncExternalStore } from './useSyncExternalStore'
(!) Module level directives cause errors when bundled, 'use client' was ignored.
packages/react-query/src/useInfiniteQuery.ts (1:0)
```

I feel it would be a really good practice from your side, to ensure the ecosystem can handle `use client` properly by default, before it starts being rolled out further. You are going to see a continuous stream of questions/issues from users, getting the warnings from bundlers - and not knowing how to deal with it.

mrazauskas:
Hope this is the right place to leave feedback.

In Jest repo few users reported an issue related with the `'server-only'` package https://github.com/facebook/jest/issues/13967

Seems like users must pass `'react-server'` to `customExportConditions` of test environment to be able to run tests in Jest.

How do you think, would it make sense to add `"node": "./empty.js"` export to the `'server-only'` package? That would allow testing server only functions in Jest’s default Node environment.

If an extra export condition is not a good idea, I think it would be useful to document that `'react-server'` export condition should be passed to Jest. Not sure whether that should do mentioned in React’s or Jest’s docs.

sebmarkbage:
@mrazauskas It wouldn't be correct to add it to `"node": "./empty.js"` as that would by-pass the main purpose of that module. The goal is for it to throw by default to indicate that it doesn't work by default in existing environments that might be SSR hybrid.

Basically `'react-server'` is an opt-in that it's ok to run server-only code in this environment (and that it's *not* ok to run client code)`.

Configuring `jest` seems the correct solution but it's tricky in jest to test multiple environments at once. E.g. even if you do that, how do you ensure that you can test client components in the same test?

mrazauskas:
Ah right.. I didn't take into account SSR.

By the way, one user came to an idea (https://github.com/facebook/jest/issues/13967#issuecomment-1468987614) to use manual mock:

```js
// __mocks__/server-only.js
module.exports = {};
```

In this case all checks of `server-only` get erased. This could be another solution.

sebmarkbage:
The issue is that `server-only` is only one of possibly many different libraries that can fork. It's likely only to cover up other issues. E.g. `"react"` itself has a forked version of itself.

tomconroy:
Hello, I hope this the right place for this question. I have a React component library that uses Relay for data fetching for the components. I am trying to accommodate RSCs by preloading data based on [this pattern](https://github.com/relayjs/relay-examples/pull/241/files).

Is it sensible for libraries other than `server-only` and `client-only` to use these conditional exports, e.g. to export different components for server context vs client context? In the server context, we would fetch and serialize the relay data before rendering the client component. Then the client component can use the preloaded serialized data, or lazy load if omitted.

aralroca:
Is there a way to specify in the same file which components are "client" and which are "server"? To use my libraries I use microbundle and it puts everything in the same file, I like microbundle but as a library maintainer should I change the bundler to create it in different files or is there a better way to do it? 🤔 Thanks

text/0000-server-module-conventions.md
```diff
@@ -0,0 +1,178 @@
...
+A package can choose to fork its exports depending the environment it executes in across two dimensions.
+
+The `"react-server"` condition applies only React Server Component environments. However, these can execute either in a `"browser"` during development for debugging purposes, or it can execute in a `"node"` environment in production.
+
+React Client Component environments are the default and they typically execute in a `"browser"` environment but they can also execute in a `"node"` environment for SSR purposes.
+
+Note: This condition can only be used for conditional `imports` within the project itself.
+
+Possible addition: A possible addition to this can be to also support the `"server-only"` condition as a less React specific alternative that could be adopted by other frameworks that also choose this strategy.
+
```
Jaxensdad2012: Tyler_mashburn

Jaxensdad2012 (commented):
99)`

mgmolisani:
Bit late to the party here, but I'm curious if there was any prior artwork to the decision to use a directive ([Directive Prologue](https://262.ecma-international.org/5.1/#sec-14.1)). Is there any risk here to polluting the global space with directives? Let's say another library that is not a front end framework wants to ship similarly bundled code for the client and server and uses these directives (as they are fairly generically named). Would those libraries have to avoid these names? Would they have to work with bundling tools to get their own namespace? Or can this be solved through additional pre-processing and configuration around the directive naming?

PR #243, opened 2023-03-03
[WIP] [DRAFT] RFC: React Module Conventions RFC v3

Pranav-yadav:
[WIP] [DRAFT]

---

<sup>Originally posted by me; under related RFC v2 PR.</sup>

# Abstract:
In this RFC I propose minor yet effective changes to [the React Module Conventions RFC v2](https://github.com/reactjs/rfcs/pull/227). The term **_"client"_** is too generic both in Software/Web Development as well as in general. This type of _**extra** generalism_ may lead to devastating effects on DX in various aspects. I propose; we should use `"use csr"` instead of `"use client"` to avoid confusion and generalism, by being explicit and not too generic.

# Concerns:

1. `"use client"` : both experienced and newbie React Developer would be confused with the word "client", it's too **_generic_**, and on top of that, it sounds like we're using hooks. <sub>(feeling sorry for classes; no one mentions them these days 😎)</sub>
  I get it, it's not that hard to get used to this but 👇 also, <sub>(see proposal after reading both of these concerns)</sub>
2. This is going to be tough to digest for JS developers, at least for those still working with/seeing `use strict`, in their free time :).
  Since `"use strict"` is supported by the language itself out-of-the-box. Imitating the same syntax will confuse some JS devs and newbies that; `"use client"` is something that's also part of official JS [^1] (which could be devastating).
  <sub>(I get that some of us may be thinking only ɗŭmb devs will have this problem, but, hey, wasn't you also a newbie at some time?)</sub>.
  Don't have hard suggestions on this but _this confusion can be reduced by being_ **_explicit_**.

# Proposal:

1. Using **`"use csr"`** : here we're _being_ **_explicit_**. As almost every frontend/backend dev is familiar w/ what is **CSR** / **`csr`**, it's _not_ that much confusing as compared to **_`client`_**
```diff js
+ 'use csr';
- 'use client';
function doSomething() {
 // ...
}
```

2. Let's _**not**_ use **_too generic_** term and confuse everyone instead let's use specific terms like SSR (ssr) and CSR (csr), that way we're trying to be _explicit_ and the literal meaning is conveyed, leading to improved DX and flexibility.

3. Reserving `"use ssr"` for future use in SSR functionalities i.e. in the context of SSR.
4. Reserving `"use ssg"` for future use in SSG functionalities i.e. in the context of SSG.
5. Reserving `"use isr"` for future use in the context of ISR.

# Glossary

- CSR / `csr`: **C**lient **S**ide **R**endering
- SSR / `ssr`: **S**erver **S**ide **R**endering
- SSG / `ssg`: **S**tatic **S**ite **G**eneration
- ISR / `isr`: **I**ncremental **S**tatic **R**egeneration

# References
[^1]: > **NOTE**
The [ExpressionStatement](https://tc39.es/ecma262/#prod-ExpressionStatement)s of a [Directive Prologue](https://tc39.es/ecma262/#directive-prologue) are evaluated normally during evaluation of the containing production. Implementations may define implementation specific meanings for [ExpressionStatement](https://tc39.es/ecma262/#prod-ExpressionStatement)s which are not a [Use Strict Directive](https://tc39.es/ecma262/#use-strict-directive) and which occur in a [Directive Prologue](https://tc39.es/ecma262/#directive-prologue). If an appropriate notification mechanism exists, an implementation should issue a warning if it encounters in a [Directive Prologue](https://tc39.es/ecma262/#directive-prologue) an [ExpressionStatement](https://tc39.es/ecma262/#prod-ExpressionStatement) that is not a [Use Strict Directive](https://tc39.es/ecma262/#use-strict-directive) and which does not have a meaning defined by the implementation. <p align="right">--- [_TC39/ecma262_](https://tc39.es/ecma262/#use-strict-directive)</p>

### DIFF
--- a/text/0000-server-module-conventions.md
+++ b/text/0000-server-module-conventions.md
@@ -1,0 +1,136 @@
+- Start Date: 2023-02-04
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+- Author: Pranav Yadav [@Pranav-yadav](https://github.com/Pranav-yadav) 
+
+# Summary
+
+Better conventions for Server/Client Components, as an revision to [the v2](https://github.com/reactjs/rfcs/pull/227) of [the original proposal v1](https://github.com/reactjs/rfcs/pull/189)
+
+<sup>Note: The v1 and v2 terms used throughout this RFC refer to the above hyperlinked RFCs only.</sup>
+
+# Changes Since v2
+
+- using `"use csr"` instead of `"use client"` i.e. replacing `client` with `csr`
+- similarly, reserve `"use ssr"` for future use with Server only.
+
+# Basic example
+
+<!-- If the proposal involves a new or changed API, include a basic code example.
+Omit this section if it's not applicable. -->
+
+To mark a component as a Client Component, add `'use csr'` directive at the top of your file. This lets you use features like state and event handlers:
+
+```js
+'use csr';
+
+import { useState } from 'react';
+
+function Button({ children, onClick }) {
+  // ...
+}
+```
+
+Diff - current syntax and conventions:
+```diff
++'use csr';
+-'use client';
+
+import { useState } from 'react';
+
+function Button({ children, onClick }) {
+  // ...
+}
+```
+
+## `"use csr"` Directive
+
+```js
+// Parent.js
+import Component from "./Component.js";
+function Parent() {
+  // Component is of type Reference<T> where T is the type
+  // of the default export of Component.js
+  // Because of this React knows that it can't render it on
+  // the server and instead will leave it as a placeholder
+  // to later be rendered on the client.
+  return <Component />;
+}
+```
+
+```js
+// Component.js
+"use csr";
+
+function Component() {
+  let [state, setState] = useState(false);
+  return <OtherClientComponent onClick={() => setState(true)} value={state} />;
+}
+```
+
+# Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+Please focus on explaining the motivation so that if this RFC is not accepted,
+the motivation could be used to develop alternative solutions. In other words,
+enumerate the constraints you are trying to solve without coupling them too
+closely to the solution you have in mind. -->
+
+
+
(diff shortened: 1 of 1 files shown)


Pranav-yadav:
@sebmarkbage, @gaearon any _initial_ comments/suggestions (about the concept, language, structure, etc.) on this RFC?
I would be happy to rewrite/update this RFC

gaearon:
It's not right because this isn't quite what it does.

`'use client'` is not the same thing as CSR. It means you get both CSR *and* SSR/SSG/ISR (depending on your framework and other settings).

The correct way to think of `'use client'` is not as a rendering mode. It's more like a split-point. It's where you cut off the server-only part of your code (which is the default at the module root and down) from the part of your code that needs to run on *both* server and the client (that's the one under `'use client'` and anything imported from it).

This is still a bit confusing.

Another way you can think of it is that the only real "server" here is RSC. It's a true server in the sense that it has access to everything on the server. You could put it close to the database, and read directly. Now, your SSR is a _client_ of that server. Your SSR could be on the edge — almost where the actual user (client) is. Your SSR's role is really to _emulate what the client would do_ (and do it faster because you don't need to fetch code). That's why your SSR renders only the initial state of your client components. It's like a "snapshot" of the first frame that you can send very fast so the browser has something to show. This is why, conceptually, SSR is another "client" in our architecture.

Yes, this naming is confusing, but that's also because it adds a new piece to the architecture which slightly changes the roles of the other boxes on the diagram. I wish I knew a less confusing option, but this is not the one.

Thanks for writing this down.

Pranav-yadav:
Thanks. I'll read more about the concepts _(and the internals)_ and try to play with them in order to get thorough understanding. 
Again, thanks for explaining it very clearly :)

PS: If it's possible please document your comments above in on react docs under advanced concepts/guides. Originally I looked at the docs there's not a _single_ mention about it on **react** docs, (yet it's integral part of react, I know it's not required by day-to-day development). I did find good documentation related to this on **Next's** docs but after writing this raw-draft rfc, and didn't get time to complete it.

PR #248, opened 2023-04-18
RFC: Implementing dynamic css classes in React components

push1kb:
The proposed feature is to add a new syntax for dynamically adding classes to React components. This syntax would be similar to the `ngClass` directive in Angular, allowing developers to use a `classNames` attribute to add or remove classes based on certain conditions.

[View RFC](https://github.com/vijaypushkin/react-rfcs/blob/feat/dynamic-class-names/text/0000-dynamic-class-names.md)

[View PR](https://github.com/facebook/react/pull/26730)

### DIFF
--- a/text/0000-dynamic-class-names.md
+++ b/text/0000-dynamic-class-names.md
@@ -1,0 +1,82 @@
+- Start Date: 2023-04-18
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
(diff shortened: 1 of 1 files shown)


OlegLustenko:
Facebook doesn't use classes heavily in internals.

And this could be implemented in the user land.

More likely it will never be adopted.

However I like your spirit

Issue #168, opened 2020-06-06
isolation-hook and a new pattern to pass hooks to another(hook or hook creator)


MHA15:
One general optimization we could do is to eagerly call the render function during propagation and see if the render yields a different result and only if it does do we clone the path to it.

Another way to look at this selector is that it's just a way to scope a part of that eager render. Another way could be to have a hook that isolates only part of the rerender.

```js
let x = expensiveFn(props);
let y = useIsolation(() => {
   let v = useContext(MyContext);
   return selector(v);
});
return <div>{x + y}</div>;
```

The interesting thing about this is that it could also be used together with state. Only if this context has changed or if this state has changed and that results in a new value, do we rerender.

```js
let x = expensiveFn(props);
let y = useIsolation(() => {
   let [s, ...] = useState(...);
   let v = useContext(MyContext);
   return selector(v, s);
});
return <div>{x + y}</div>;
```

Another way to implement the same effect is to just memoize everything else instead.

```js
let x = useMemo(() => expensiveFn(props), [props]);
let [s, ...] = useState(...);
let v = useContext(MyContext);
let y = useMemo(() => selector(v, s), [v, s]);
return useMemo(() => <div>{x + y}</div>, [x, y]);
```

It's hard to rely on everything else being memoized today. However, ultimately I think that this is where we're going. Albeit perhaps automated (e.g. compiler driven).

If that is the case, I wonder if this API will in fact be necessary or if it's just something that we get naturally for free by memoizing more or the *other* parts of a render function.

_Originally posted by @sebmarkbage in https://github.com/reactjs/rfcs/pull/119#issuecomment-571002690_

well this post make me so interested 
there were many good replies to this post by saying that it's so easier to isolte  hook instead of memoizing all other parts 
and there is only one problem on using useIsoltion as writes above which is this can break rules of hooks and I will give you an easy way to prevent breaking rules of hooks 

using a hook creator named createIsolateHook like 
```js
const useIsoltedUser = makeIsoltedHook(useUser)
```

this approach will also preventing isolated hook to use other scoped variables of component like setState or state without passing explicitly to our isolated hook

MHA15:
@dai-shi
@theKashey
@GasimGasimzada

I think that you want to be participate

thanks in advance

theKashey:
For me one of the current limitations of hook is that any change triggers a `whole thing`.
[...]

PR #257, opened 2023-12-13
RFC: `useIsolation`

Ayc0:
This RFC related to:
- `useIsolation` in https://github.com/reactjs/rfcs/issues/168,
- `useContextSelector` in https://github.com/reactjs/rfcs/pull/119.

This is about performance & memoization

[**>>> View rendered text <<<**](https://github.com/Ayc0/react-rfcs/blob/main/text/0000-use-isolation.md)

### DIFF
--- a/text/0000-use-isolation.md
+++ b/text/0000-use-isolation.md
@@ -1,5 +1,5 @@
 - Start Date: 2023-12-13
-- RFC PR: (leave this empty)
+- RFC PR: https://github.com/reactjs/rfcs/pull/257
 - React Issue: (leave this empty)
 
 # Summary
--- a/text/0000-use-isolation.md
+++ b/text/0000-use-isolation.md
@@ -39,22 +39,21 @@ This topic is mostly for performance reasons. A few of other RFCs are proposing 
 
 When working with components, you can bail out of re-renders with `React.memo` / `shouldComponentUpdate()`.
 
-And given this tree: `<A><B/></A>`, when `A` changes, `B` may not rerender.
-But when working with hooks, if you have 2 hooks `useA` and `useB` (used within `useA`), when `useB` re-renders,
-there is no way to bail out from its renders in neither `useA`.
+And given this tree: `<A><B/></A>`, when `A` changes, `B` may not rerender.<br>
+But when working with hooks, if you have 2 hooks `useA` and `useB` (used within `useA`), when `useB` re-renders, there is no way to bail out from its renders in neither `useA`.
 
-This is an issue for large codebases that can share a lot of hooks (without having the ability of auditing all of them).
+This is an issue for large codebases that can share a lot of hooks (without having the ability of auditing all of them).<br>
 Or for libraries like `react-router-dom` that expose large objects (the router state) and where users want to only register for changes that matter to them.
 
 # Detailed design
 
 This hook is inspired by both `React.memo` where you can provide a `areEqual` function as a 2nd parameter, and by `useMemo`.
 
-When you want to make a piece of code run in _isolation\*_ from its parent component/hook, you can use `useIsolation`.
+When you want to make a piece of code run in _isolation\*_ from its parent component/hook, you can use `useIsolation`.<br>
 This hook takes 2 parameters:
 
-Using MDN’s notation: `useIsolation(callback, [deps])`
-Using TS’s notation: `useIsolation<T>(callback: () => T, deps?: ReadonlyArray<unknown>)`
+- Using MDN’s notation: `useIsolation(callback, [deps])`
+- Using TS’s notation: `useIsolation<T>(callback: () => T, deps?: ReadonlyArray<unknown>)`
 
 \*: this wouldn’t strictly run in isolation, as you may define dependencies to ease the communication between the parent scope and the isolated one.
 
@@ -126,7 +125,7 @@ With this piece of code,
 
 # Drawbacks
 
-The base principle of this new hook is to be able to create new _call scope_ (aka component-like scopes).
+The base principle of this new hook is to be able to create new _call scope_ (aka component-like scopes).<br>
 But this may be a huge change in React’s internals.
 
 As this is deeply related to this "component-like scope", it’s also impossible to polyfill / re-create on the user world and has to be implemented within React (I may be wrong on this).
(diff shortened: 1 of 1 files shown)


EECOLOR:
If I understand correctly you want the equivalent of the following:

Usage:
```jsx
import React from 'react'

export function App() {
  console.log('render App')
  return (
    <IsolationProvider>
        <Test />
    </IsolationProvider>
  );
}

function Test() {
  console.log('render test')

  const x = useIsolation(() => {
    const [x, setX] = React.useState(0)
    React.useEffect(
      () => {
        let interval = setInterval(() => setX(x => x + 1), 1000)
        return () => { clearInterval(interval) }
      },
      []
    )
    return React.useMemo(() => x - (x % 2), [x])
  })

  return <p>{x}</p>
}
```

Non-optimized user space implementation:
```jsx
const isolationContext = React.createContext(null)

function useIsolation(unsafeHook) {
  const hook = useEvent(unsafeHook)

  const [result, setResult] = React.useState(null)

  const registerHook = React.useContext(isolationContext)

  React.useEffect(
    () => registerHook({ hook, callback: (...args) => setTimeout(() => setResult(...args), 0) }),
    []
  )

  return result
}

function IsolationProvider({ children }) {
  console.log('render isolation provider')

  const [hookInfo, setHookInfo] = React.useState([])

  const registerHook = React.useCallback(
    (hookInfoToBeIsolated) => {
      setHookInfo(existing => existing.concat(hookInfoToBeIsolated))
      return function cleanup() {
        setHookInfo(existing => existing.filter(info => info !== hookInfoToBeIsolated))
      }
    },
    []
  )

  return (
    <isolationContext.Provider value={registerHook}>
      {children}
      {hookInfo.map((info, i) => 
        // key should be handled differently
        <Isolation key={i} {...{ info }} />
      )}
    </isolationContext.Provider>
  )
}

function Isolation({ info }) {
  const { callback, hook } = info
  const result = hook()

  console.log('hook executed', result)

  useCallOnChange({ ifChanged: result, callback })

  return null
}

function useCallOnChange({ ifChanged, callback }) {
  const changeRef = React.useRef(null)
  if (changeRef.current !== ifChanged) callback(ifChanged)
  changeRef.current = ifChanged
}

function useEvent(f) {
  const fRef = React.useRef(null)
  fRef.current = f

  return React.useCallback((...args) => fRef.current(...args), [])
}
```

Ayc0:
> If I understand correctly you want the equivalent of the following:

Yes, the idea is here with 2 differences:
- `useIsolation` could run synchronously during the 1st render (like a `useMemo`)
- you could also provide dependencies to `useIsolation(hook, [deps])` to conditionally re-call the hook when its parent scope changes (but not if hooks within it would trigger a re-render)

gaearon:
Thank you for the RFC. I wanted to note that we’ve had a very similar proposal planned except that we wanted to roll this behavior into the existing `useMemo` Hook.

jaskp:
Wouldn't this concept of isolation also be useful for creating suspense boundaries around hooks? 

```ts
const useSuspenseFoo = (): Foo => React.use(fooPromise)

const ComponentThatDoesNotSuspend = () => {
    const maybeFoo: Foo | undefined = React.useIsolation(() => useSuspenseFoo())
}
```

Ayc0:
> Wouldn't this concept of isolation also be useful for creating suspense boundaries around hooks?
> 
> ```ts
> const useSuspenseFoo = (): Foo => React.use(fooPromise)
> 
> const ComponentThatDoesNotSuspend = () => {
>     const maybeFoo: Foo | undefined = React.useIsolation(() => useSuspenseFoo())
> }
> ```

I don't know, not fully sure 🤔 
Suspense boundaries need to have fallbacks. Maybe if an API like what Preact does with [`useErrorBoundary`](https://preactjs.com/guide/v11/hooks/#useerrorboundary) could work here?

Like another `useSuspense()` (that internally could use `useIsolation()` indeed), like:

```ts
const [value, isSuspensed] = useSuspense(() => useSuspenseFoo())
```

But I'd move this to another RFC

Ayc0:
That said, based on the React's team direction, this RFC is highly certain to never exist. 

I thought a lot about this RFC and I still think it could be awesome as:
- it gives us a lot of control on how React can work,
- allow us more fine grained detailed optimization,
- allow us to optimize 3rd party dependencies,
- and in conjunction with https://github.com/reactjs/rfcs/pull/261, pairs well with the Compiler

But, based on how the compiler auto optimizes things, and based on the fact we can now do `use(Context)`, if someone would use the compiler and do something like:

```tsx
const Foo = () => {
  const foo = React.useContext(CustomHeavyContext).foo;
};
```

or

```tsx
const Bar = () => {
  const bar = useSearchParams().get("bar");
};
```

then the variables `foo` & `bar` would be memoized deeply. And even if `Foo` & `Bar` would auto-re-render when any of the context / search params would change, no other variable using `foo` nor `bar` would change, and neither would children of `Foo` & `Bar`. So maybe we don't need this RFC anymore with the compiler.

NotYoojun:
> So maybe we don't need this RFC anymore with the compiler.

I really appreciate the idea of rolling this behavior into `useMemo`, but I want to emphasize that **not everyone will be using (or be able to use) the React compiler** or relying on compiler-driven optimizations. Having `useIsolation` as a dedicated hook provides a clear, explicit API that developers can adopt regardless of their tooling setup.  

So while I understand the motivation to unify this into `useMemo`, I **strongly agree that `useIsolation` should be added as its own hook**. Having it as a dedicated hook provides flexibility for developers who prefer explicit APIs, and ensures we retain control over when and how isolation is applied.

Ayc0:
@NotYoojun I agree with you: a lot of companies can't use the compiler as it's babel based (and thus too slow in their build pipelines).

Also idk if unification with `useMemo` I think was shipped and instead if they aren't pushing for the compiler more 🤔 (I need to check again in the docs)

NotYoojun:
> Also idk if unification with `useMemo` I think was shipped and instead if they aren't pushing for the compiler more 🤔 (I need to check again in the docs)

Currently `useMemo` does not support calling hooks inside, right? As far as I can remeber calling a hook in useMemo will result in a invalid hook call error, neither the React documentation mentioned anything about that

NotYoojun:
I do believe the implementation of `useIsolation` can resolve a massive amount of React performance issue, as well as making `useContextSelector` possible and flexible. We do have a userland useContextSelector by dai-shi. However, it is just simulating the behavior. The data flow is still updated async-ly (using useLayouEffect), thus not a real solution.

The community has been expecting this for a while: https://x.com/TkDodo/status/1741193371283026422.

However, I do understand that implementation requires a certain amount of effort (and probably even massive refactoring), so I really appreciate all your effort.

PR #258, opened 2024-06-10
First proposal of the reflection in React

Nefcanto:
Hi guys,

I have added a proposal for reflection in React. Please let me know if I've done it correctly or not.

Thank you so much.

### DIFF
--- a/0000-reflection.md
+++ b/0000-reflection.md
@@ -1,0 +1,262 @@
+- Start Date: 2024-06-10
+- RFC PR:
+- React Issue:
+
+# Summary
+
+Reflection let's dynamic applications to analyze a given component and extract information about it.
+
+# Basic example
+
+```
+import {
+    isArray,
+    hasProp,
+} from "react-reflection"
+
+const ParentComponent = ({ children, ...rest }) => {
+
+    const isArray = isArray(children)
+    let filteredChildren
+    if (isArray)
+    {
+        filteredChildren = children.filter(i => !hasProp(i, 'superAdmin'))
+    }
+    else
+    {
+        filteredChildren = hasProp(children, 'superAdmin') ? null : children
+    }
+
+    return <>
+        <h1>Parent component</h1>
+        {
+            isArray
+            ?
+            <ul>
+            {
+                filteredChildren.map((child, index) => <li key={index}>{child}</li>)
+            }
+            </ul>
+            :
+            children
+        }
+    </>
+}
+```
+
+# Motivation
+
+Right now developers have no way to extract information about a passed component.
+Knowing about a passed component opens the door for more dynamic systems, yet with less verbosity.
+
+# Detailed design
+
+Let's create a dynamic `<Icon>` component. It can accept a wide range of inputs. Let's support a range of possible inputs:
+
+- Inline SVG: `<Icon show={<svg></svg>} />`
+- Third party component:   
+```
+import SettingsIcon from "@mui/icons-material/Settings"   
+<Icon show={SettingsIcon} />
+```
+- Custom function components:  
+```
+const CustomIcon = () => <div>I am a custom icon</div>  
+<Icon show={CustomIcon} />
+```
+- A variable:   
+```
+let icon
+switch (entity.type) {
+    case "instructor":
+        icon = TeacherIcon
+    case "student":
+        icon = StudentIcon
+}
+<Icon show={icon} />
+```
+- A fragment:
+```
+const fragment = <>
+    <div>Some shape here</div>
+    <div>Another shape here</div>
+</>
+<Icon show={fragment} />
+```
+
+Now let's build this component:
+
+```
+import {
+    isFragment,
+    isFunctionComponent,
+}
+const Icon = ({ show }) => {
+    if (isFunctionComponent(show)) {
+        const Component = show
+        return <Component />
+    }
+    if (isFragment(show))
+    {
+        return show
+    }
+}
+```
+
+Reflection has been around in many languages and the have proven to be a great tool when it comes to make things more dynamic.
+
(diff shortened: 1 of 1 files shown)


0000-reflection.md
```diff
@@ -0,0 +1,262 @@
...
+        return <Component />
+    }
+    if (isFragment(show))
+    {
+        return show
+    }
+}
+```
+
+Reflection has been around in many languages and the have proven to be a great tool when it comes to make things more dynamic.
```
landsman: Yes, reflection in Kotlin is pure gold. I love it.

Nefcanto: @landsman, we're developing in C# and React. C# (.NET in fact) has reflection and we have made amazing dynamic functionality using reflection.

One example is fake data insertion. We analyze a given entity and create fake data based on its type. Developers can override this behavior, but they get 90% good results from the base itself. In a system with more than 500 entity types in it, that's an amazing value we can deliver.

Another real example is the security check in our controllers. We inspect each action via our homemade linting tool and make sure that no mass-assignment vulnerability could be exposed by analyzing each parameter.

Another usage is a centralized change log. We use reflection and interceptors to answer:   
Who, when, created what (with fields), edited what (changed fields), deleted what (deleted fields).

We have tons of these examples. All made possible only because reflection was there. Reflection is a very powerful tool indeed.

theKashey:
Wondering which particular helpers you have in mind, because the ones listed in the RFC are covered by native js functionality (`Array.isArray`) or [react-is](https://www.npmjs.com/package/react-is)

Nefcanto:
@theKashey, thank you for the link. I did not know about `react-is`. But as I looked at it and the name suggests, it's only about checking a fact about a given thing.

Let me give a list of ideas I have in mind for `react-reflection`:

1. Retrieving information about the app itself => something like `getComponents`
2. Finding a specific component => `getComponent("ComponentName")`
3. Get information about props => `getProps(component)`
4. Get one specific prop => `getProp(component, "PropName")`
5. Check for the presence of a prop => `hasProp(component, "PropName")`
6. Get the list of hooks used in the component => `getHooks(component)`
7. Get the list of states used in the component => `getStates(component)`
8. Get the list of dependencies for a given hook => `getDependencies(hook)` or `getDependencies(component, "HookName")`
9. Getting the type of an object (not checking if it is something) => `getType(object)` => These could be: Fragment, Array, FunctionComponent, ClassComponent, Hook, State, etc.
10. Creating a dynamic instance of a component if possible, without being in DOM => `createInstance(component)`

The list can go on. We do use all of these reflection ideas in C#.

EECOLOR:
I might be missing something.

From my perspective reflection in programming language is mostly there because the programmer "lost" some information. Sometimes this is by accident, in other cases this is because there is a "dynamic" situation where the writer of the code does not know all the details of how it is being used.

In my experience, when using a form of reflection, most of the time the complexity is introduced where it could have been prevented at another point in the program. A simple example:

> Let's say I am receiving an array with mixed types, something like `[true, 1, { banana: 'yellow' }]`. If I need to process this information I might need to do a check for `typeof`. I could have prevented this by supplying different data: `[{ type: 'boolean', value: true }, { type: 'number', value: 1 }, { type: 'object', value: { banana: 'yellow' } }]`.

So in most cases I try to resolve the need for reflection by making code more explicit. For me reflection is a red flag that signals accidental complexity in my application.

What is the use case you need reflection for?

Nefcanto:
@EECOLOR, your example is true. We can add more data to the leaf nodes. The more data we feed, the less data we need to extract dynamically.

But this approach has many downsides:

1. It works as long as the leaf components are under your control. You won't be able to do this if you do not have access to the code you want to use. Third-party components won't be covered.
2. It increases boilerplate code in leaf components by the multitude. Let's say you have 500 lists in an application. And you want to send each list a bunch of actions. One list might have a simple "icon + text" action. Another list might need an array of iconic actions. Which one makes more sense? Providing the metadata in 500 places? Or extracting the metadata in one place? This is one example of the requirements we do have. Consider this list:

```
const Customers = () => {
    return <List
        title="Customers"
        listActions={<ListAction 
                title="CalculateBalances"
                iconic
            />}
    />
}
```
And compare it to:

```
const Customers = () => {
    return <List
        title="Customers"
        listActions={<ListAction 
                title="CalculateBalances"
                iconic
            />}
        listActionsType="singleComponent"
        listActionsPresentation="iconic"
    />
}
```

The first component reduces the boilerplate code. That counts as the system grows.

3. Providing metadata sometimes becomes extremely tricky. Extracting metadata is a better approach for some use cases. Consider these lists:
```
const Customers = () => {
    return <List
        title="Customers"
        listActions={<ListAction 
                title="CalculateBalances"
                iconic
            />}
    />
}

const Users = () => {
    return <List
        title="Users"
        listActions={<>
            <ListAction title="SendNotification" />
            <ListAction title="AnalyzeForVulnerabilities" />
        </>}
    />
}

const Orders = () => {
    return <List
        title="Orders"
        listActions={<>
            <ListAction title="CalculateTotals" />
            <ListActionMenu>
                 <ListAction title="MakeSpreadsheet" />             
                 <ListAction
                       title="ApplyDiscount"
                       admin
                 />
            </ListActionMenu>
        </>}
    />
}

```

What metadata can we provide to signal to the `List` component that the list actions have a menu in it, and one item of that menu should only be shown for admin users?

The point is, while your example is true, there are some use cases when providing more metadata to the system is not a better approach, and extracting data from the system makes more sense.

Nefcanto:
I didn't want to close this. I just wanted to clean the clutter on my account. I didn't know that deleting a fork causes this to be closed. This feature is still very much needed indeed.

Nefcanto:
I'm opening this, as we do need this feature for our complex SPL.

PR #259, opened 2024-11-06
Creating light components

Nefcanto:
The proposal for light components. Light components are even lighter than function components. They don't accept props. They don't have states. They are just a simple JSX constant/variable that can be rendered like a normal component using `<LightComponent />`.

### DIFF
--- a/0000-light-components.md
+++ b/0000-light-components.md
@@ -1,0 +1,141 @@
+- Start Date: (2024-10-6)
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+React at first had class components. Creating a class component was hard. So it introduced function components.
+Function components were easier to create. They were more readable. They had less boilerplate.
+Then server components came into the scene.
+
+Now I would like to propose *light components*. They are simple valid JSX variables or constants. They do not have props and they do not have states. They are just a simple JSX bundle.
+
+# Basic example
+
+```
+const Contacts = <div>
+   <div>some phone number</div>
+   <div>some email</div>
+</div>
+
+<Contacts />
+```
+
+# Motivation
+
+It's all about unnecessary boilerplate. Our team has built an SPL (software product line). The panel is built using React.
+We minimize the boilerplate as much as we can. The reduced boilerplate adds direct business value to us (enhanced maintainability, lower learning curve, faster time-to-market).
+
+One aspect of the boilerplate is to be forced to add at least two parentheses and a lambda (4 characters) to turn a JSX constant into a function component:
+
+```
+const Contacts = () => <div>
+   <div>some phone number</div>
+   <div>some email</div>
+</div>
+
+<Contacts />
+```
+
+It does not look much in one single instance. But multiply it by more than 500 lists that we have, and each list has a couple of JSX constants and the value shows itself.
+
+For example, look at this `/SomePath/Product/List.jsx` file:
+
+```
+const headers = <>
+    <th>Name</th>
+    <th>Price</th>
+</>
+
+const row = product => <>
+    <td>{product.name}</td>
+    <td>{product.price}</td>
+</>
+
+const listActions = <>
+    <Categories />
+    <Tags />
+    <Attributes />
+    <Export />
+</>
+
+const entityActiions = <>
+    <Images />
+    <Categories />
+    <Tags />
+    <Price />
+</>
+
+const Products = <List
+    title='Products'
+    entityType='Product'
+    headers={headers}
+    row={row}
+    listActions={listActions}
+    entityActions={entityActions}
+/>
+
+export default Products
+```
+
+Rendering a JSX fragment inside a file/component is not hard. You render it like `{jsxFragment}`. But when it comes to big systems where you split the system into parts, and develop part separately and reuse parts, that's when it becomes hard.
+
+# Detailed design
+
+I recommend that we start with a build configuration flag. In CRA (WebPack) or Vite or any other build system we can configure React to treat JSX framgnets/constants as components:
+
+```
+{
+    treatJsxAsComponent: true
+}
+```
+
+Then we should be able to capitalize the first letter of the JSX constant/variable and render it like a normal component:
+
+```
+const LightComponent = <div>Light component without props and state</div>
+
+<LightComponent />
+```
+
+The point is that light components won't accept props (they can ignore them with a warning in the console) and they won't have states.
+
+```
+<LightComponent someKey="someValue" />
+// creates a warning in the console that light components won't support props and those props will be ignored
+```
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+Since we can create this behavior as a progressive configurational behavior, I can think of no drawback. Any team who wants light components can turn on a flag and get a new behavior. Otherwise, React is just like how it was before.
+
+# Alternatives
+
+There are a couple of alternatives. One is to use something like a higher-order component to get the JSX constant, and render it.
+
+```
+const Contacts = <div>Some contact data here</div>
+
+const JsxRenderer = ({ jsx }) => {
+    return <div>
+        {jsx}
+    </div>
+}
+
+<JsxRenderer jsx={Contacts} />
+```
+
+But that adds to the boilerplate even more. It defies the first goal which is to reduce the boilerplate.
+
+# Adoption strategy
+
+This feature won't be a breaking change. At first, it can be enabled via a flag. That way no team would be affected by it.
+It can stay as a flag for the rest of the React lifecycle.
+However, if it become a permanent behavior, then no team should be affected. Because both `<LightComponent />` and `{LightComponent}` should be OK and work.
+
+# How we teach this
+
+A new section should be created in the docs called `Light components`. It should explain some limitations for these types of components.
+It should also present some examples on when are they useful. And example could be a `<SalesBadge />` simple JSX that can be created in one place and be reused across an online shop without any change.
--- a/text/0188-server-components.md
+++ b/text/0188-server-components.md
@@ -60,7 +60,7 @@ This example renders a simple Note with a title and body. It renders a non-edita
 
 import db from 'db'; 
 // (A1) We import from NoteEditor.js - a Client Component.
-import NoteEditor from 'NoteEditor';
+import NoteEditor from 'noteEditor';
 
 async function Note(props) {
   const {id, isEditing} = props;


0000-light-components.md
```diff
@@ -0,0 +1,141 @@
...
+# Basic example
+
+```
+const Contacts = <div>
+   <div>some phone number</div>
+   <div>some email</div>
+</div>
+
+<Contacts />
+```
```
slorber, line 22: Why not just do this instead?

```tsx
const contacts = (
  <div>
   <div>some phone number</div>
   <div>some email</div>
  </div>
)


const app = (
  <>
    <h1>My App</h1>
    {contacts}
  </>
);
```

I really don't see the point of introducing this new concept. Instead of light components, you can simply use constant JSX elements.

0000-light-components.md
```diff
@@ -0,0 +1,141 @@
...
+const Contacts = <div>Some contact data here</div>
+
+const JsxRenderer = ({ jsx }) => {
+    return <div>
+        {jsx}
+    </div>
+}
+
+<JsxRenderer jsx={Contacts} />
+```
```
slorber, line 128: What's the point of `JsxRenderer` here?

You can just write this and skip this useless boilerplate

```tsx
const Contacts = <div>Some contact data here</div>

{Contacts}
```

slorber (commented):
> Light components are even lighter than function components. They don't accept props. They don't have states. They are just a simple JSX constant/variable

So, they can just be plain constant JSX elements

> that can be rendered like a normal component using `<LightComponent />`.

It's unclear what's the point of allowing this instead of just `{LightComponent}`. If it's only about removing boilerplate, `{LightComponent}` saves 2 extra chars 😄.

Nefcanto:
@slorber, assume you write this code:

```
import { SetPrice } from "pricing"
import { Images } from "media"
import { Categories } from "taxonomy"

const entityActions = <>
    <SetPrice />
    <Images />
    <Categories />
</>
```

The point is, how do you know which is a simple JSX and which is a function component?

The above code would throw an exception if:

```
const SetPrice = <EntityAction
    title="Set price"
    dialog={PriceDialog}
/>

export default SetPrice
```

and

```
const Images = () => {

    const { entity } = useContext(EntityContext)

    return <EntityAction
        title="Manage images"
        link={`/media/images?entity=${entity.id}`}
    />
}

export default Images
```

and

```
class Categories extends React.Component {
  render() {
    return <EntityAction
        title="Manage categories"
        link={`/category/manage?entity=${entity.id}`}
    />
  }
}

export default Categories
```

See the difference?

A developer does not need to look into the source of a third-party library (or libraries from other team members) to change its syntax for function components or class components.

However, in the case between a JSX and a component, the rendering syntax is different. This means that the developer should know the internals of another library to change the rendering syntax.

It even gets worse for refactoring. When a component needs to become more complex and changes from a simple JSX into a function component, then you should see failure across any place that has used it as `{SomeCompnent}` and manually change them all to `<SomeComponent />`.

Nefcanto:
> It's unclear what's the point of allowing this instead of just `{LightComponent}`

@slorber two points:

- No need to know the internals of exports of a third-party library (or a team-mate's module) to change the rendering syntax
- No need to change the rendering syntax in case of a refactor

slorber:
I understand the consistency usage, but still think it's a feature for the developer to have to understand what is constant and what is dynamic on the call site.

> how do you know which is a simple JSX and which is a function component?

TypeScript?

CleanShot 2024-11-06 at 16 21 05

> It even gets worse for refactoring. When a component needs to become more complex and changes from a simple JSX into a function component, then you should see failure across any place that has used it as {SomeCompnent} and manually change them all to <SomeComponent />.

If a JSX constant becomes a React component, I actually want to know about it and refactor all the call sites that assumed it was constant. 

If that refactor happens, there are probably props to add to that component anyway, otherwise why would you transition them from constant element to component? The only case I can see is if all the props you add are optional, but again I'll probably want to double-check all the call sites.

josephsavona:
Thanks for the proposal! I don't think we're likely to proceed with this RFC. As @slorber already noted, this use-case can be achieved by assigning JSX to a constant. Or, just defining a function component that returns static JSX. React Compiler will already optimize such a component so that the JSX content is created once per instance, and we will likely explore an optimization that hoists the static JSX out of the component to make it equivalent to 

```js
const element = <div>...</div>
function Component() {
  return element;
}
```

In other words, the new concept here isn't actually unlocking any new capabilities/performance, so it's strictly cognitive overhead.

Nefcanto:
> If that refactor happens, there are probably props to add to that component anyway, otherwise why would you transition them from constant element to component?

@slorber, Contexts. The constant is changed into a component, but the dynamic props are taken from a higher context.

This code (taken from our real codebase):

```
import AdminPanelSettingsIcon from "@mui/icons-material/AdminPanelSettings"
import { EntityAction } from "List"
import RolesDialog from "./Dialog"

const ManageRoles = <EntityAction
    title="AccountsManageRoles"
    icon={AdminPanelSettingsIcon}
    dialog={RolesDialog}
/>

export default ManageRoles
```

Becomes this code:
```
import { useContext } from "react"
import AdminPanelSettingsIcon from "@mui/icons-material/AdminPanelSettings"
import { EntityContext } from "Contexts"
import { EntityAction } from "List"
import RolesDialog from "./Dialog"

const ManageRoles = () => {

    const { entity } = useContext(EntityContext)

    return <EntityAction
        title="AccountsManageRoles"
        icon={AdminPanelSettingsIcon}
        goTo={`/roles?entityType=${entity?.relatedItems?.entityType}&entityGuid=${entity?.guid}&pet=${entity?.relatedItems?.entityType}&pid=${entity.id}`}
    />
}

export default ManageRoles
```

As you can see, in both cases we can call it without props. `<ManageRoles />`.

And with all due respect, in a refactor the ultimate skill is to change the internals without changing the externals. Change in externals is not refactoring, it's redefining contracts.

A simple search related to refactoring shows that. We can find it in Wikipedia too:

> code refactoring is the process of restructuring existing [source code](https://en.wikipedia.org/wiki/Source_code)—changing the [factoring](https://en.wikipedia.org/wiki/Decomposition_(computer_science))—without changing its external behavior.

Nefcanto:
@slorber, thank you for mentioning the TypeScript. I see that it indeed helps to find out the problem while developing. Yet for a lot of reasons many teams don't use TypeScript.

Nefcanto:
@josephsavona, thank you for responding. Facebook created React and as once claimed in the original legacy docs:

> At Facebook, we use React in thousands of components, and we haven’t found any use cases where we would recommend creating component inheritance hierarchies.

This was a very unique way of claiming that *we know it better, OK?*. And it was true. You guys knew it better. To the extent that many people including me left Angular altogether and migrated to React.

Yet the point is that *while you know it better in general, you don't know everything*.

Thousands of projects are built upon React, and none has come to our request. Why? Because they are not this dynamic. If you consider my other proposal (React reflection) you realize that our use cases are here because we go to extremes.

You see, we have built an SPL. This means that we create thousands of applications using this meta-framework (built upon other frameworks like yours). While it's not fair to Next.js, to understand the point you can compare us to it for example.

As an example. consider this form (real code) in `/Blog/Post/Form.jsx:

```
import {
    DialogForm,
    LongText,
    Slug,
    Title,
} from "Form"
import AuthorField from "../Author/Field"

const inputs = <>
    <Title />
    <Slug />
    <LongText
        placeholder="InfraSummary"
        property="Summary"
    />
    <AuthorField />
</>

const PostForm = <DialogForm
    entityType="BlogPost"
    help="PostForm"
    humanReadableEntityType="BlogPost"
    inputs={inputs}
/>

export default PostForm
```

These are all injected dynamically into this form (and hundreds of other forms):

1- Action buttons (save + cancel)
2- Title
3- i18n & l10n
4- Validation (save button is not active until form becomes valid)
5- UX (no layout shift, the first field gets focused when the dialog opens, prevents and warns about unsaved data on navigation, detecting when the form becomes dirty )
6- CRUD and API communication (including form mode for creation and edition)
7- Managing each field (initial value, current value, is dirty, is changed, is valid, ...)

We have seen a lot of other solutions built upon React. None comes close to this level in terms of being dynamic. Therefore we encounter problems that others have not encountered.

You have thousands of components, and we also have thousands of them (2273 at this moment). So we do suffer when we change a constant JSX into a component (to make it internally dynamic) and we have to change hundreds of other components manually. This is a true pain. It doesn't show itself for a simple web app with at most 100 components. Especially when they are all loaded in the same VS Code instance. But our components are spread across more than 100 modules each in a different Git repository. We're even more extended than Next.js because it's just a meta-framework, while we are a meta-framework + a lot of pre-built modules like Blog, Products, Orders, Shipment, Settings, Dashboards, Reporting, Units, Social, Tasks, etc.

Sometimes a change does not take that much on your side. If this change does not take that much, please do us a favor and consider it. If it takes much time, then drop it.

guillaumebrunerie:
Here is a suggestion for you: write a custom Babel plugin that automatically adds `() =>` to all top level variables containing only JSX and whose name starts with a capital letter (turning JSX constants into function components). This way you will get your desired API without having to change React in any way, and you get more control over it if needed in the future.

Assuming your build process already uses Babel, I found it actually surprisingly easy to write custom babel plugins, here is an example wrapping all function components with `observer` (from MobX), as I found it too annoying to do it by hand: https://github.com/guillaumebrunerie/LD56/blob/main/tools/babel-plugin-auto-observe.js

Nefcanto:
@guillaumebrunerie, thank you for the suggestion. We use Vite. Thus I think we should try creating plugins for esbuild. However, two points come to my mind that may not make this solution successful:

1. When we get errors, the error messages do not match the original code (because we have changed the original code on the fly before giving it to the compiler)
2. It's not just top-level JSX variables/constants. They sometimes go deeper. This becomes very tedious to find out all instances of them.

guillaumebrunerie:
> @guillaumebrunerie, thank you for the suggestion. We use Vite. Thus I think we should try creating plugins for esbuild.

The project I shared above uses Vite as well and it wasn't a problem.

> However, two points come to my mind that may not make this solution successful:
> 
> 1. When we get errors, the error messages do not match the original code (because we have changed the original code on the fly before giving it to the compiler)

Babel plugins automatically generate source maps, so you should see your original code (at least it seems to be so for me). 

> 2. It's not just top-level JSX variables/constants. They sometimes go deeper. This becomes very tedious to find out all instances of them.

It's actually the opposite, by default Babel traverses the whole AST, so you automatically get all variables in nested scopes as well (whether you actually want to define nested function components is another question).

Nefcanto:
@guillaumebrunerie, thank you. I give it a try.

EECOLOR:
@Nefcanto Couldn't you use a custom `jsx` function? 

https://esbuild.github.io/content-types/#using-jsx-without-react

In this function you could detect the first argument, if it is not a `string` or `function` you can wrap it in a function.

Nefcanto:
@EECOLOR, I read the link you sent, but it clearly says that those functions are good for when we use JSX outside React. Maybe by changing them other problems get into the scene.

EECOLOR:
@Nefcanto I don't see how there are problems replacing those functions with wrappers. Unless you implement them wrong, but you can create tests if you want to be more confident.

This is generic code of a wrapper:
```
function wrapper(...args) {
  // customization here
  return original(...args)
}
```

If you want more performance and better readability you specify the arguments.

PR #262, opened 2024-12-18
RFC: `useForEach` Hook

pschiffmann:
In this RFC, I propose adding the `useForEach` Hook to React.

### [View RFC](https://github.com/pschiffmann/react-rfcs/blob/use-for-each/text/0000-use-for-each.md)

# Summary

The `useForEach` Hook provides a sane mechanism for calling React Hooks inside loops.

# Basic example

The `useForEach(keys, callback)` Hook calls `callback` once for each element in the `keys` iterable.
The `callback` function is allowed to call Hooks, as if it was at the top level of the component.

```ts
import { useEffect, useForEach, useMemo, useState } from "react";

function MyComponent({ keys }) {
  const results = useForEach(keys, (key) => {
    const [state, setState] = useState(/* ... */);
    useEffect(/* ... */);
    return useMemo(/* ... */);
  });
  // ...
}
```

### DIFF
--- a/text/0000-use-for-each.md
+++ b/text/0000-use-for-each.md
@@ -26,7 +26,7 @@ function MyComponent({ keys }) {
 
 # Motivation
 
-Once you have learned to think in React, synchronizing a _single_ external system with React is straight-forward:
+Synchronizing a _single_ external system with React is straight-forward:
 Connect to the system inside an effect, disconnect from the system inside the cleanup of that same effect.
 React guarantees that effects and cleanups are executed in a well-defined, predictable order, which makes it relatively easy to reason about race conditions and memory leaks.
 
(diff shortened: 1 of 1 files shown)


pschiffmann:
After writing the RFC, I had the idea to check for similar RFCs not just the open issues, but also the closed ones. (Genius, I know 😅) And I found #158 that seems to address the very same problem, just with a slightly different API.

I can see that #158 was closed because:

> It is, in our opinion, a bit too complicated and advanced to be used widely, and not worth it if it's used rarely.

I know this is just a single data point, but at least for me and the type of apps I'm working on, this is a frequent real-world use case. So I'd be really happy if you could give this a second thought, or at least leave it open to see if it gathers more community interest.

vezaynk:
```tsx
function MyComponent({ keys }) {
  const [state, setState] = useState(/* ... */)
  useEffect(() => {
      keys.forEach(...)
  });
  const results = useMemo(() => keys.map(...));
}
```

PR #265, opened 2025-06-26
RFC: `useKey` for resetting state as an alternative to the `key` prop

vezaynk:
# Summary

React has a special prop called `key` that is used to distinguish between rendered instances of a component within the same parent. It's essential for mapping a data array to a component array.

The `key` prop also has an application outside of mapped components. It can be used to force a component to re-render by changing the `key` prop value. This is useful when you want to reset the state of a component or trigger a re-render without changing the component's props.

This technique is described in [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes).

However, writing code where the parent is responsible for the state reset is somewhat of a "tail wagging the dog" situation, as often the component itself is the one that knows best when it needs to reset its state. This RFC proposes to allow components to use the `key` prop behavior in a more flexible way, enabling them to control their own re-rendering behavior by introducing a new hook called `useKey`.

[Link to full formatted RFC](https://github.com/vezaynk/rfcs/blob/main/0000-use-key.md)

### DIFF
--- a/0000-use-key.md
+++ b/0000-use-key.md
@@ -1,0 +1,164 @@
+- Start Date: 2025-06-26
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
+React has a special prop called `key` that is used to distinguish between rendered instances of a component within the same parent. It's essential for mapping a data array to a component array.
+
+The `key` prop also has an application outside of mapped components. It can be used to force a component to re-render by changing the `key` prop value. This is useful when you want to reset the state of a component or trigger a re-render without changing the component's props.
+
+This technique is described in [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes).
+
+However, writing code where the parent is responsible for the state reset is somewhat of a "tail wagging the dog" situation, as often the component itself is the one that knows best when it needs to reset its state. This RFC proposes to allow components to use the `key` prop behavior in a more flexible way, enabling them to control their own re-rendering behavior.
+
+# Basic example
+
+From the [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) article:
+
+Problematic component:
+
+```jsx
+export default function ProfilePage({ userId }) {
+  const [comment, setComment] = useState('');
+
+  // 🔴 Avoid: Resetting state on prop change in an Effect
+  useEffect(() => {
+    setComment('');
+  }, [userId]);
+  // ...
+}
+```
+
+Current solution (note: the parent is responsible for resetting the state):
+
+```jsx
+export default function ProfilePage({ userId }) {
+  return (
+    <Profile
+      userId={userId}
+      key={userId}
+    />
+  );
+}
+
+function Profile({ userId }) {
+  // ✅ This and any other state below will reset on key change automatically
+  const [comment, setComment] = useState('');
+  // ...
+}
+```
+
+Proposed solution (note: the component is responsible for resetting its own state):
+
+```jsx
+export default function ProfilePage({ userId }) {
+  // ✅ This and any other state below will reset on key change automatically
+  useKey(userId);
+  const [comment, setComment] = useState('');
+}
+```
+
+# Motivation
+
+A component should be able to control its own re-rendering behavior, especially when it comes to resetting its state. The current pattern of using the `key` prop requires the parent component to manage this behavior. This leads to the same components internal state behaving differently depending on whether the parent has passed a `key` prop or not, which can be confusing and error-prone.
+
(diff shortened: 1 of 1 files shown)


0000-use-key.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+Given that React's reconciliation algorithm expects to have a determined `key` _before_ processing the body of a component, the `useKey` hook would need to be called at the top level of the component, before any other hooks. This would ensure that the component's identity is established before any state or effects are initialized.
+
+This is particularly important to avoid `useKey` consuming dynamic values that could change during the component's lifecycle such as state or suspense hooks.
+
+```tsx
+export default function ProfilePage({ userId }) {
+  const [comment, setComment] = useState('');
+   // 🔴 Bad: call `useKey` with dynamic internal state is inappropriate
```
trangcongthanh, line 106: IMHO, new hook with new rule to use is not make any sense. We should avoid human error rather than create new potential problem.

Someone gonna write an article to recommend a best practice when working with `useKey` like:

```jsx
function ProfilePage({ userId }) {
  useKey(userId)
  return <Proflile userId={userId} />
}
```

It's just use `key` prop 🤔

0000-use-key.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+
+More broadly, calls to `useKey` should be strictly derived from props. 
+
+Aside from the requirement of being called at the top level of the component, `useKey` would behave similarly to other hooks and force rules of hooks to be followed. This means it cannot be called conditionally or within loops.
+
+# Drawbacks
+
+- `useKey` is technically possible to call within other hooks. This could lead to library hooks hijacking the behavior of components and unintentionally resetting all state.
```
trangcongthanh, line 118: Yes, how can we debug when "something" reset my hooks/components state?

0000-use-key.md
```diff
@@ -0,0 +1,165 @@
...
+- It could lead to confusion for developers who are not familiar with the `useKey` hook, as it introduces a new way of thinking about component re-rendering and state management.
+- It could be used in scenarios where it is not necessary, leading to unnecessary re-renders and performance issues.
+- This is already possible with Higher Order Components or purpose-built wrapper components.
+- Breaks existing assumptions about hooks since it may not be called in any order with other hooks, but rather at the top level of the component.
+
+# Alternatives
+
+The impact of not implementating the RFC would be that developers would continue to use the `key` prop in the parent component to reset the state of child components. This would lead to the same issues described in the motivation section, where the parent is responsible for managing the child's state reset behavior. In other words, the tail will keep wagging the dog.
+
+An alternative to this RFC, which can already be implemented in userland are HOCs or wrapper components that manage the `key` prop behavior. They can look like this:
```
trangcongthanh, line 129: Or the ProfilePage take the userId as key as well.

```jsx
function ParentOfProfilePage() {
  return <ProfilePage userId={userId} orgId={orgId} key={[userId, orgId].join('-')} />
}

function ProfilePage({ userId, orgId }) {
  const [comment, setComment] = useState('');
  // ...
}
```

0000-use-key.md
```diff
@@ -0,0 +1,165 @@
...
+export default withKeys(ProfilePage, (props, Component) => {
+  return <Component key={`${props.userId}-${props.organizationId}-${props.projectId}`} {...props} />;
+});
+```
+
+I don't like this approach because:
+
+1. It requires additional boilerplate code to create the HOC or wrapper component.
+2. There are a dozen subtly different ways to implement this behavior, which can lead to inconsistencies across codebases.
+3. It does not resolve the fundemental issue that the component itself should be responsible for managing its own re-rendering behavior.
```
trangcongthanh, line 146: The child always have ability to re-rendering itself.

When the key change, the component will be unmounted, then remounted as a fresh new. Its not re-rendering.

```jsx
useEffect(() => {
  return () => {} // <-- Unmount will trigger cleanup, re-render may not.
}, [])
```

0000-use-key.md
vezaynk: Ideally, "something" should never do that. I say it's technically possible because you can't prevent functions from calling one another.

However, I think if React strictly enforces `useKey` being above any other hook, this issue largely disappears.

0000-use-key.md
vezaynk: Yes, but this still retains the main issue I perceive with the `key` prop. I think it's backwards for `ParentOfProfilePage` to tell `ProfilePage` when it should construct a new instance instead of `ProfilePage` directly owning it.

To me, this feels akin to if SQL databases inverted the concepts of primary keys and foreign keys.

0000-use-key.md
vezaynk: By "re-rendering behavior" I mean whether it goes through a rerender or a full remount. Both are a kind of "rerender", but the component itself is currently unable to dictate which one it should be.

PR #266, opened 2025-07-12
RFC: Derived Contexts

probeiuscorp:
# [View Rendered Text](https://github.com/probeiuscorp/rfcs/blob/main/text/0000-derived-contexts.md)

Selectors for Context are an in-demand feature, but with a slightly different interface, can be made much more universally useful ([.slice as proposed here is similar](https://github.com/reactjs/rfcs/pull/119#issuecomment-512529871])).

Following from functional programming techniques, this RFC proposes adding:
- a `ReadonlyContext<T>` type which is a supertype of `Context<T>` without `.Provider`

and these methods to Context:
- `.map`, of type: `<T, U>(this: ReadonlyContext<T>, fn: (data: T) => U) => ReadonlyContext<U>`
- `.apply`, of type: `<T, U>(this: ReadonlyContext<T>, fn: ReadonlyContext<(data: T) => U>) => ReadonlyContext<U>` (_I actually recommend one of its alternatives_)

`.map` allows users to refine and select their Contexts, while `.apply` allows users to combine multiple Contexts together. This is covered in detail in Motivation.

Every call, without respect to the identity of the given argument (function / context of functions), returns a distinct Context. To be usable within renders, existing user-land techniques like WeakMap can be used to stabilize the identity of the returned Context to the identity of the given argument.

---

[RFC 119 useContextSelector](https://github.com/reactjs/rfcs/pull/119) draws from much of the same motivation.


probeiuscorp:
@gnoff as you proposed #119, if you had any thoughts on this approach they'd be incredibly interesting. My proposal is essentially an extension to dantman's `.slice`, and [your comment](https://github.com/reactjs/rfcs/pull/119#issuecomment-513310208) on useContexts is very similar to my proposed sequence alternative to `.with`. It's very interesting that seemingly you both came up with applicative functors. I proposed this because it draws from established, well-known patterns and theory. Also because it eliminates worrying about whether the selector is stable or not.

I would love to see progress towards a solution to this problem, but I don't know if you or the community in general has moved on and is no longer interested in this direction.

PR #268, opened 2025-08-30
RFC: Context should accept non-children props as value

Gokul-Gireesh:
This RFC proposes allowing React Context to accept non-children props as the provided value.

- Keeps `value` for backward compatibility.
- If no `value` is passed, all non-children props become the context value.
- Excludes `children` from injection.
- Simplifies context usage and makes it consistent with normal prop-passing in React.

Example:

```jsx
<MyContext foo={foo} bar={bar}>
  <App />
</MyContext>
```
Consumer:

const { foo, bar } = useContext(MyContext);


josephsavona:
Thanks for posting. I don’t think we’re likely to proceed with this change: it’s confusing. In the existing API there’s a clear, 1:1 match between the value you pass to a context provider and the value you get back out w useContext(). The proposal has a number of issues:
* implicit object creation
* unclear what should happen if you pass both a `value` and other props
* Hard to type in type systems

More fundamentally, it isn’t clear this is solving a real problem. We’re always open to exploring api changes, but in this case I don’t think we will proceed given the clear problems and lack of a strong motivation.

Gokul-Gireesh:
Thanks for the detailed feedback. I realize I may not have fully updated the PR comment when submitting — the Value Handling Rules in my commit already clarify how `value` is handled alongside other props.

I understand the concerns about implicit object creation, TypeScript typing, and whether this really solves a problem, and I see how that could add complexity.

Some of the points may reflect my earlier misunderstandings, so I appreciate the clarification and the focus on keeping the API predictable and backward-compatible.

vezaynk:
I think there's very little value in something that's trivially easy to implement by yourself.

`const Context2 = ({ children, ...value }) => <React.Context value={value} children={children} />`

fuunnx:
Yes but be carefull about performances here : you're recreating a new `value` object during each render. You may want to memoize it

PR #270, opened 2025-09-14
RFC: Add useDebounce hook proposal

ytahirkose:
This RFC proposes adding a first-class useDebounce and useDebouncedCallback hook to React.
Debouncing is one of the most common UI patterns (search inputs, resize/scroll events, autosave), but today developers rely on userland utilities that often break under Strict Mode or Concurrent Rendering.

By introducing an official API, React can:

ensure lifecycle correctness and proper cleanup,

reduce subtle bugs from ad-hoc debounce implementations,

provide a canonical, teachable solution for developers.

File
text/0228-use-debounce.md

### DIFF
--- a/text/0228-use-debounce.md
+++ b/text/0228-use-debounce.md
@@ -1,0 +1,518 @@
+- Start Date: (2025-09-14)
+- RFC PR: (leave this empty)
+- React Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


rickhanlonii:
We do! It's called useDeferredValue. If you need denouncing or throttle specifically, the user land solutions work great already.

PR #275, opened 2026-04-20
RFC: Stable `useId()` across Suspense for element-scoped caching

brillout:
[View formatted RFC](https://github.com/brillout/rfcs/blob/brillout/useId-element-cache/text/0000-useId-element-cache.md).


MaxwellCohen:
The instability in `useId()` across suspense appears to me to be a symptom of a larger problem. The larger problem is this [Suspense Caveat](https://react.dev/reference/react/Suspense#caveats) "React does not preserve any state for renders that got suspended before they were able to mount for the first time. When the component has loaded, React will retry rendering the suspended tree from scratch." 


While I am sure there are good reasons not to preserve state for renders, I believe preserving it would unlock many powerful optimizations/coding patterns, such as those listed in this RFC.   I would personally want to use preserving state between suspensions to simplify loading data with use and the React Compiler. When I look at the compiler output for this [example](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEASggIZx4A0BwBAylGFggHZgIEC+BAZjBBgIByGGQpCAOiykyWCAB458BACYIepKABtCPKCwpoILAgFkAngEEsWABQBKGlIIE4xsITQqCAXmJi8ADpGBDo8UjwEWwAGe2cCUTxYE1sAHgYmVnYAPniXVIsAEQjSAi8fYC9uAHpckwJU6ozmNgRsuJZOWSkFJV19Q2Mzc2Lw21ovZAIPGDQWAHMuR2B4tzZCFRLff3IgkNseBDw4AAtbAANq0iw0aowrG+qAEkqVTnP7QLwT1ltbUkcPmyBFIgQAVpAWA57B0XIlkg0sKJssBNuFOI0kW0ANxSLrSAkgSggNY8NDzFAgFb1arVNzYNBaCJGFimCBqKYSECkLRaLl4ongE4QADuAEkWJEYCweWAUBotOxOEA) it looks like the code should work but it will cause an infinite loop.


Update:
I just remembered that React is experimenting with keeping state across renderers, which is usually lost with the experimental [`optimisticKey`](https://github.com/facebook/react/pull/35162).

PR #276, opened 2026-04-30
RFC: Box: eliminate intermediate re-renders by separating data transport from consumption

elakov:
This RFC proposes Box — a referentially stable wrapper that decouples data transport from reactivity in the component tree. Data wrapped in a Box flows through intermediary components without triggering re-renders; consumers explicitly opt in to the slices they need, localizing updates to exactly where data is read. We'd love to discuss whether this is a problem worth solving, and if so — whether the right path is a first-class React implementation or exposing low-level primitives that enable this pattern in userland.

It is primarily aimed at long-lived applications with legacy code, where global refactoring is not feasible or too risky. It gives developers a way to "manually" (but still in react way) control data reactivity and opens the door to other advanced optimization patterns.

## Basic Example

```tsx
const Parent = () => {
    const [value, setValue] = useState(0);
    const box = useWrap(value);
    return (
        <>
            <Middle box={box} />
            <button onClick={() => setValue(value + 1)}>Increment</button>
        </>
    );
};

// This component does not re-render when the parent's state changes
const Middle = React.memo(({ box }) => {
    return <Child box={box} />;
});

const Child = ({ box }) => {
    // You can select only a slice of data, and the component will only rerender if that slice has changed.
    const value = useUnwrap(box, value => value);

    return <div>{value}</div>;
};

```


elakov:
This will also allow to implement https://github.com/reactjs/rfcs/pull/119 in userland

